PR #2, opened 2015-05-29
Add "reserved" TLDs (RFC 2606) to the PSL.

mikewest:
RFC 2606 reserves the following top level DNS names
(https://tools.ietf.org/html/rfc2606#section-2):

* test
* example
* invalid
* localhost

This patch adds them to the public suffix list, ensuring
similar behavior to other top level domains.

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -8382,6 +8382,15 @@ zip
 zone
 
 // ===END ICANN DOMAINS===
+// ===BEGIN RESERVED DOMAINS===
+
+// Reserved Top Level DNS Names (RFC 2606): https://tools.ietf.org/html/rfc2606
+test
+example
+invalid
+localhost
+
+// ===END RESERVED DOMAINS===
 // ===BEGIN PRIVATE DOMAINS===
 
 // Amazon CloudFront : https://aws.amazon.com/cloudfront/


mikewest:
@sleevi: Is this a terrible idea? I noticed this while rewriting some Blink tests that live on `*.test`; that's the only one I really care about, but it seems reasonable to add the others as well.

/cc @gerv

sleevi:
Hi Mike,

There's a couple things with this:
1) This isn't the right way to contribute to the public suffix list. Please use the method documented on publicsuffix.org (a Bugzilla bug with a patch)

2) You describe this as granting the same behaviour of other TLDs, but this is unnecessary, because of the PSL algorithm implictly recognizing them with a "*", per the algorithm. I suspect you are wanting some Chrome-specific behaviour, and we should discuss that in a Chrome bug.

3) If this is well and truly what you meant, I think this will be wrong anyway. See https://bugzilla.mozilla.org/show_bug.cgi?id=1161102

mikewest:
1. https://publicsuffix.org/submit/ says "If there are any additions, removals or amendments to be made, the best thing to do is to create a pull request for our Github repo." Is that outdated?

2. I want `test` to act in the same way `com` does in Blink's unit tests. :) I'll poke you about specifics, I suppose. (FWIW, Blink doesn't seem to implement the "fallback to `*` for no matching rules" bit of the algorithm (and I think that would be a bad result in this specific case)).

3. I think I disagree with your logic in that bug, but I'll hop over there to discuss. :)

sleevi:
Re: 1 - No, I'm outdated. I hadn't realized @gerv had finalized that now.

Re: 2 - Let's follow-up offline. Chrome's implementation in //net certainly does, and if Blink isn't respecting that, that is a real Blink bug we need to fix.

gerv:
Yes, I've put in place the process that this is now the PSL master; however, I've not yet got a process for getting copies back into Mozilla :-)

PR #4, opened 2015-06-18
Added Slovak government domain (gov.sk).

richard-mihalovic:
Added Slovak government domain (gov.sk).

Some example domains:
http://www.finance.gov.sk/
http://www.telecom.gov.sk/
http://www.employment.gov.sk/

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -5857,6 +5857,7 @@ sj
 // sk : http://en.wikipedia.org/wiki/.sk
 // list of 2nd level domains ?
 sk
+gov.sk
 
 // sl : http://www.nic.sl
 // Submitted by registry <adam@neoip.com> 2008-06-12


weppos:
Thanks for your submission @richard-mihalovic. As per the PSL rules, we can approve new suffixes for an existing TLD only if the patch comes from the registry itself or can be validated with public registry information.

Are you somehow connected with the registry? I tried to check on the official .SK website whether they list the SK suffixes, but I was not able to find any official page. Are you able to provide me some guidance on it?

Otherwise, I will have to contact the .SK registry in order to have them validate the request.

gerv:
weppos: any news?

Googling for "site:gov.sk" reveals a load of hits, so I think this is a valid public suffix.

richard-mihalovic:
Hello sorry for the late response (I was on vacation). I can't give you answers on your questions because I'm not connected to any of these organizations. But maybe you can try to contact somebody from these sites:

http://www.nases.gov.sk/26687/ (National Agency for Network and Electronic Services)
https://www.sk-nic.sk (Slovak top level domain registry)

https://www.sk-nic.sk/documents/domeny_1.txt (list of all registerd Slovak domains)

R.

gerv:
Some governments might like to use single signon, and adding gov.sk to the PSL would prevent the Slovak government from doing that. So I don't think we'd want to do this without their explicit permission. If someone wants to get that, go right ahead, but I don't have time to chase this. :-|

Gerv

PR #5, opened 2015-06-19
update public_suffix_list.dat for .za

xaks:
ZADNA administers the TLD .za for South Africa. Information for this addition was found here:
http://www.zadna.org.za/content/page/domain-information


dougforpres:
Hi moderator, any news on the progress of this patch?

weppos:
I was able to validate the patch using the supplied link. However, it seems the list is missing the 3 private domains. I'll add them.

PR #8, opened 2015-06-26
Add new TLDs contracted through 6/7/2015

dnsguru:
I'll make a subsequent branch request for those through 6/25 shortly - pardon while we learn this new process for updates.  We'll hit a cadence soon.

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -6780,7 +6780,16 @@ xxx
 *.zw
 
 
-// List of new gTLDs imported from https://newgtlds.icann.org/newgtlds.csv on 2014-10-04T06:02:06Z
+// NEW TLDS SECTION FROM https://newgtlds.icann.org/newgtlds.csv on 2015-06-09T06:02:08Z
+
+// aaa : 2015-02-26 American Automobile Association, Inc.
+aaa
+
+// aarp : 2015-05-21 AARP
+aarp
+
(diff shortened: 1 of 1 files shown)


gerv:
I've just realised that the copy of the PSL in this repo is messed up, and I did an incomplete conversion. It only has the changes until October last year :-( That is why your patch is so big. I'm trying to fix that now.

dnsguru:
@gerv is this addressed?

gerv:
Hi Jothan,

Sadly not yet :-( I need to extract the history of the relevant files from the mozilla-central history, still as an hg repo. That takes about 8 hours. I did it last week, but foolishly wrote the resulting repo to /tmp - and then rebooted my machine. So I need to do it again. I'm travelling today and it's a CPU-intensive operation so it doesn't really fit well. I'll kick the job off again tomorrow, and then the day after will convert the new, much smaller, repo to git, work out how to merge the changes since October into this repo, and then push it. And then we'll be in a position to start taking updates again.

Sorry it's taking so long. :-|

gerv:
Hi Jothan,

I've now merged in what I _think_ are all the remaining changes from the original Mozilla repo. So we are now unblocked for adding new stuff. Please can you re-create this patch?

gerv:
Never mind - I just created a new patch. Thanks!

Issue #7, opened 2015-06-23
.cy (cyprus) ccTLD 2nd level domains


hserus1477:
Hi - you currently just have 

// cy : http://en.wikipedia.org/wiki/.cy
*.cy

However there are multiple 2nd level domains here.  

I am NOT affiliated with the cyprus ccTLD. However this is right off their website.

http://www.nic.cy/nslookup/online_database.htm
The dropdown shows you - 

com.cy
biz.cy
ltd.cy
pro.cy
net.cy
org.cy
name.cy
tm.cy
ac.cy
ekloges.cy
press.cy
parliament.cy

Here's a screenshot

![image](https://cloud.githubusercontent.com/assets/13009456/8296912/2db2d972-1974-11e5-9e2c-22d617efd88e.png)

weppos:
Hi @hserus1477, thanks for your patch. `*.cy` means that the domain is only allowing one-level suffixes, therefore it's mostly equivalent with listing all the suffixes.

However, I believe we were trying to switch from generic definitions to explicit definitions because of a preference from Google Chrome guys, therefore the patch may probably make sense.

@gerv @sleevi do you confirm that an explicit list is preferable instead of a generic wildcard in this case?

sleevi:
Personally, I've preferred the wildcard, but Peter Kasting preferred the more explicit form.

The argument for the explicit form is:
1. If the ccTLD ever wishes to allow first level registrations, then having a wildcard policy prohibits them from doing so until all consumers of the PSL are updated
2. With Chrome's policy of using the PSL for determining whether to search vs navigate, having a *.ccTLD means that if a user types a.baz.ccTLD when they meant a.bar.ccTLD, we'll attempt a navigation (because we treat baz.ccTLD as a valid ccTLD), the user will get an NXDOMAIN page, rather than our default behaviour, which would be to execute a search.

The second issue is less important (and we're more fluid to change), and with our rewritten public suffix implementation, we can efficiently handle many explicit enumerations. However, implementations such as Firefox (which use a hash solution, similar as Chrome used to), many explicit entries may be worse for performance for them.

So to directly answer your question: A weak preference for a more explicit listing.

dnsguru:
The request from chrome is not alone; The explicit listing is also desired from many of the third party integrators and developers where possible, such as for anti-spam or form / data entry validation in sofware libraries.  The more pedantic the detail, the better.

hserus1477:
[...]

PR #14, opened 2015-07-12
update tests/test_psl.txt since .cy is no longer wildcarded

dkg:
1bb3c8c34b17e9afe34e308069daeffac91380b2 resolved
https://github.com/publicsuffix/list/issues/7 by removing the *.cy
wildcard in favor of an explicit listing.

However, it did not update test_psl.txt to match, so any tests that
use it would fail.  This patch updates the test suite to use *.il
instead of *.cy.


weppos:
Thanks for the patch!

PR #18, opened 2015-07-27
Add new gTLDs up to 27-07-2015

sleevi:


sleevi:
@gerv What's the review process for these sorts of updates? Anything special?

// cc @weppos

gerv:
Not really - particularly if it's a known good source ;-)

PR #19, opened 2015-07-31
Add new gTLDs up to 2015-07-31

sleevi:
@gerv @weppos @dnsguru - Any objections for me just merging these explicitly each update? Or do you want to review each one of them?

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -6919,7 +6919,7 @@ web.za
 *.zw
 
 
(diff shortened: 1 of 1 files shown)


dnsguru:
I think merging explicitly each update is a good idea - there was just a very large block of .brand TLDs that pushed through - so we're well into 3/4 done with the new gtlds.

sleevi:
@dnsguru I moreso meant "Should I wait for approval from one of you" or should I just use the script and update.

gerv:
sleevi: I think it's reasonable for you to merge updates from the ICANN list without further review or approval. As long as you run the test suites (that we don't have yet) of course :-)

PR #24, opened 2015-08-11
update Israel TLD

xaks:
The IANA lists isoc.org.il as the group delegated to managing the .il TLD.
http://www.iana.org/domains/root/db/il.html

isoc.org.il states that co.il and org.il are the only two domains available from accredited registrars.
http://www.isoc.org.il/domains/

The following page describes the .il TLD namespace.
http://www.isoc.org.il/domains/il-domain-rules.html

However, that page can be a bit unreliable. A cached copy from March 2015 can be found here:
https://web.archive.org/web/20150318213406/http://www.isoc.org.il/domains/il-domain-rules.html


xaks:
Hey guys, just checking in to see if there's any progress on getting Israel's TLDs merged into the list. Thanks.

xaks:
Hello again, any news on the progress of this patch?

gerv:
I think this will break the tests, which use "*.il" as an example of a wildcard rule. So that file would need to be updated also.

xaks:
I was not able to find any specific references to any .il domains in the test data, and the tests appear to be passing:

https://travis-ci.org/xaks/list/builds/80351931

xaks:
As a note, I've had to create a travis-ci build specifically for this branch, which is the cause of the merge conflict.

rockdaboot:
@xaks You have '*.il' which already includes 'ac.il' which you also specify. I am unsure what exactly you are going to do.

gerv:
.il is used on lines 41-44 of https://github.com/publicsuffix/list/blob/master/tests/test_psl.txt .

Gerv

xaks:
Ah, I see what you're saying now @rockdaboot. I've updated the list and the tests for the wildcard TLD check. Thanks for pointing that out @gerv. I had those showing up as "*.cy" tests. 

As an aside, how are the .js files used for testing? I couldn't get them to run using a node console, and they don't look like QUnit tests.

gerv:
@xaks: Do we actually want those changes to .travis.yml? It seems like we don't...

xaks:
@gerv The changes to the travis configuration file were there to test my specific branch on travis-ci.org. I've reverted the travis configuration changes.

PR #29, opened 2015-08-19
Add withyoutube.com as a new public suffix

pames:
Please advise as to what verification is necessary to make this change (As FYI, I'm the same Phil Ames as the one that requested a similar change in https://bugzilla.mozilla.org/show_bug.cgi?id=965744)

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -11019,6 +11019,7 @@ googleapis.com
 googlecode.com
 pagespeedmobilizer.com
 withgoogle.com
+withyoutube.com
 
 // Heroku : https://www.heroku.com/
 // Submitted by Tom Maher <tmaher@heroku.com> 2013-05-02


pames:
friendly ping?

sleevi:
@gerv I'm not sure if you still wanted to manually approve these, or whether you'd set up notifications for all pull requests / comments, and that satisfies as keeping you in the loop.

But yes, I can independently confirm this is my colleague Phil Ames from Google, who operates with withyoutube.com, which has been confirmed out of band with Mark Monitor (plus, yanno, the ns1-ns4 at Google).

On a procedural point, do you need @pames to update line 10941 as well? I'm not sure how you want the metadata on 10940/10941 to read.

weppos:
@sleevi as far as you can confirm the validity of the patch, you can merge it. As far as I know, @gerv would like to be notified of merged tickets. Technically, each commit would likely be a merge, hence I'm still not sure how the process would work.

So far I've been keeping @gerv in the loop mentioning him each time I was able to verify the authorship and I was proceeding with a merge.

As for 10941, I've been keeping track of them but that was mostly because Bugzilla was separate from Hg. To be honest, I wonder if we should continue to do it given some entries may be outdated or not accourate. Moreover, we don't need the date (we can get it from the SCM) nor the action.

What we can probably do is keep a list of approved/verified emails at the top of each section, e.g.

```
// Google, Inc.
// Eduardo Vela <evn@google.com>, Ryan Sleevi <...>
```

The only reason for doing it is because it may make it simpler to contact one of the trusted recipients or quickly check for authorized requests. However I don't think we need any other info (such as submitted, updated, etc).

gerv:
The metadata is supposed to be useful, but it is not supposed to be structured and there is no requirement that it's comprehensive. It's just handy to know who to talk to. We could keep it in a Rolodex :-)

I don't seem to get notified when someone opens a PR against this repo; anyone know how I can set that up?

Gerv

weppos:
> I don't seem to get notified when someone opens a PR against this repo; anyone know how I can set that up?

@gerv You should click on "watch" at the top of the repo and select "Be notified of all the conversations".

PR #32, opened 2015-09-01
Adds umbraco.io domains

Shazwazza:
Adding umbraco.io and *.umbraco.io to the list for our upcoming Umbraco as a service release

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -11144,6 +11144,11 @@ hk.org
 ltd.hk
 inc.hk
 
+// Umbraco : https://www.s1.umbraco.io/
+// Submitted by Shannon Deminick <shannon@umbraco.com> 2015-09-01
+umbraco.io
+*.umbraco.io
+
 // Yola : https://www.yola.com/
 // Submitted by Stefano Rivera <stefano@yola.com> 2014-07-09
 yolasite.com


gerv:
These two rules contradict each other. Say you have a domain "www.foo.umbraco.io". The first rule says that "www.foo" is the registered part and "umbraco.io" is the suffix. The second rule says that "www" is the registered part, and "foo.umbraco.io" is the suffix. Which is it? :-)

Shazwazza:
Hi, thanks for the reply and clearly I've misunderstood :)

I was under the impression that with these rules

> umbraco.io
> *.umbraco.io

* Cookies may not be set for umbraco.io
* Cookies may not be set for foo.umbraco.io
* Cookies may be set for foo.bar.umbraco.io

But I guess the first rule is not necessary? The domains hosted under umbraco.io will always be 4 levels, like: foo.bar.umbraco.io and hello.world.umbraco.io but never foo.umbraco.io or just umbraco.io.

gerv:
Actually, exactly how the PSL should behave for umbraco.io when it has simply the following rules:

```
io
*.umbraco.io
```

is currently a matter of debate. So we should make it clear by including the additional rule.

Gerv

gerv:
@Shazwazza: this got reverted due to test failures. Can you open another PR to add it again, and we can continue the discussion?

Gerv

PR #35, opened 2015-09-11
Revert "Adds umbraco.io domains"

gerv:
Reverts publicsuffix/list#32 -- fails tests. Let's work out whether the test is wrong or the patch is wrong, then act accordingly.

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -11145,11 +11145,6 @@ hk.org
 ltd.hk
 inc.hk
 
-// Umbraco : https://www.s1.umbraco.io/
-// Submitted by Shannon Deminick <shannon@umbraco.com> 2015-09-01
-umbraco.io
-*.umbraco.io
-
 // Yola : https://www.yola.com/
 // Submitted by Stefano Rivera <stefano@yola.com> 2014-07-09
 yolasite.com


gerv:
https://travis-ci.org/publicsuffix/list/builds/79863583

gerv:
@rockdaboot : comments?

rockdaboot:
@gerv I look at it in a few hours. Travelling right now.

rockdaboot:
@gerv To add a domain and it's subdomains: just add the domain to the list
Example: for .com and it's subdomains we have 'com'
Having both 'com' and '*.com'... well these rules 'overlap' somehow.
Solution here: just add 'umbraco' and that's it.

gerv:
rockdaboot: that's not necessarily so; that behaviour is implementation-dependent. If the PSL contains (lines numbered for clarity):

```
1) foo
2) *.wibble.foo
```

then some implementations say that wibble.foo is a public suffix, and some don't. Which is why I wanted to be explicit by adding both. However, it seems to break your test:

psl_is_public_suffix(xxxx.umbraco.io)=1 (expected 0)

Your test says that with the following PSL:

```
1) foo
2) wibble.foo
3) *.wibble.foo
```

then is_public_suffix(xxxx.wibble.foo) should return 0. But it shouldn't, it should return 1, due to rule 3). I suspect this is because you are testing the rules individually, and are testing the rule "wibble.foo" (rule 2), and therefore assuming that xxxx.wibble.foo should not be a public suffix. But it is, according to rule 3.

Does that make any sense?

Gerv

rockdaboot:
@gerv First part makes sense (and was subject to discussion on the libpsl mailing list in the past).

Second part describes a missing feature in libpsl.
Before today we never had a 'wibble.foo' together with a '*.wibble.foo'. The current libpsl code does not expect it and thus fails. I looked at the code yesterday and it is absolutely clear what goes wrong in this case. A side effect is that changing the order to '*.wibble.foo' 'wibble.foo' would make is_public_suffix(xxxx.wibble.foo) return 1. 

I'll fix that ASAP (hopefully today).

rockdaboot:
@gerv Just released libpsl v0.9.0 which fixes the umbraco.io/*.umbraco.io.

PR #42, opened 2015-10-15
Add  Nicaragua's SLD

matiasvillanueva:
// ni : http://www.nic.ni/
com.ni    
gob.ni    
edu.ni
org.ni
nom.ni
net.ni
mil.ni  
co.ni
biz.ni
web.ni    
int.ni
ac.ni
in.ni  
info.ni

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -4437,8 +4437,21 @@ gov.ng
 mil.ng
 mobi.ng
 
-// ni : http://www.nic.ni/dominios.htm
-*.ni
+// ni : http://www.nic.ni/
+com.ni    
+gob.ni    
+edu.ni
+org.ni
+nom.ni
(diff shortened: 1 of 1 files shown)


gerv:
Matias: how do you know that this is a complete list? If a 2LD was only available to e.g. the government or the military, it might not be listed on such a page? (in this case, there are obviously entries for those two examples, but the point remains - there may be unlisted domains.)

Gerv

matiasvillanueva:
Hi Gerv!

Yeap, good point. But it's the same in other TDLs like *.bo, and *.ar. I don't know in witch cases you can realy verify that the list is complete.

If you go to the statistics (nic.ni) you will see a full ranking of the SLDs created so far. 
You can also check this link https://es.wikipedia.org/wiki/.ni

Matias

PR #55, opened 2015-11-28
Delete trailing whitespace from .ni

ko-zu:


weppos:
Thanks for the patch @ko-zu. It's important to note that whitespaces are not relevant as per the specification of the list format, but I agree that it's better to not add them and I merged your patch.

Thanks!

/cc @gerv

PR #63, opened 2015-12-03
Add prgmr.com customer domain suffix

sarahn:


weppos:
Validated via email.

/cc @gerv

PR #67, opened 2015-12-07
add domains actindo.com, actindo.io, dev.actindo.com, dev.actindo.io …

pprasse:
add domains actindo.com, actindo.io, dev.actindo.com, dev.actindo.io to public prefix list

(actindo internal issue ERP-288)

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -10622,7 +10622,7 @@ zuerich
 // (Note: these are in alphabetical order by company name)
 
 // Actindo AG : https://www.actindo.com/
-// Submitted by Patrick Prasse <hostmaster@actindo.com> 2015-12-07
+// Submitted by Patrick Prasse <hostmaster@actindo.com>
 actindo.com
 dev.actindo.com
 actindo.io
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -10561,6 +10561,13 @@ zuerich
 // ===BEGIN PRIVATE DOMAINS===
 // (Note: these are in alphabetical order by company name)
 
+// Actindo AG : https://www.actindo.com/
+// Submitted by Patrick Prasse <hostmaster@actindo.com> 2015-12-07
+actindo.com
+dev.actindo.com
+actindo.io
+dev.actindo.io
+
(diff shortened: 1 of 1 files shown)


sleevi:
Hi @pprasse - Sorry for the delays on this.

Can you:
1) Update this PR according to the format in https://github.com/publicsuffix/list/wiki/Guidelines
  * Remove the date
2) Indicate which validation method to use
  * While it may be possible to validate actindo.com via WHOIS+email, I'm unable to find contact information for actindo.io from nic.io. It would be strongly preferable to include TXT records
3) Provide a brief description of why you want these suffix(es) to be added. For example, given that the PSL's rapid growth will make it harder for applications to use, it would be desirable to use alternative schemes. That is, rather than `foo.actindo.io` and `foo.dev.actindo.io`, having `foo.actindo.io` and either `dev.foo.actindo.io` or `foo-dev.actindo.io` can help the PSL greatly, be reducing the number of entries that need to be added.

pprasse:
Hi @sleevi,

- Date was removed
- Please use the "DNS Authentication" validation method. The necessary TXT records were just added

Actindo AG 's product is a cloud based ERP and business processing platform (named IntelliCore) which can be heavily extended with own plugins by each tenant.
As a PaaS and SaaS provider we have two different systems for our customers: the productive system (`[TENANTPREFIX].actindo.com`), and the development system (`[TENANTPREFIX 1-n].dev.actindo.com`).

Those systems would be very difficult to separate using `foo-dev.actindo.com` and `foo.actindo.com`, as we will be using completely separate infrastructure and therefore nameservers etc.

`actindo.com` and `actindo.io` separation: The ERP platform is served using actindo.com (where the tenant's plugins may overwrite any path, for example `/`), while the tenant's files are served using `actindo.io` (using WebDAV amongst others). So if a tenant's plugins would overwrite `/` on `TENANT.actindo.com`, access to the tenants files would not be possible anymore. Therefore we need `actindo.io` here.

sleevi:
I'm unable to validate the domain using TXT records for any of the domains.

weppos:
Closing as no followup.

PR #79, opened 2015-12-15
Add freedns.afraid.org domains to public_suffix_list.dat

flyser:
The subdomains of these domains are publicly available on freedns.afraid.org.

The list has been generated using the following command (may the owner of the page forgive me):

    PAGE=1
    while true ; do
      [[ $(curl -s -H "Accept-Language: en-US,en" "http://freedns.afraid.org/domain/registry/?page=${PAGE}&sort=1" | gawk 'BEGIN { FS="[<>]" } /action=\/domain\/registry\// {flag=1} /<\/table>.*<\/table>/{flag=0} flag { if ($(NF-16) == "public") print $(NF-32); }' | tee -a result.txt | wc -l) -lt 1 ]] && break
      ((PAGE++))
    done


gerv:
We certainly aren't going to do this without the consent of the afraid.org admins. What if the same person has registered two subdomains of one of these domains (seems quite likely) and wants to communicate between them?

IMO it is not reasonable to increase the size of the PSL by multiples just to support one provider.

afcady:
What if the same person has registered two ``.com`` addresses and wants to communicate between them?  That's exactly the same thing.

The domains in the list here *are* public suffixes, so it is not in the interest of web clients to allow them to read each others' cookies.  If the public suffix list cannot support them, it just means that the public suffix list cannot support the number of public suffixes that actually exist.

(Probably the way DNS was supposed to work is that *every* domain should be assumed to be a public suffix, but I guess that ship has sailed.)

LNFWebsite:
As a note of interest, I am contacting the owner of FreeDNS involving the addition of some (maybe not all) subdomains to the PSL.

I have suggested to him that only subdomains under his control should be added.

As I understand it, the PSL volunteers have to contact the owner of the domain, not just the administrator of the DNS service under which these subdomains are run.

However, it should not be deemed unreasonable to add a considerable quantity of domains if there is such a provider who requests them to be added. The only conclusion that can be made from this statement is that the volunteerism of the PSL is not in fact looking towards the discovery of all public suffixes, but is rather operating on an as needed basis.

If this is the case, then furthermore bulk additions to the PSL are unwarranted and unnecessary. Services like Let's Encrypt should look elsewhere or rather form a list or alternate method of their own.

sleevi:
On Sunday, July 10, 2016, LNFWebsite <notifications@github.com> wrote:

> As I understand it, the PSL volunteers have to contact the owner of the
> domain, not just the administrator of the DNS service under which these
> subdomains are run.
>
The party responsible for delegating these domains is responsible for
submitting the request - not the party to whom they have been delegated
(per-se)

However, it should not be deemed unreasonable to add a considerable
> quantity of domains if there is such a provider who requests them to be
> added. The only conclusion that can be made from this statement is that the
> volunteerism of the PSL is not in fact looking towards the discovery of all
> public suffixes, but is rather operating on an as needed basis.
>
I'm afraid you've misunderstood the objection rather considerably.

The PSL is both security critical and performance critical in the
applications that use it. As it stands, because no suitable technology yet
exists for this to be discoverable in even a remotely efficient manner, the
PSL must be treated as a "shared resource" by the Internet; every addition
makes it slower for everyone consuming the PSL, makes it harder to securely
manage correctness of the PSL, and makes it more difficult for application
software providers - browsers - to consume the PSL.

At the risk of a bad analogy, the request for 25K domains would be similar
to a party requesting a class A IPv4 block, simply so they could make
documentation of their network easier. While it isn't fundamentally
unreasonable to request a class A - or the addition of 25K domains - it is
consuming a substantial portion of a finite resource (IP addresses, PSL
entries) for what is, in essence, vanity.

Purely pragmatically speaking, that sort of DNS delegation being done is
being done so irresponsibly and without consideration for the broader
ecosystem, and it's reasonable to reject such a request if it would
interfere with the secure operation of others.

For example, if the PSL were to grow to 250K in size, that would make it
impossible to use in a number of memory-constrained environments.
Optimizing the PSL - by judicious use of wild cards and minimizing
unnecessary entries ("test", "Dev", "staging", etc) - is a responsible
thing to do and reasonable thing to request.

> If this is the case, then furthermore bulk additions to the PSL are
> unwarranted and unnecessary. Services like Let's Encrypt should look
> elsewhere or rather form a list or alternate method of their own.
>
Or Let's Encrypt should use an alternate form of rate limiting and abuse
detection.
>
>
Or such domain holders should stop focusing on vanity domains that either
leave their users insecure (absent the PSL) or focus on spending
engineering resources to help the broader Internet community support their
use case without all of the pain and damage they currently inflict.

LNFWebsite:
Thank you for clearing up some of the confusion.

> The party responsible for delegating these domains is responsible for
submitting the request - not the party to whom they have been delegated
(per-se)

So, as I understand it, it would be the task of the owners of the domains who are delegating sub-domains to ensure that their domain is considered an eTLD through the PSL.

This leaves me on my first impression that adding all FreeDNS domains would be likely impossible.

> As it stands, because no suitable technology yet
exists for this to be discoverable in even a remotely efficient manner, the
PSL must be treated as a "shared resource" by the Internet; every addition
makes it slower for everyone consuming the PSL, makes it harder to securely
manage correctness of the PSL, and makes it more difficult for application
software providers - browsers - to consume the PSL.

This quite frankly seems like an archaic method for defining public suffixes.

"No suitable technology exists"... I don't mean to be blunt, but... Really?

In the age of vast quantities of data flowing freely through DN servers to define random IP addresses to other often random domains, why couldn't this seemingly simple property of domains be defined by the DN servers around the globe. They are storing data on these domains anyway, giving each a direction to an IP address on the net. An addition of an eTLD flag to be stored with the domain (or subdomain like co.uk) would be of little consequence to workload or memory (much less than a giant list).

Given that adaptation of DN servers to hosting more advanced properties of domains and subdomains will probably not be happening in the near term, there are other viable solutions which may be implemented into the existing PSL system.

Might I suggest that instead of loading an entire list of public suffixes for each individual use of the PSL, that the list be stored on a system which may be queried by the browser or organization with a specific request.

Instead of loading the entire list:

```
https://publicsuffix.org/list/public_suffix_list.dat
```

Load only the *pertinent data* required:

```
https://publicsuffix.org/list/?q=co.uk
```

Returning a true or false statement as to the presence of the domain or sub-domain on the list.

> Purely pragmatically speaking, that sort of DNS delegation being done is
being done so irresponsibly and without consideration for the broader
ecosystem, and it's reasonable to reject such a request if it would
interfere with the secure operation of others.

The fact that a 25K introduction to the PSL may actually adversely affect the operation of other important systems on the web shows the dire need for a new method of implementation, *not a fault on the part of individuals on the net delegating their sub-domains to other people.*

Such a line of thinking supports a theory of the PSL being faded out very soon. The internet is expanding and has been since its inception. Over the next couple of years, the PSL will have hundreds of thousands if not millions of additions. If managers of the PSL are worried about 25,000 new suffixes being added, they are not prepared for the future.

> Or such domain holders should stop focusing on vanity domains that either
leave their users insecure (absent the PSL) or focus on spending
engineering resources to help the broader Internet community support their
use case without all of the pain and damage they currently inflict.

You are calling for an end to sub-domain registration. I disagree with your premise that the "internet community" has to look out for the greater good when sub-delegating their domains. The PSL is broken if it cannot support and advocate for a major part of the internet. The thought that placing their sub-domains on the PSL will slow down every browser on the face of the earth should not have to cross their minds.

It is the job of communities like this to support the internet how it is, not to reject parts of it which do not work with a antique system.

sleevi:
> So, as I understand it, it would be the task of the owners of the domains who are delegating sub-domains to ensure that their domain is considered an eTLD through the PSL.
> This leaves me on my first impression that adding all FreeDNS domains would be likely impossible.

This means it's FreeDNS's responsibility, and if those parties are sub-delegating, their responsibility. I'm not sure what you see at odds here.

> I don't mean to be blunt, but... Really?

Unfortunately, your reply, while perhaps meant to sound incredulous, comes off extremely rude.

Yes. Really. Your suggestion - online querying - is already completely at odds with how every system implements it (shipping a copy of the public_suffix_list.dat *precisely because* it is too slow to query). We're talking performance concerns on the matter of nanoseconds mattering here (yes, really; Opera, Mozilla Firefox, and Google Chrome have all continued to optimize the handling of the PSL because it's essential for processing cookies, which is essential to processing requests, and those nanoseconds quickly add up to microseconds which, in terms of page load times, can make significant differences on user engagement on websites, on the order of millions-to-billions of dollars in aggregate)

I'm glad you're excited about exploring viable alternatives. The ecosystem has as well. There's even an IETF discussion for this. This is an issue that's been going on for _8 years_ and yet we still haven't found a proper technological solution that would be able to work with clients and servers alike. I'm sure it sounds easy, but reality isn't. I would encourage you to explore the history of the discussion, or at least, try to be considerate in tone to understand the nature of the problem and why this isn't a lack of creativity.

While I appreciate your claiming that this is the "Public Suffix maintainers" fault for failing to 'keep with the times', please realize this list is volunteer operated, what you're truly complaining about is the parties that *consume* the public suffix list (e.g. browsers, client software, anything dealing with cookies). If you want there to be a scalable solution, then you should work within the relevant communities of browsers and IETF to first understand the history of why this came to be, understand the challenges that have for the past decade prevented reasonable alternatives, and understand that this is an ecosystem issue that affects more than just the PSL.

For further reading, I would suggest you read https://www.icann.org/en/system/files/files/sac-070-en.pdf and https://www.ietf.org/mailman/listinfo/dbound to better understand the problem before commenting further to suggest solutions.

LNFWebsite:
Thank you for your response.

I'm sorry if my tone sounded rude, it's merely out of frustration.

So, do browsers package this list along with the software of the browser, or do browser organizations store it on their own and query their database for their user base?

LNFWebsite:
> There is currently no programmatic way to determine public suffixes for DNS
entries with more than one label from a registered multi-level domain, yet
tracking public suffix definition and use accurately is critically important for
security, privacy, and usability issues in many modern systems and applications.
- https://www.icann.org/en/system/files/files/sac-070-en.pdf (page 9)

This is well understood. What isn't as well understood is the implementation of a general use list rather than a simple flag upon the DNS resolving request.

Likewise, if such "public suffix" only affects the party which owns the domain, why should the median processing speed of resolving PSL requests be 23.9 days? - https://www.icann.org/en/system/files/files/sac-070-en.pdf (page 16)

If it is the effort put into maintaining a gargantuan list of requested "public suffixes," then all the more it should be handled in an automated and secure way through the worldwide DNS system.

sleevi:
> do browsers package this list along with the software of the browser

https://publicsuffix.org/learn/ lists Public Suffix consumers. Of those, (nearly?) every one of them ships their own copy of the PSL, directly, as part of the software. Often processed in some form (for example, Firefox creates a perfect hash of the domains for efficient querying, Opera/Chrome/libpsl use a DAFSA, various Java libraries use either a pre-computed or on-demand trie, etc).

Some library-implementations allow application callers to supply their own (full) copy of the PSL, which is then processed into an internal data structure.

I'm not aware of any implementation that does any form of online querying. That would be horribly inefficient for virtually all PSL use cases.

As for DNS flags, they are practically unworkable. I encourage you to continue to read more about the ecosystem of software, clients, and domain servers to understand why such a change has issues. For example, you will realize it's more than just a single flag, you need significantly more response data to indicate that, e.g. `internal.subdomain.something.example.com` has the public suffix of `example.com`, but `internal.subdomain.somethingelse.example.com` has the public suffix of `somethingelse.example.com`. Then you also have to work through how to reliably deliver that information through existing DNS clients and servers, and you'll understand more about why that's not workable.

That said, this bug would seem an inappropriate venue to continue to discuss about future PSL-alternatives. I would suggest you examine the archives and previous drafts (including the expired ones) in the IETF DBOUND WG to understand more about the problem space - including the I-Ds that try to outline the set of problems :)

LNFWebsite:
Yes, thank you for the rigorous debate!

There are many options open for making this system better, and I look forward to learning more about the possibilities and their shortcomings.

Now, I've gotta work to get a few FreeDNS domains on the list ;-)

Koxx3:
any news on this ?
it doesn't seems to be supported ?
did freedns request the inscription to the white list ?

LNFWebsite:
@Koxx3 Unfortunately, the addition of FreeDNS domains to the PSL may only be done by the Domain registrant themselves.

I contacted the owner of FreeDNS and he did not know anything about this. Likewise, I attempted to contact the owner of us.to domains (which make use of FreeDNS) but never got any reply.

As of now, you just have to wait until the Let's Encrypt rate limit expires, then you may grab your SSL certificate.

As for security, the domains on FreeDNS are still venerable to cookies being shared across what your browser thinks are related sub-domains of the same organization... Even though they are not.

PR #85, opened 2015-12-17
Add GoIP DNS Services http://www.goip.de

goipde:


weppos:
@goipde Can you please amend the patch to include a valid contact and submission detail for the entry.

```
// GoIP DNS Services : http://www.goip.de
// Submitted by John Doe <example@example.com>
```

Also please quickly explain here why you want the domain to be included in the list.

goipde:
Some of my users asked me, to add the domain to this list to be able to use "Lets Encrypt".

weppos:
@goipde Is Let's Encrypt the only reason to be listed? If that's the case, or you don't fully know the implications of being listed, then you should probably not request the listing. Rather, the best thing to do is to wait until Let's Encrypt will lift the limit.

Being listed with the only intent to bypass Let's Encrypt current limit is not the most appropriate choice.

nitmir:
It seems to be a dynamic dns provider, so it should be on the PSL. no ?

weppos:
@nitmir it doesn't have to, it depends. We have cases where dynamic providers explicitly denied their consensus to be in the list, because of some of the side effects on how browsers than handled cookies and SSL certificates, as well other consumers.

goipde:
Yes, I'm offering dyndns services (like dyndns, noip ...). Hosts below goip.de belong to different users. 

After reading "Learn more" for my understanding there is no negative impact, as every site below goip.de should be independent without cross references. I could not find anything about SSL, but I guess wildcard certificates won't work anymore?

sleevi:
Thank you @goipde and sorry for the delays

Can you please update this patch according to the Guidelines ( https://github.com/publicsuffix/list/wiki/Guidelines )

Specifically:
* Rebase to master
* Remove the date
* Indicate the means to be used to authenticate this request (ideally, the TXT record)

goipde:
- added a TXT record
- rebased
- removed date
- resorted

public_suffix_list.dat
```diff
@@ -11174,6 +11174,10 @@ gist.githubcloud.com
 // Submitted by Zoltan Egresi <egresi@globehosting.com>
 ro.com
 
+// GoIP DNS Services : http://www.goip.de
```
sleevi: Update to HTTPS

sleevi:
Validated via TXT. Ideally the link would be HTTPS, but a quick scan shows we're not requiring that yet ;)

PR #90, opened 2015-12-23
Add compute.estate and alces.network for Alces Software Ltd

mjtko:
These two domains are used to provide subdomains for different users.

Confirmation may be sought at the admin email address or registrant email address for these domains.

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -10621,6 +10621,11 @@ zuerich
 // ===BEGIN PRIVATE DOMAINS===
 // (Note: these are in alphabetical order by company name)
 
+// Alces Software Ltd: http://alces-software.com
+// Submitted by Mark J. Titorenko <mark.titorenko@alces-software.com> 2015-12-23
+*.compute.estate
+*.alces.network
+
 // Amazon CloudFront : https://aws.amazon.com/cloudfront/
 // Submitted by Donavan Miller <donavanm@amazon.com>
 cloudfront.net


sleevi:
Sorry for the delays, @mjtko 

Can you please update this patch according to https://github.com/publicsuffix/list/wiki/Guidelines

Specifically:
* Include the company domain name
* Sort by right-most label ( `estate` before `network`)
* Indicate the means to authenticate these requests (ideally, via DNS TXT records)

mjtko:
Hi @sleevi,

Thanks for getting back to me.  I've updated the PR as per the guidelines.  I've also updated the domains to include wildcards to reflect that cookies should not be set below level 4.

We've added TXT records for `_psl.compute.estate` and `_psl.alces.network` containing the URL of the PR.

sleevi:
"To reflect that cookies shot not be set below level 4" - that doesn't match up with what a wildcard indicates, but perhaps we're simply communicating wrong.

Could you describe more about how these domains are used and your purpose for adding them to the PSL? Including example domains and use? That would help us validate that the rules are correct and matching intent.

mjtko:
Hi @sleevi,

The domains are used to provide different web services to different customers on domains such as:

customer.division.compute.estate
www.customer.division.compute.estate

We're wanting each 'customer level' to be treated separately.

sleevi:
OK, thanks for explaining. In that case, the wildcard rules look correct - the public suffix will be `division.compute.estate`, and thus `customer1` vs `customer2` will be treated as different registrable domains, while `www.customer1` will be part of `customer1`'s public suffix.

sleevi:
Confirmed via TXT for both.

sleevi:
D'oh, forgot to remove date.

mjtko:
Thanks @sleevi!

dnsguru:
@mjtko will these be renewing?

mjtko:
Hi @dnsguru - yes, they will be renewed before expiry (12th December).

dnsguru:
Outstanding.  Hope you are well and healthy in these strange times.

As an update, we've been requiring 2 years minimum registration term
remaining on new submissions, and due to the amount of 'set-and-forget'
entries and growing file size, we are looking to automation to cull names,
which will be based upon the days until renewal being under 1 year (this is
what flagged your names, I swear I am not a stalker =] ).

Because that stuff will be automated, I recommend you to consider renewing
>1 year to avoid getting caught by that logic.

mjtko:
@dnsguru Ok, thanks for the heads up. I'll deal with the renewals over the next few days. It's likely that we'll be renewing for a further 5 years, and I'll set myself a just-under-4-years-into-the-future reminder to renew again to prevent the domain names from being culled from the list.

Best wishes, and many thanks for your continued work on the public suffix list.

groundcat:
Hello @mjtko Is the project still active? If the service has been decommissioned and the domain is no longer linked to an active project, as outlined in your original request, please confirm whether it can be safely removed from the PSL. If your project or organization is still active and requires continued inclusion in the PSL, please also reply at your earliest convenience. If we do not hear back from you for an extended period (over 30 days), your entry may be subject to removal.

PR #107, opened 2016-01-18
adding dynv6 dynamic DNS service

dmke:
Added dynv6.net as public suffix for the dynv6.com service.

Will update a TXT record pointing to this PR URI for verification. Feel free to contact me at dom@digineo.de.

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -11014,6 +11014,10 @@ webhop.org
 worse-than.tv
 writesthisblog.com
 
+// dynv6 : https://dynv6.com
+// Submitted by Dominik Menke <dom@digineo.de> 2016-01-18
+dynv6.net
+
 // EU.org https://eu.org/
 // Submitted by Pierre Beyssac <hostmaster@eu.org> 2015-04-17
 


dmke:
Here's the record:

```
$ dig -t TXT dynv6.net +short
"https://github.com/publicsuffix/list/pull/107"
```

(it may take some time to propagate)

dmke:
So, how's it coming? What can I do to speedup the process?

weppos:
We are currently working on improving our approval process. Please be patient.

Also note that if your request is specifically related to Let's Encrypt, being merged will not immediately provide you any advantage regarding rate-limits as the list will still have to be pulled on their side and there are discussions about improving that mechanism either.

Therefore, if your interested is just connected to Let's Encrypt, you get better chance to wait for them to increase their limits (which is what they will do at some point) or provide a way to grant specific permissions to selected domains.

Conversely, if you have an interest to be in the list because of the PSL goal itself and you are fully aware of the implications, then we'll be happy to review your request as soon as possible.

I appreciate your patience.

dmke:
Thanks for your reply. I was just making sure this project isn't abandoned.

We are indeed interested in circumventing the current LE rate limitation, but the larger goal of this project is valuable, too, of course :-)

gerv:
This seems fine to me. The DNS record is there as promised. weppos - let me know if there's more checking I should have done.

Gerv

PR #110, opened 2016-01-22
Add apps.fbsbx.com to a new section for Facebook, Inc.

fmoo:
Feel free to reach out to me at ruibalp@fb.com with any questions.


fmoo:
Anything else you need from me?  I've confirmed the email alias and everything checks out from our end.

Issue #124, opened 2016-02-02
.PRO missing some entries


sherwinflight:
The list is missing some entries for the .PRO TLD.

They are: recht.pro, aaa.pro, acct.pro, and avocat.pro

They can be seen on the Wikipedia page for .PRO: https://en.wikipedia.org/wiki/.pro

weppos:
More context http://registry.pro/get-pro

weppos:
I opened a PR at https://github.com/publicsuffix/list/pull/126

PR #126, opened 2016-02-02
pro: Add new suffixes

weppos:
Validated using http://registry.pro/get-pro

> Select third level .PRO domains are also available including:
> law.pro, avocat.pro, bar.pro, recht.pro, jur.pro, cpa.pro,
> aaa.pro, aca.pro, acct.pro, med.pro, eng.pro.

/cc @sherwinflight

---

Closes GH-124

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -5577,15 +5577,19 @@ est.pr
 prof.pr
 ac.pr
 
-// pro : http://www.nic.pro/support_faq.htm
+// pro : http://registry.pro/get-pro
 pro
+aaa.pro
 aca.pro
+acct.pro
+avocat.pro
 bar.pro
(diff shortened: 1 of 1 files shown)


gerv:
It seems like that list is not complete ("including:"). Should we drop them a line and suggest they give us a complete list?

But r=gerv for this change in the mean time.

weppos:
@gerv I don't have any afilias contact and the .PRO was added in 2007. I've sent an email to them, I used the email address listed on the site.

I'll hold off the merge for one more day, in case they should respond to my email.

weppos:
The response I got:

> Thank you for contacting .PRO Support.
> The one's listed are only second level zones we offer for .PRO.
> http://registry.pro/get-pro

PR #127, opened 2016-02-02
cryptonomic.net

afcady:


weppos:
@afcady are you affiliated with this domain?

afcady:
Yes, I just registered it.

weppos:
Please update the PR to add contact information. Also, if I can ask, what is the purpose of listing that specific domain you just registered in the list?

afcady:
Done.

The purpose of listing the domain is that I intend to delegate its subdomains (for free!) to the general public.  Kind of like dyndns.org, but it will be different in technical operation.

afcady:
Is there anything else I need to do?

sleevi:
Please update this patch according to the [Guidelines](https://github.com/publicsuffix/list/wiki/Guidelines)

Note: It's unlikely that you need both entries. The wildcard entry doesn't seem correct, based on your description. Could you explain more why you included it?

afcady:
I was originally going to issue domains that look like this:

    <cryptographic hash>.cryptonomic.net

E.g.:

    d545b24e96e644e2bc1e02e65751472425d37771.cryptonomic.net

Then I realized I might also want to issue domains that look like this:

    <cryptographic hash>.<hash type>.cryptonomic.net

E.g.:

       d545b24e96e644e2bc1e02e65751472425d37771.sha1.cryptonomic.net

I thought that having both entries would allow me to do either one, or both.  Is that not correct?

Thanks.

afcady:
Anyway, I added the `TXT` records:

    $ dig TXT _psl.cryptonomic.net +answer +short
    "https://github.com/publicsuffix/list/pull/127"

    $ dig TXT _psl.anything_else.cryptonomic.net  +answer +short
    "https://github.com/publicsuffix/list/pull/127"

afcady:
Was there something else besides those records?

sleevi:
@afcady 

1) Remove the date
2) Note how the format is <company name> : <web address> for the first line (preferably HTTPS)
3) The inclusion of both 'domain' and '*.domain' is known to create issues for a number of PSL implementations. As part of the integration tests, they properly flag this as problematic. Because of this, we're presently hesitant to include such a PR (as it would break these systems). Would it be possible for you to use your <hash>.<hash-alg>.cryptonomic.net system for the time being? I don't know if/when PSL implementations will be updated to handle the dual scenario here.

@weppos You approved the checks from @rockdaboot , is this accurate?

afcady:
OK.  Those changes are made.

afcady:
Bump...

Issue #109, opened 2016-01-21
Remove date from `Submitted by` label


weppos:
```
// Company : http://www.example.com/
// Submitted by John Doe <example@example.com> 2015-12-17
example.com
```

I propose to change it to

```
// Company : http://www.example.com/
// Submitted by John Doe <example@example.com>
example.com
```

I'm aware this comment field is unstructured, however it is becoming a kind of de-facto standard (especially for private suffixes) and I think we should try to find a rule.

The date is irrelevant, I can get it from git. The name and the email is still useful, especially to lookup the email in my account (or for future communications).

```
// Submitted by
```

can be kept as it is, or we can use a different word. We have used variants such as `Submitted`, `Requested`, `Updated`, etc. I propose to use just a single word that represents the last person who updated it. We can use

```
// Changed by John Doe <example@example.com>
```

that can potentially make sense also for multiple entries

```
// Changed by John Doe <example@example.com>
// Changed by Someone Else <example@example.com>
```

@gerv @sleevi feedback?

sleevi:
For John Doe / Someone Else, won't the git history have it? I can just see that part growing perhaps unbounded if we keep adding to it, since people who are authorized to make changes at time N may change positions or companies, and no longer be authorized to make changes. So only the 'last person wins', and the history of changes can be recovered via blame.

weppos:
> For John Doe / Someone Else, won't the git history have it?

I originally had the same impression, but after I tried to manage the list for a while without the comments, I must admit having the name is quite handy for quick lookups. In fact, the name on Git doesn't always match the submitter name.

> So only the 'last person wins', and the history of changes can be recovered via blame.

Makes sense to me, in fact I was not encouraging multiple listing (probably I was not clear enough in my proposal).

sleevi:
My remark about the Git history was more about when John Doe previously authorized domains X, Y, Z for organization Foo, and then Someone Else authorizes domain A for organization Foo.

I don't think we want
` // Changed by John Doe <john@example.com>
  // Changed by Someone Else <someone@example.com>`

What I was suggesting that if Someone Else starts doing stuff for Foo, we just have

` // Changed by Someone Else <someone@example.com>`

If a request comes back from John Doe (say, to add Domain B), the Git History will capture that John was previously (and thus presumed still) authorized. And then it would be updated back to

` // Changed by John Doe <john@example.com>`

That's what I meant by git history having it :)

weppos:
[...]

PR #133, opened 2016-02-05
Remove date from `Submitted by` label

weppos:
This is a trivial change. It closes GH-109 (I decided to keep Submitted instead of Changed, and I normalized all the existing entries).

/cc @gerv or @sleevi for cross-review, just in case I made some error by mistake.


sleevi:
Looks good to me - +1

weppos:
That was fast! Thanks @sleevi

PR #143, opened 2016-02-11
Add spdns.de Dynamic DNS Provider to list

nvsystems:


mommmomm:
Are you going to change the pull request or will there be a second one?

nvsystems:
I've now tried changing the pull request. I'm not quite sure if I did this right, I'm a total GitHub newbie.

weppos:
@mommmomm @nvsystems is any of you the owner of these domains?

nvsystems:
I'm not, and I don't think mommmomm is either. Do we need an official letter signed and stamped with 3 carbon copies? The fact that spdns is a dynamic DNS provider is not exactly top secret, and this list is supposed to include dynamic dns providers... or isn't it?

weppos:
That's not the point. Being on the list has several implications on how the clients will handle cookies and SSL certificates (among other things) about the domain, and only the owner has the right to take this decision.

nvsystems:
I've asked in the SecurePoint forum if they are ok with this. Maybe someone from them will comment here if they are ok with it.

Naugrimm:
As @nvsystems noticed, you normally separate the customer domains from the service domain itself. Sadly, we did not make this separation, when we created this service "just for ourselves" a few years ago.

We are currently transitioning the website/update-servers to another domain, but as long as this process is not completed, we hesitate to add the spdns.(de|org|eu) domains to the public suffix list.

I will check internally if we can at least add the other domains in a first step.

mommmomm:
Is there a rough timeframe for this transition to happen?

weppos:
I'm closing the PR for now until we receive further updates. @Naugrimm, assuming you have ownership on this entity, please open a new PR when and if you'll want to add the domain to the list.

Naugrimm:
@mommmomm We only talk about weeks here.
@weppos I will do that, thank you.

PR #161, opened 2016-02-29
Add ChangeIp Free Dynamic DNS

jpiedel:
Add ChangeIp Free Dynamic DNS


weppos:
@jpiedel do you work for ChangeIp?

jpiedel:
No. I'm just one of satisfied users.

I've extracted list of domains one can choose as free dyndns-like 
service. It's publicly accessible here: 
https://www.changeip.com/services/free-dynamic-dns/
I'm interested in extending your list with those sufixes, because 
letsencrypt uses your service to differentiate independent subdomains 
from SANs. And without entry in your registry all i.e. myddns.com users 
are treated as SANs of one service which is untrue.

---
with regards
Jakub

W dniu 2016-02-29 15:05, Simone Carletti napisał(a):

weppos:
@jpiedel unfortunately unless the request is approved or submitted by the domain owner, at the moment our policy is to not include the domains.

PR #162, opened 2016-02-29
Add Exosite.com [PS]aaS domains

anthonyrisinger:
These domains automatically provision certificates via Let's Encrypt for clients on Exosite's platform.

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -11096,6 +11096,11 @@ tr.eu.org
 uk.eu.org
 us.eu.org
 
+// Exosite LLC https://exosite.com/
+// Submitted by C Anthony Risinger <anthonyrisinger@exosite.com>
+exosite-dev.io
+exosite.io
+
 // Facebook, Inc.
 // Submitted by Peter Ruibal <public-suffix@fb.com>
 apps.fbsbx.com


anthonyrisinger:
There are many open issues for inclusion, some with varying degrees of validation and what appears to be varying degrees of success moving forward.

What exactly is the criteria? It's unclear to me whether or not I need to do anything further.

It's not an immediate concern, but in the coming months we fully expect to blow the (rather low) 10 cert limit per "domain", *instantly* (we could right now with just our own supporting platform services). We've been acquiring traditional certs in the interim even though we'd rather be further advancing a Let's Encrypt integration :-(

Please advise!

weppos:
@anthonyrisinger part of the reason we slowed down the approval of new PRs is because we are in the process of defining a more formal process. It hasn't been necessary so far, because the changes were mostly around the IANA TLDs (that essentially require the registry approval and they are easy to validate in terms of ownership).

Before LE, the list was approached by a very limited number of relateive large providers outside registries (e.g. Google, Amazon) and we handled the validation manually.

This process has proven to not scale very well with the recent rush of requests. We are also trying to determine what's the best way to proceed, given most requests are submitted with the only intent of bypassing LE limits.

I made clear more than once, that these kinds of requests should not be submitted, instead you should wait LE to waive the limits (they will do that once they will end the beta) or release the form they promised where you can request specific grants.

I'm sorry to hear that you're hitting the LE limits, but honestly this is not something that depends on us and the maintenance of this list is independent from the specific decision of LE to adopt the list for rate limiting.

As a personal note, LE it's still in beta and while in beta they will continue to have a strict rate limiting. You should be conscious of their decision, and do not rely on us as the vehicle to bypass such limits.

On our side, what I can tell you is that we're definitely not sleeping, just check the activity history to see how much work we did in the last weeks. We have a roadmap that doesn't necessarily fit into the LE one, and we'll do our best to continue evaluating the requests in the best interest of the PSL project itself.

anthonyrisinger:
Thanks for you comments @weppos! Much appreciated.

It was not my intention to imply stagnation, only that many LE-related requests were sitting idle. I now understand your role... I actually thought the PSL and LE were much more intertwined, so much as thinking LE spawned the PSL. I do agree this list is not appropriate for the probably billions of little SaaS-y things folks will want to build, but security by default is a thing *everyone* needs nowadays, so the massive influx is not surprising. Sorry about that :-(

I will consider your response with my team and try to make headway with LE directly on the limits. I'll keep this PR open, but I'd maybe suggest a very bold-color-strobing-marquee warning in your PR template reminding people that this is not appropriate for LE-requests? You mentioned making this clear before, maybe point people there? I try to do my homework prior to in limited time, but I did not encounter such text.

At the end of the day people want to use LE and everything I read prior to submission made it seem like getting on this list is the proper/expected thing to do.

weppos:
@anthonyrisinger I appreciate your understanding.

Let me just rephrase my warning for clarity. If your primary intent is to be in the list _because_ of Let's Encrypt, then you probably better wait for them to increase the limits. Instead, if you understand the real implications of being in this list and you still believe the purpose of the PSL fits your need, then it's fine to keep this PR here.

The PSL has been around for years, but a lot of people just suddenly discovered it because they believe it's the way to bypass PSL limits. What we want, is to make sure that when you request to be listed you fully understand the implications, as getting de-listed will not be as easy once browsers and other consumers will embed your domain in their packages.

anthonyrisinger:
Closing because we don't belong here.

PR #86, opened 2015-12-20
Add www.ddnss.de DynDNS Service http://www.ddnss.de

ddnss:
Thanks!

home-webserver.de
myhome-server.de
dyndns1.de

Please feel free to contact me (postmaster at ddnss.de)  for confirmation.

PR #73, opened 2015-12-10
Add DDNSS.de dynamic dns service.

3mb3dw0rk5:
Please add the domain names from the dynamic DNS service provider DDNSS.de.

PR #178, opened 2016-03-08
Update for ddnss.de

grasshide:
Please include this domain to the List. I'm not sure if I placed it right. If not, feel free to move it to the right position.
Why the request? I'm using ddnss.de since a long time. It is very reliable and works well with my router.
I recently discovered let's encrypt but unfortunately the rate limit is always reached as more and more people are using it.
I asked the owner of ddnss.de to add the service to this list. In return he asked me to do it on his behalf since his requests have been denied.
Please let me know the reason, should this request be denied as well.

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -10751,6 +10751,10 @@ cupcake.is
 cyon.link
 cyon.site
 
+// ddnss.de : https://ddnss.de/
+// Submitted on behalf of Robert Niedziela <webmaster@megacomputing.de>
+ddnss.de
+
 // DreamHost : http://www.dreamhost.com/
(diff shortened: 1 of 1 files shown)


weppos:
@grasshide we can only accept a request if submitted by the domain owners. I believe the PR you are talking about is #86. It was rejected because it contained a lot of changes not relevant with the specific request.

I'm closing this PR as there is already another duplicate #73.

ddnss:
weppos why the Close ?

weppos:
@ddnss I closed the PR as a duplicate of #73 and because the request was not opened from the owner of the domain.

As explained in #73, please do open a new PR, containing only and all the domains and properly following the format of the other private domains (hence clearly stating also a valid contact), and we'll take care of it.

PR #183, opened 2016-03-19
Update publix suffixes for *.ke

lab-at-nohl:
Updated list according to https://en.wikipedia.org/wiki/.ke and KeNIC

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -3420,8 +3420,17 @@ uenohara.yamanashi.jp
 yamanakako.yamanashi.jp
 yamanashi.yamanashi.jp
 
-// ke : http://www.kenic.or.ke/index.php?option=com_content&task=view&id=117&Itemid=145
-*.ke
+// ke : http://www.kenic.or.ke/index.php?option=com_content&task=view&id=117&Itemid=145 || https://en.wikipedia.org/wiki/.ke
+ke
+co.ke
+or.ke
+ne.ke
+go.ke
+ac.ke
+sc.ke
+me.ke
+mobi.ke
+info.ke
 
 // kg : http://www.domain.kg/dmn_n.html
 kg


weppos:
Thanks for the update @lab-at-nohl. Unfortunately, we can't really use wikipedia as a trusted sources in this case.

I tried to access the KeNIC website, but I'm stuck in an infinite redirect.

```
➜  ~ curl -i http://www.kenic.or.ke/index.php/en/\?option\=com_content\&view\=article\&id\=999
HTTP/1.1 302 Found
Date: Sun, 20 Mar 2016 03:57:41 GMT
Server: Apache/2.4.6 (CentOS) PHP/5.4.16
X-Powered-By: PHP/5.4.16
Set-Cookie: 21ddeb866254b9c481437954a857cdaf=34klphl8pv5uo7kls4skqnffp5; path=/
Set-Cookie: c28be2f2179a5f32e5fb2ebfc36f6d74=en-GB; expires=Mon, 20-Mar-2017 03:57:41 GMT; path=/
Location: /index.php?option=com_content&view=article&id=999
Content-Length: 0
Content-Type: text/html; charset=UTF-8
```

Is the URL in the patch working for you?

lab-at-nohl:
Got that. No, link doesn't work for me, too. I took the source as it was and added wikipedia. 

Please go to homepage http://www.kenic.or.ke/index.php/en/ and check drop down list at "Get Your .ke Now" - this is how KeNIC is organized. 

Your precaution is appreciated.

weppos:
If a link doesn't work, there is no point to keep it in the PR. I also have one more question: why did you add `.ke`? Did you get some insight that it's possible to register a domain at top level? Do you have a trusted confirmation for that?

lab-at-nohl:
No, I don't have. I received spam from a top level beneath .ke - but that may be spoofed. In the list there was *.ke before so I assumed.

sleevi:
@weppos I'm closing this based on https://github.com/publicsuffix/list/pull/183#issuecomment-198954187 - but please feel free to re-open if you disagree.

Issue #181, opened 2016-03-17
TLD xn--e1a4c is missing in PSL


erikvanderpoel:
The TLD xn--e1a4c appears to be missing in the PSL.

http://www.iana.org/domains/root/db/xn--e1a4c.html

sleevi:
Thanks! Would you like to submit a pull request?

(The root issue is the PSL got out of sync with the IANA RZD again; we may have other omissions or excessive inclusions)

erikvanderpoel:
Sorry, I am new to git. I was hoping that simply filing this bug would get the ball rolling on your side.

gerv:
I don't actually see this TLD in https://newgtlds.icann.org/newgtlds.csv, so I don't think this is a sync issue...

Gerv

weppos:
@gerv @sleevi `xn--e1a4c` (`.ею`) is not a new gTLD, it's a ccTLD.
https://www.iana.org/reports/2016/eu-report-20160118.html

That's why it's not in https://newgtlds.icann.org/newgtlds.csv
although it was delegated at the end of 2015.

You can also get confirmation here
http://www.iana.org/domains/root/db

```
.ею	country-code	EURid vzw/asbl
```

sleevi:
@weppos Right, I think Gerv and I were speaking past eachother. We sync the ICANN gTLD DB, but don't periodically sync the IANA root zone database - which is why we also haven't picked up past IANA ccTLD removals. We need both feeds, and we're only pulling one now.

But yes, that's separate from this request, which is easy and correct to add.

gerv:
Ah! &lt;light dawns&gt;

weppos:
Added in #184 b6575c06d4b99ac102ec23cb6005ba52e37d345a

PR #184, opened 2016-03-19
Add TLD .xn--e1a4c

weppos:
Closes GH-181

/cc @sleevi @gerv for cross review


sleevi:
Looks good to me. Thanks, I was having trouble with my editors Unicode support.

PR #187, opened 2016-03-28
Update .ZM TLD

### LINKED ISSUE #140: .ZM TLD rule is restricting valid domains
The issue was raised today [in the Let's Encrypt forum](https://community.letsencrypt.org/t/issue-with-zm-tld/10593). Apparently, it's possible to register second-level `.ZM` domains such as `ischool.zm`. However, our current rule is `*.zm` that totally denies this possibility.

[According to Wikipedia](https://en.wikipedia.org/wiki/.zm) the .ZM TLD has several different suffixes. However, there is no official documentation from the registry, and the website is not providing any hint.

I'm going to contact the registry to get a list of officially approved suffixes, and to get more hints about second-level TLDs. Specifically, I'd like to know if they are exceptions, and in that case how long is the exception list.

If you have .ZM contacts, feel free to share.

/cc @gerv @sleevi

weppos:
After an email exchange with the registry, I got the list of official suffixes (#140).
It's not completely clear how many registered second level domains there are for the .ZM zone, therefore I went with listing all the suffixes _and_ the TLD itself.

/cc @gerv @sleevi for cross-review

---

Closes #140

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -6929,8 +6929,20 @@ school.za
 tm.za
 web.za
 
-// zm : https://en.wikipedia.org/wiki/.zm
-*.zm
+// zm : https://zicta.zm/
+// Submitted by registry <info@zicta.zm>
+zm
+zm.ac
+zm.biz
+zm.co
+zm.com
+zm.edu
+zm.gov
+zm.info
+zm.mil
+zm.net
+zm.org
+zm.sch
 
 // zw : https://en.wikipedia.org/wiki/.zw
 *.zw
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -6932,17 +6932,17 @@ web.za
 // zm : https://zicta.zm/
 // Submitted by registry <info@zicta.zm>
 zm
-zm.ac
-zm.biz
-zm.co
-zm.com
-zm.edu
-zm.gov
-zm.info
-zm.mil
-zm.net
-zm.org
-zm.sch
+ac.zm
+biz.zm
+co.zm
+com.zm
+edu.zm
+gov.zm
+info.zm
+mil.zm
+net.zm
+org.zm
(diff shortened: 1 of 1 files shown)


sleevi:
@weppos Shouldn't the domains outside of the .ZM zone be added to the private section? Since they're not true ICANN TLDs, but instead independent offerings by the registrar?

weppos:
@sleevi are you referring to a particular suffix? Those suffixes are offered by the .ZM registry, exactly as `.co.uk` or `com.ac`.
https://github.com/publicsuffix/list/blob/suffix-zm/public_suffix_list.dat#L7-L14

So far, we always listed them in the ICANN TLDs sections, when they were provided by the registry. Or have I misunderstood your comment?

sleevi:
@weppos `.co.uk` is subordinate to `.uk`, ,and `.com.ac` is subordinate to `.ac`. In these cases, the domains you're adding aren't subordinate to the ICANN ccTLD operated by the `.zm` registry, but in fact subordinate to independent registries. That is, there is no guarantee by `.com` that `.zm.com` will always be operated by the ccTLD `.zm` - it's just a product offering by the entity that happens to be contracted to run the `.zm` registry. To me, that seems like it's moreso for the PRIVATE section, because that's not part of the ccTLD delegation/administration, but just a registry that happened to register certain names and offer sub-delegation to potentially interested customers (e.g. the same as all the dynamic DNS providers, effectively)

weppos:
@sleevi actually, the problem is that the patch here is wrong. :sweat_smile:  The suffixes should be like `com.zm`, not `zm.com`. My fault.

I'll update the patch.

weppos:
@sleevi I updated the changeset. I'll squash and merge it if it's :+1: to you. Thanks for the cross-review.

sleevi:
Looks good to me

PR #188, opened 2016-03-29
Add ru.net as defined by nic.ru

Trozz:
nic.ru is now offerring .ru.net domains https://www.nic.ru/dns/domain/en/ru_net.html

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -5683,6 +5683,7 @@ int.ru
 net.ru
 org.ru
 pp.ru
+ru.net
 // Geographical domains
 adygeya.ru
 altai.ru


public_suffix_list.dat
```diff
@@ -5683,6 +5683,7 @@ int.ru
 net.ru
 org.ru
 pp.ru
+ru.net
```
sleevi: Note: This does not belong in this section, because it's not part of the .ru ccTLD. Instead, NIC.ru is selling subdomains from its own registration.

As such, NIC.ru would need to submit this change, to the PRIVATE section of the PSL.

sleevi:
Closing as Invalid, for the reasons explained in the above line note.

PR #189, opened 2016-03-31
Add free selfhost dyndns

frankenstein91:
+selfhost.eu

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -11058,6 +11058,9 @@ webhop.org
 worse-than.tv
 writesthisblog.com
 
+//selfhost : https://secure.selfhost.de
+selfhost.eu
+
 // dynv6 : https://dynv6.com
 // Submitted by Dominik Menke <dom@digineo.de> 2016-01-18
 dynv6.net


nexosone:
It is a free DDNS hoster, it's ok and save.

weppos:
Are you the owner of the domain?

frankenstein91:
No I'm not the owner of the domain. But it is needed for Let's encrypt.
Many German users are connected to this ddns hoster and can't create
certificates because of the domain limitation.
Am 31.03.2016 12:07 schrieb "Simone Carletti" <notifications@github.com>:

weppos:
I'm sorry, the list is independently managed from Let's Encrypt.
Our guidelines requires the request to come from the owner.
https://github.com/publicsuffix/list/wiki/Guidelines

Let's Encrypt is currently in beta. They have recently relaxed the limits, and they will continue doing it while working towards the general availability.

frankenstein91:
It's a no go the we must relay on a slow support for security. So please
contact them from your side if a approve is really needed. I think you
didn't speak with the owner of dotde.

You should support the security of the Web if some organization like let's
encrypt uses this list.

It's such a easy way to close and not helping.
Am 31.03.2016 13:32 schrieb "Simone Carletti" <notifications@github.com>:

nexosone:
We all want to improve the Let's encrypt project, so it will be helpfull to extend the supported DDNS hoster. Also it would be nice for the security of the Internet. In my opinion Frank requests the right way.
It is the wrong way, to wait for the support of a DDNS hoster.

weppos:
> We all want to improve the Let's encrypt project

The main issue in this discussion, is that _this_ is not the Let's Encrypt project. This is the public suffix list. If you want to improve Let's Encrypt, then you should move the discussion to the Let's Encrypt forums and issue tracker.

We don't plan to change our policy to please users of a particular service.

Moreover, you also have to understand that Let's Encrypt limit exists for a reason, and the service is still in beta.

PR #205, opened 2016-04-13
Add happyfuntimes.net to the list

greggman:
happyfuntimes generates dynamic dns subdomains per game to direct local players to the local games.

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -11251,6 +11251,11 @@ pagespeedmobilizer.com
 withgoogle.com
 withyoutube.com
 
+// Happyfuntimes : http://www.happyfuntimes.net
+// Submitted by Gregg Tavares <psl@greggman.com>
+happyfuntimes.net
+games.happyfuntimes.net
+
 // Hashbang : https://hashbang.sh
 hashbang.sh
 
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -11253,7 +11253,6 @@ withyoutube.com
 
 // Happyfuntimes : http://www.happyfuntimes.net
 // Submitted by Gregg Tavares <psl@greggman.com>
-happyfuntimes.net
 games.happyfuntimes.net
 
 // Hashbang : https://hashbang.sh


greggman:
dig TXT +short _psl.happyfuntimes.net 
"https://github.com/publicsuffix/list/pull/205"

sleevi:
Could you explain further why this needs to be on the PSL?

Simply generating dynamic subdomains does not require inclusion to the PSL. The primary purpose would be if you're allowing arbitrary user content AND that cookie theft would be an issue (for example, that you're also running other services on happyfuntimes.net, or that both games use cookies)

Please also see https://github.com/publicsuffix/list/wiki/Guidelines for the structure and format of the DNS records

sleevi:
Also, can you explain why both entries are needed? Why is games.happyfuntimes.net insufficient?

Our goal is to ensure users properly understand the PSL and its use cases, as we've received a number of pull requests that believe it's only for purposes of certificates; rather, it has to deal with security standboxing. Once added, it may take up to 18 months for it to be removed, and that may affect your services.

greggman:
It's actually security that's the issue. Chrome (and other browsers have or are in the process of banning features my project needs unless https). With my library users run games on their PC that serve webpages to phones they use as custom game controllers on the same LAN. 

Currently that happens by them going to http://<local-ip-address-of-machine-running-game>. Because Chrome has/is banning features that now has to be https://dynamicallygeneratedomain.games.happyfuntimes.net so I can get those features back. 

But, there's no way to get certs to do this. letsencrypt has a 20 cert a week limit but I have far more users than that. The project is open source so I don't have funds to buy certs for every user just out of the goodness of my heart. Letsencrypt uses PLS to raise those limits so I was directed here by people on the letsencrypt forum.

You're right about the happyfuntimes.net entry. Will remove

sleevi:
It sounds like your use case would be better addressed by a wildcard cert
or by discussing with Lets Encrypt. It does not seem like your use case,
bypassing Let's Encrypt limits, is a valid use case for the PSL.

greggman:
A wildcard cert won't work because sending the private key to 1000s of users is not ok.

sleevi:
Ah, thanks for clarifying that these domains will be under user control /
remotely hosted.

greggman:
So someone else suggested that since I will be giving away subdomains happyfuntimes.net belongs on the PLS period. Yes? No?

sleevi:
That is not necessarily a reason to be included on the PSL, no. I encourage you to review the documentation of the PSL on https://publicsuffix.org to better understand the uses.

greggman:
Can I ask, if I made a public service "freedomainswithcerts.org" where anyone that wanted a domain and a cert could click a button (or call an API) and get randomally generated domain with cert as in `<randomid>.freedomainswithcerts.org` would it get approved for the PSL. Would even be appropriate for the PLS?

There are times people need/want a cert but they don't want to pay for the domain and they aren't in a place where they can make their machine publically visible so that LE can ping them. 

This might solve my problems and other peoples and be more generic.

I'm not sure what kinds of abuse it might get. For my uses you'd call some https endpoint with "need a domain" and get returned the 2 domains and a cert

     internal.<randomid>.freedomainswithcerts.org // your internal IP address if you provided one
     external.<randomid>.freedomainswithcerts.org // your external IP address if you asked for one

I'm just running ideas. I can't be the only project that needs something like this. I saw some discussion in other thread about LE lifting their limit after beta but it's after beta now and they haven't lifted it. So, looking for other solutions. I'm happy to help find a solution that meets more than just my needs. I'm kind of lost on how to make this happen.

gerv:
> Can I ask, if I made a public service "freedomainswithcerts.org" where anyone that wanted a domain and a cert could click a button (or call an API) and get randomally generated domain with cert as in <randomid>.freedomainswithcerts.org would it get approved for the PSL.

Well, it would be a fairly bad idea, because you would be generating the keypairs for the person and so would know what their private key was.

sleevi:
> Would it be approved for the PSL?

Nothing's preventing this request from getting approved, other than requests are being processed in the order they're received (if you can't tell, there's a lot, and it's a very time consuming task for the group of volunteers it is; especially due to the backlog induced by Let's Encrypt using the PSL)

If Let's Encrypt supported wildcard certificates, you could do something like https://blog.filippo.io/how-plex-is-doing-https-for-all-its-users/ . You could still approximate that easily.

sleevi:
> Well, it would be a fairly bad idea, because you would be generating the keypairs for the person and so would know what their private key was.

No, that's not correct :)

greggman:
I don't see how a wildcard cert would help because I'd only get 1 cert right? And I'd be giving that cert out to 1000s of people (the private part). I thought Plex did it by partnering with a CA to make a cert for every user. Otherwise they wouldn't have needed to partner, just buy a wildcard cert.

sleevi:
@greggman Your understanding is correct; I was not suggesting a *single* wildcard certificate, but rather, a wildcard certificate for each of your customers. While it's well outside the remit of the PSL to provide certificate consultation services, your proposed approach `internal.<some-unique-user-id>.example.com` would suffer from DNS issues and caching, but in theory, would work. However, as noted by @gerv, you would need to have users themselves generate the certificates and keys, or else you would find those certificates revoked (sharing keys is prohibited by CAs). The `*.<some-unique-user-id>.example.com` would avoid the DNS caching issues.

greggman:
Got it. So I need 2 things from LE. I need no cert limits AND I need wildcard certs . ugh

sleevi:
Apologies, somehow this PR dropped to the floor. Based on the discussion, were you still interested in adding this to the PSL?

greggman:
Not at this time. This whole issue has effectively killed the project. It's too much work and will cost too much money to work around the issues

pinobatch:
In case greggman changes his mind and wants to reopen this later, let me explain the situation based on my understanding of what he's explained elsewhere:

A web server is set up on a private LAN in a home or in a museum or wherever. It acts as a matchmaker for a video game, and it can't be reached from outside the LAN. This server is given a subdomain `SOMETHING.games.happyfuntimes.net` by greggman's dynamic DNS service. Addition of `games.happyfuntimes.net` would make `SOMETHING.games.happyfuntimes.net` "registrable" in PSL lingo, with two desirable effects:

1. A matchmaking server operated by one entity (e.g. `foo12345.games.happyfuntimes.net`) won't be able to see cookies placed by someone else's matchmaking server (e.g. `bar67890.games.happyfuntimes.net`). This protects the privacy of the users' information associated with each server.
2. Each matchmaking server could request its own certificate from a CA that relies on the PSL.

Otherwise, each matchmaking server's operator would have to buy a domain or use a different dynamic DNS service that's already on the PSL.

greggman:
The reason I didn't reapply is I agree with Sleevi, my reason to be on the PSL is basically solely to get unlimited certs since I need a 1 or more certs per user. I agree that getting my domain on the PSL would solve that but I also agree with Sleevi it's arguably an abuse of the system. Ideally the solution should be scalable.

The only solution I could think of is make a non happyfuntimes solution like `iot-dns.com`, get that on the PSL maybe so anyone can use it for any open source project (maybe pay for non-open source). `iot-dns.com` would have some API to give you domains like `<usersExternalIp6or4addresssha256base64>.yourprojectname.iot-dns.com`.

Then, so as not to abuse letsencrypt I'd have to run my own separate CA (basically just clone letsencrypt). The only difference would be unlike letsencrypt I'd only issue certs for stuff under iot-dns.com so people hopefully wouldn't be inclined to abuse it (because they couldn't choose the domain).

But, becoming a CA sounds like too much work. All the auditing etc, begging to be added to OSes and browsers. And, also too costly, running all the servers etc. (not sure there's a market for any non-open source projects).

So I basically gave up. I wish supporting this type of project had a solution. Maybe if Google or Mozilla or the EFF or someone wants to fund it but as it is it was way too much of a commitment to just do as a hobby to keep my project going.

sleevi:
@greggman Note that many of my comments were wanting to make sure to understand the need and to advise on how best to structure in a way that works with the PSL, not necessarily a rejection.

Note that running a CA is likely a $2-$4 million investment over the first year, and unless you find a cross-signer, may not be viable for many years. Of course, because you're using for a specific name, you could look to get a name-constrained sub-CA or managed sub-CA, which limits certificates to just your domain. That relaxes the requirements for running a CA, and may only be a few hundred thousand a year.

Regardless of your CA choice, if you do end up offering subdomains to users with user controlled content that you need to isolate, adding yourself to the PSL is going to be a good choice, and we're happy to help advise on how best to structure those records.

Note that LE has updated its rate limit policy at https://letsencrypt.org/docs/rate-limits/ - you could consider reaching out via the form, which it sounds like you might need to do anyways because of the IP restrictions.

Anyways, it's not a dead end if you want to pursue it; my goal was finding how best to help you.

greggman:
Thanks but you've only confirmed what I said about. Open Source projects that want to do things in IoT with certs are **DEAD**. You just told me to do this will cost $2-$4 million. So yes you've confirm that basically there are now a new subclass of Open Source projects that 12 months ago were possible for free and today are impossible except with a seven figure bankroll.

Similarly thank you for the link the letsencrypt but it's frustrating everyone seems to forget what this entire thing is about. Updating to 100, 300, or 500 cert limit **IS NOT ENOUGH**. To support an open source project that might become popular with non-devs requires 10s of thousands of certs.

sleevi:
I tried to be helpful, as the link discusses how to contact them for greater limits, but it seems you've reached your own conclusions and ignored what I said.

I'm sorry you feel your project is not viable, although I will note a number of options still exist for you. But it does seem you've settled on architecture decisions, and since those aren't supported, will have a bad time.

greggman:
Let me try to make it clearer since my software seems to cause so much mis-understanding

Let's say I wanted to add Plex like streaming support to [VLC](http://www.videolan.org/vlc/index.html). VLC already has a web server built in so this is not far fetched. The idea being you run VLC and then from any device in your house that has a web browser you can stream movies fullscreen from the machine running VLC. 

In September 2015 it was possible to do that. As of December 2015 the browsers requires HTTPS to go fullscreen. HTTPS requires a cert. There were 71 million downloads of the latest version of VLC. In only 0.1% used the "stream to browser fullscreen" feature that would be 71k certs needed, more if the feature became more popular.

You said there are solutions, what would the solution for VLC be? I'm not seeing one above.

My project is no different except I don't have 71 million downloads a month. But my project is a library used in multiple games. It needs 1 cert per game per user. If any one game becomes popular it could easily require 10s of thousands of certs.

sleevi:
@greggman There's no misunderstanding of what you're asking, but I'm suggesting solutions exist that don't require the same pain you're complaining about. However, this is as productive as complaining about how unwieldy IP addresses are, and it's not fair to expect that users have to have a DNS server they talk to.

In any event, it sounds like this issue can remain closed.

pinobatch:
The grandma-friendly solution for VLC would be for VLC to support the API of one or more dynamic DNS services that are on the PSL.

greggman:
Thanks for that suggestion. I wouldn't call it grandma friendly. More like Grandma's hacker granddaughter friendly. Sign up for dynamic DNS, verify account, create API key, copy credentials into game, repeat for each and every game, run out of free domains, find new dynamic DNS service, repeat.

Anyway, this discussion should be taken to https://github.com/greggman/HappyFunTimes/issues/20 if you want to continue discussing solutions

PR #215, opened 2016-05-02
Suffix gls.slg.br

rhuano:
.


weppos:
@rhuano please check the patch. It seems to contain only two changes I made, there is no reference to anything related to the title of this PR.

PR #216, opened 2016-05-03
Add greatnet.de, greatnet.com

jomat:
Customers get a subdomain to use their webspace and servers without the need to register an own domain.


weppos:
@jomat before I move forward, as you aware that if the patch will be merged your primary domain may not work anymore in browsers or other clients using the PSL, because of the fact you requested it to be treated as a suffix `greatnet.de`?

sleevi:
Closing due to no response.

PR #223, opened 2016-05-14
Neocities: add neocities.org

kyledrake:
This adds neocities.org to the public suffix list.

Neocities allows anyone to run arbitrary code on our subdomains, and we enforce a strict cookie separation between neocities.org and *.neocities.org.

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -11330,6 +11330,10 @@ cloudapp.net
 // Submitted by glob <glob@mozilla.com>
 bmoattachments.org
 
+// Neocities : https://neocities.org/
+// Submitted by Kyle Drake <kyle@neocities.org>
+neocities.org
+
 // Neustar Inc.
 // Submitted by Trung Tran <Trung.Tran@neustar.biz>
 4u.com


kyledrake:
```
$ dig TXT _psl.neocities.org

; <<>> DiG 9.10.3-P4-Ubuntu <<>> TXT _psl.neocities.org
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 46921

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 512
;; QUESTION SECTION:
;_psl.neocities.org.		IN	TXT

;; ANSWER SECTION:
_psl.neocities.org.	300	IN	TXT	"https://github.com/publicsuffix/list/pull/223"

;; Query time: 37 msec
;; SERVER: 127.0.1.1#53(127.0.1.1)
;; WHEN: Fri May 13 21:44:27 EDT 2016
;; MSG SIZE  rcvd: 105

```

kyledrake:
After discussions, it's been determined that merging this into the public suffix would probably prevent Neocities from using cookies on the root domain.

I might in the future decide to use www.neocities.org and then this would likely work, but I kindof like using the root domain for the main site as a structure, so I'll close this for now and maybe come back to it in the future.

da2x:
Did you ever revisit this, @kyledrake?

kyledrake:
I would love to do this, but I wasn't able to guarantee that it wouldn't ruin the root domain so I didn't do it.

freddyb:
Just commenting here in light of the [neocities block from bing search engines](https://arstechnica.com/tech-policy/2026/02/neocities-founder-stuck-in-chatbot-hell-after-bing-blocked-1-5m-sites/). 

Putting the www-domain on the PSL would ensure that a search engine (among others) would block whatever site they find undesirable rather than the whole *.neocities.org namespace.

@kyledrake

simon-friedberger:
Yes, but it would also mean that you cannot send mail with a valid DMARC organization anymore. But if mail is not terribly important or can be moved elsewhere, it is probably worth adding.

PR #225, opened 2016-05-19
Add certmgr.org

blechschmidt:
certmgr.org is supposed to become an aiding domain for server management. Any server operator is supposed to be able to obtain a subdomain *.certmgr.org for his server pointing back to the server under the operator's control. In order to avoid the misuse of content on foreign subdomains, strict cookie separation is required.

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -10757,6 +10757,7 @@ co.com
 c.la
 
 // certmgr.org : https://certmgr.org
+// Submitted by B. Blechschmidt <hostmaster@certmgr.org>
 certmgr.org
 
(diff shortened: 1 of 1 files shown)


blechschmidt:
```
$ dig +short TXT _psl.certmgr.org
"https://github.com/publicsuffix/list/pull/225"
```

weppos:
@blechschmidt before I move forward, are you aware that if the patch will be merged your primary domain may not work anymore in browsers or other clients using the PSL, because of the fact you requested it to be treated as a suffix `certmgr.org`?

blechschmidt:
@weppos yes, I am aware of that and this is okay.

PR #226, opened 2016-05-22
Update .JP as requested by info@jprs.jp

weppos:
Removed `kesennuma.miyagi.jp`.


weppos:
Confirmed by registry.

PR #227, opened 2016-05-23
Added 0emm.com to Google's list

scottwug:
Hi Eduardo (@sirdarckcat),
This request is to add the root 0emm.com as well.
If *.0emm.com added last time already implied its root domain, please just ignore.

Thank you very much!
Scott Wu

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -11217,6 +11217,7 @@ goip.de
 
 // Google, Inc.
 // Submitted by Eduardo Vela <evn@google.com>
+0emm.com
 *.0emm.com
 appspot.com
 blogspot.ae


sleevi:
please explain why this is necessary.

From the documentation I have seen, it is not. Given the PSL is used in a wide variety of ways, there is a desire to keep it as minimal as is reasonable.

scottwug:
You are right, from the document this should be implied. But we are not very clear about this, just in case for double check.

sleevi:
As you can see, this PR creates conflicts for applications.

Since it is not necessary from the Google documentation I've seen, nor is it well supported (*.foo and foo both being added), in going to close this as WontFix. Provided you stick to your documented plan of x.y.0emm.com, you will have no issues.

scottwug:
Cool, just want to confirm that x.y.0emm.com cannot put cookie on 0emm.com, is this true?

sleevi:
Implementation defined due to historical ambiguities, but generally so across the platforms you care about.

scottwug:
Great!
Thank you for your quick response, we really appreciate your help!

PR #233, opened 2016-05-27
callidomus: please add mycd.eu

mknx:
callidomus is providing it's customers access to their homes with dyndns and https (with let's encrypt cert).
They are using XXXX.mycd.eu to access their homes.

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -10701,6 +10701,10 @@ myfritz.net
 // Submitted by Adrian <adrian@betainabox.com>
 betainabox.com
 
+// callidomus: https://www.callidomus.com/
+// Submitted by Marcus Popp <admin@callidomus.com>
+mycd.eu
+
 // CentralNic : http://www.centralnic.com/names/domains
 // Submitted by registry <gavin.brown@centralnic.com>
 ae.org


mknx:
$ dig +short TXT _psl.mycd.eu
"https://github.com/publicsuffix/list/pull/233"

weppos:
@mknx can you please rebase your patch against master?

groundcat:
Hello @mknx , it seems that the project or organization associated with your domain, which was merged into the Public Suffix List (PSL), may no longer be active based on the observable usage of subdomains. 

- mycd.eu 
  - no observable usage of subdomains
  - https://www.callidomus.com/ does not work
  - Email Marcus Popp <admin@callidomus.com> unreachable

If the service has been decommissioned and the domain is no longer linked to an active project, as outlined in your original request, please confirm whether it can be safely removed from the PSL. This will help prevent leaving unnecessary entries for PSL volunteers and future registrants, as requesters often "set and forget" and neglect to request the removal of their domains from the PSL when their project reaches its end of life (EOL).

The PSL is a globally used resource, and it's important to understand that tens of millions, if not hundreds of millions, of devices and users incorporate the list. We aim to keep it as minimal as possible for those who truly need PSL inclusion. Projects that are smaller in scale, such as those without thousands of users or subdomains belonging to thousands of individual entities, will likely be ineligible for inclusion. If you no longer need your entry to remain on the PSL, **please create a new pull request to remove it** or reply here to confirm.

If your project or organization is still active and requires continued inclusion in the PSL, please reply at your earliest convenience. If we do not hear back from you for an extended period, **your entry may be subject to removal**. 

Additionally, ensure that the required `_psl` TXT record is maintained, as specified in the [PSL Guidelines](https://github.com/publicsuffix/list/wiki/Guidelines#validation-of-submissions--always), and that your domain is kept renewed with an expiry date of more than 2 years at all times to avoid future removal.

Thank you.

PR #251, opened 2016-06-21
Add browsersafetymark.io to publicsuffix/list

dtharp:
2nd attempt.  Rebased to latest master.

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -10720,6 +10720,10 @@ betainabox.com
 // Submitted by Axel Fontaine <axel@boxfuse.com>
 boxfuse.io
 
+// callidomus: https://www.callidomus.com/
+// Submitted by Marcus Popp <admin@callidomus.com>
+mycd.eu
+
 // BrowserSafetyMark
 // Submitted by Dave Tharp <browsersafetymark.io@quicinc.com>
 browsersafetymark.io
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -10708,6 +10708,10 @@ betainabox.com
 // Submitted by Axel Fontaine <axel@boxfuse.com>
 boxfuse.io
 
+// BrowserSafetyMark
+// Submitted by Dave Tharp <browsersafetymark.io@quicinc.com>
+browsersafetymark.io
+
 // CentralNic : http://www.centralnic.com/names/domains
 // Submitted by registry <gavin.brown@centralnic.com>
 ae.org
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -10720,13 +10720,13 @@ betainabox.com
 // Submitted by Axel Fontaine <axel@boxfuse.com>
 boxfuse.io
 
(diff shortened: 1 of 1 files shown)


dtharp:
@weppos  Hey, I finally got the TXT record in place for psl.browsersafetymark.io. Can you please get this landed?  Thanks for your help :)

weppos:
@dtharp I was able to validate the request, but I can't merge the patch because it looks like your change is deleting another item in the list. See https://github.com/publicsuffix/list/pull/251/files

Can you please fix it?

dtharp:
@weppos: hmm.. not sure how that happened.  Something weird must have happened during the rebase. If you cycle through the patch chain, it all looks logical (as if that entry disappeared during the rebase). Anyway, I validated that the missing entry is indeed in the current list on your site (it is), and re-added it.  The patch chain is ugly, but the result is correct now...

weppos:
@dtharp there are still 2 issues:

- it must stay before Calidomus (alphabetical order), not after
- as requested per https://github.com/publicsuffix/list/wiki/Guidelines, please provide information of why you are requesting the domain to be added

dtharp:
@weppos: new commit resolves alphabetical order. 

As for why we are requesting:  The site is a privacy and security benchmark. We run a series of tests on the visitor's browser and rank the browser against other browsers that have run the benchmark.

One of the tests we run checks for Tracking Protection via several vectors (third party cookies etc.) We have worked with EFF on this feature, and the test is somewhat involved.  One of the factors we look for during the test is that an "ad" is served from a particular domain on at least 3 "first party" sites, and the "ad" is observed doing tracking behavior (setting a third-party cookie for instance). If this scenario is detected, and the browser is capable of protecting against this sort of tracking, we expect a 4th serve of that "ad" to be blocked.

Now, that is all pretty straightforward, but the devil is in the details.  One thing to consider is that a user may re-run the test...  The first run would pass, but if we used the same domain to serve the "ads", the "ad" would immediately be blocked on a subsequent run because the browser had already registered that domain as a tracker.  To get around this without being a Top Level Domain, we'd have to have an infinite number of standard domains serving our test ads.  This is clearly not an option.

Our solution is to make browsersafetymark.io a TLD. Then, all we need to do is serve the our test "ads" from a subdomain of browsersafetymark.io.  If browsersafetymark.io is a TLD, then third-party1.browsersafetymark.io is seen as a unique domain in the Tracking detection software in the browser.  What we have actually done is set up a wildcard DNS entry that points to our test ad server, and we randomize the subdomain we hit for each test.  In this way we do actually have an unlimited number of domains to serve our test "ads" from.

This has been tested in a chrome derivative build where we insert the contents of this patch into chrome's working copy and build.  The reason we need this in the actual public list is because we want other implementations of Tracking Protection (i.e EFF's Privacy Badger extension) to pass our test. Without an entry in the public list, any new tracking protection implementations would fail our test.

Hope this is clear. Let me know if you have questions.

dtharp:
@weppos  Is there anything else I need to do to get this landed?

Issue #243, opened 2016-06-14
Remove Lib.de.us


ehmoore:
It appears that Chrome will not trust wildcard certificates off of domains listed under ICANN Domains. Lib.de.us is one of these domains. We need to get lib.de.us removed from this list.

sleevi:
Do you want it moved to the private section? Have you considered the
security implications of outright removing it?

ehmoore:
Dear Sleevi,

Thank you for your quick response.  We have had great difficulty getting the Chrome browser to trust wild certificates created by Entrust for our domain.  This issue has dragged for months with the final recommendation from Entrust to remove our name from this list.
Please forgive my ignorance but I am not aware of the security implications behind complete removal and I would certainly appreciate your input on this matter.  It is our belief that being on this list has relegated our wildcard certificates to untrusted within Chrome.
What would be the benefit to just moving our domain listing to private?

Best Regards,

Ed

Edward Moore
Library Technologist
Tel: +1(302) 241 2992
Cel:+1(856) 465-8141
SFB: ed.moore@lib.de.us
ed.moore@lib.de.us<mailto:ed.moore@lib.de.us>

[cid:image002.png@01CF8AEF.6152D8F0]

sleevi:
While it's true that Chrome (and possibly Firefox?) will not respect wildcards for names in the ICANN section (e.g. you don't want a certificate for `*.com` to be used to attack any website), it does allow them in the PRIVATE section.

The biggest security implication to wholesale removal is that anyone would be able to set cookies for `lib.de.us` at that domain level. If subdomains are controlled by separate entities (e.g. `foo.lib.de.us` should not be able to control content at `bar.lib.de.us`), then that represents a security risk, as `foo` could maliciously set a cookie on `lib.de.us` that could allow foo to compromise `bar.lib.de.us`. You can read a bit more about this at https://publicsuffix.org/learn/

[...]

PR #256, opened 2016-07-01
Move lib.de.us to the private section

weppos:
> This email is to authenticate the request to remove lib.de.us
> from the PUBLIC section of the public suffix list and to add
> it to the PRIVATE section only.
> We need this done because Chrome does not respect wildcards
> for names in the ICANN (public) section, but they do in the
> PRIVATE section.

Closes https://github.com/publicsuffix/list/issues/243

/cc @sleevi @gerv can you please cross-review?
/cc @ehmoore

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -6526,7 +6526,7 @@ lib.ca.us
 lib.co.us
 lib.ct.us
 lib.dc.us
-lib.de.us
+// lib.de.us - Moved to Private section at request of Ed Moore <Ed.Moore@lib.de.us>
 lib.fl.us
 lib.ga.us
 lib.gu.us
@@ -11652,6 +11652,10 @@ hk.com
 hk.org
 ltd.hk
 inc.hk
+
+// .US
+// Submitted by Ed Moore <Ed.Moore@lib.de.us>
+lib.de.us
 
 // Viprinet Europe GmbH : http://www.viprinet.com
 // Submitted by Simon Kissel <hostmaster@viprinet.com>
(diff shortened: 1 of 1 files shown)


weppos:
Ping for @sleevi

public_suffix_list.dat
```diff
@@ -6526,7 +6526,7 @@ lib.ca.us
 lib.co.us
 lib.ct.us
 lib.dc.us
-lib.de.us
+// lib.de.us - Moved to Private section at request of Ed Moore <Ed.Moore@lib.de.us>
```
sleevi: For consistency with the other removals (e.g. 6463, 6454, 6423, etc) seems like this 
1) Should use two spaces after the entry (omit the -)
2) Include the Issue # that requested this ( https://github.com/publicsuffix/list/issues/243 - presumably "Issue #243" in code)
3) followed by -

that is
// lib.de.us  Issue #243 - Moved to PRIVATE at request of Ed Moore <Ed.Moore@lib.de.us>

wdyt?

sleevi:
Left an inline comment. This LGTM with some slight tweaks

ehmoore:
Hi Sleevi,
Is there anything more that we need to do to have this request executed?

gerv:
@ehmoore Looks like you need to do the stuff in Sleevi's earlier comment?

sleevi:
@gerv @weppos did the patch, not @ehmoore - just waiting for Simone's tweaks :)

weppos:
@sleevi @gerv I will be travelling for the next week, so I won't be able to handle it before the end of August. Feel free to take over, if you need/want.

Issue #145, opened 2016-02-14
Clarify number and position of wildcard labels


weppos:
I made a commit a few weeks ago that introduced a rule like `*.*.private.domain` and the commit [caused the build to fail](https://travis-ci.org/publicsuffix/list/builds/105548386).

[According to our website](https://publicsuffix.org/list/), that is supposed to be a valid format:

> Wildcards are not restricted to appear only in the leftmost position, but they must wildcard an entire label. (I.e. _._.foo is a valid rule: *bar.foo is not.)

@rockdaboot [mentioned](https://github.com/publicsuffix/list/pull/123#issuecomment-177281704) a potential incompatibility of Chromium if we allow multiple wildcards. `libpsl` is currently not compatible with multiple wildcards, and to be fair [I haven't tested my Ruby implementation either](https://rubygems.org/gems/public_suffix).

@gerv @sleevi can we clarify whether multiple wildcard labels are accepted? Specifically, we should be more clear if the following rules are valid:

```
// multiple leading wildcards (common case)
*.*.foo.bar

// single wildcard, but inside the rule
foo.*.bar

// multiple willdcards, inside the rule
foo.*.*.bar

// multiple willdcards, inside the rule, non consecutive
foo.*.bar.*.baz

// I suppose this is invalid
foo.*
```

The current list definition doesn't explicitly deny these rules, they are supposed to be valid.

Once the decision is taken, I think we should:
1. Update the website to make it more clear
2. Add some corresponding tests in the test file
3. Should the rules be considered invalid, I'll also add the corresponding rules to the [linter I'm working on](https://github.com/weppos/pslint), 
4. Should the rules be considered valid, we should ping the various lib maintainers

rockdaboot:
> // single wildcard, but inside the rule
foo.*.bar

This would not allow exceptions right now, since (AFAIR) the exclamation mark has to be the first character (and must not be followed by a dot). 
If this form will be allowed, the rules for exceptions must IMO change (e.g. foo.!baz.bar).

gerv:
I believe the Mozilla implementation also didn't allow multiple wildcards. It seems like there are lots of implementations out there which don't. Given that their use is quite niche (what was the use case for the rule you added?), perhaps we need to bow to the inevitable and update the spec to match reality.

sleevi:
Right, the Chrome code doesn't presently, but we could likely fix that. But in the implementations I've seen, I haven't seen any support for multi-wildcards, despite what the website says.

[...]

Issue #105, opened 2016-01-15
AWS ap-northeast-2 for compute and S3 to Private Domains


lawells:
Please add the following domains:

ap-northeast-2.compute.amazonaws.com
s3.ap-northeast-2.amazonaws.com
s3-ap-northeast-2.amazonaws.com

http://docs.aws.amazon.com/general/latest/gr/rande.html#s3_region

<pre>// Amazon Elastic Compute Cloud: https://aws.amazon.com/ec2/
// Submitted by Luke Wells <lawells@amazon.com> 2016-02-05
ap-northeast-1.compute.amazonaws.com
ap-northeast-2.compute.amazonaws.com
ap-southeast-1.compute.amazonaws.com
ap-southeast-2.compute.amazonaws.com
cn-north-1.compute.amazonaws.cn
compute.amazonaws.cn
compute.amazonaws.com
compute-1.amazonaws.com
eu-west-1.compute.amazonaws.com
eu-central-1.compute.amazonaws.com
sa-east-1.compute.amazonaws.com
us-east-1.amazonaws.com
us-gov-west-1.compute.amazonaws.com
us-west-1.compute.amazonaws.com
us-west-2.compute.amazonaws.com
z-1.compute-1.amazonaws.com
z-2.compute-1.amazonaws.com</pre>

<pre>// Amazon S3 : https://aws.amazon.com/s3/
// Submitted by Luke Wells <lawells@amazon.com> 2016-02-05
s3.amazonaws.com
s3-ap-northeast-1.amazonaws.com
s3.ap-northeast-2.amazonaws.com
s3-ap-northeast-2.amazonaws.com
s3-ap-southeast-1.amazonaws.com
s3-ap-southeast-2.amazonaws.com
s3-external-1.amazonaws.com
s3-external-2.amazonaws.com
s3-fips-us-gov-west-1.amazonaws.com
s3-eu-central-1.amazonaws.com
s3-eu-west-1.amazonaws.com
s3-sa-east-1.amazonaws.com
s3-us-gov-west-1.amazonaws.com
s3-us-west-1.amazonaws.com
s3-us-west-2.amazonaws.com
s3.cn-north-1.amazonaws.com.cn
s3.eu-central-1.amazonaws.com</pre>

lawells:
Hi,
Any chance we can have someone make this addition?
Thanks!

weppos:
@lawells do you work at Amazon or are you affiliated with them?

lawells:
[...]

PR #259, opened 2016-07-06
AWS ap-south-1 EC2 and S3 domains

lawells:
*Adding ap-south-1.compute.amazonaws.com EC2 domain
*Adding s3-ap-south-1.amazonaws.com and s3.ap-south-1.amazonaws.com S3 domains

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -10661,52 +10661,66 @@ beep.pl
 cloudfront.net
 
 // Amazon Elastic Compute Cloud: https://aws.amazon.com/ec2/
-// Submitted by Philip Allchin <pallchin@amazon.com>
-compute.amazonaws.com
-ap-northeast-1.compute.amazonaws.com
-ap-northeast-2.compute.amazonaws.com
-ap-southeast-1.compute.amazonaws.com
-ap-southeast-2.compute.amazonaws.com
-eu-central-1.compute.amazonaws.com
-eu-west-1.compute.amazonaws.com
-sa-east-1.compute.amazonaws.com
-us-gov-west-1.compute.amazonaws.com
-us-west-1.compute.amazonaws.com
-us-west-2.compute.amazonaws.com
-compute-1.amazonaws.com
-z-1.compute-1.amazonaws.com
-z-2.compute-1.amazonaws.com
+// Submitted by Luke Wells <lawells@amazon.com>
+*.compute.amazonaws.com
+*.compute-1.amazonaws.com
+*.compute.amazonaws.com.cn
 us-east-1.amazonaws.com
-compute.amazonaws.com.cn
-cn-north-1.compute.amazonaws.com.cn
 
 // Amazon Elastic Beanstalk : https://aws.amazon.com/elasticbeanstalk/
-// Submitted by Adam Stein <astein@amazon.com>
-elasticbeanstalk.com
+// Submitted by Luke Wells <lawells@amazon.com>
+elasticbeanstalk.cn-north-1.amazonaws.com.cn
+*.elasticbeanstalk.com
 
 // Amazon Elastic Load Balancing : https://aws.amazon.com/elasticloadbalancing/
-// Submitted by Scott Vidmar <svidmar@amazon.com>
-elb.amazonaws.com
+// Submitted by Luke Wells <lawells@amazon.com>
+*.elb.amazonaws.com
+*.elb.amazonaws.com.cn
 
 // Amazon S3 : https://aws.amazon.com/s3/
 // Submitted by Luke Wells <lawells@amazon.com>
 s3.amazonaws.com
 s3-ap-northeast-1.amazonaws.com
 s3-ap-northeast-2.amazonaws.com
+s3-ap-south-1.amazonaws.com
 s3-ap-southeast-1.amazonaws.com
 s3-ap-southeast-2.amazonaws.com
 s3-eu-central-1.amazonaws.com
 s3-eu-west-1.amazonaws.com
 s3-external-1.amazonaws.com
-s3-external-2.amazonaws.com
 s3-fips-us-gov-west-1.amazonaws.com
 s3-sa-east-1.amazonaws.com
 s3-us-gov-west-1.amazonaws.com
+s3-us-east-2.amazonaws.com
 s3-us-west-1.amazonaws.com
 s3-us-west-2.amazonaws.com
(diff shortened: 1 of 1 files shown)


sleevi:
@lawells - Would Amazon be willing to consider `*.amazonaws.com` and `*.*.amazonaws.com` (or `*.compute.amazonaws.com` ) ?

I'm wondering if that might offer a more scalable, efficient means of handling these additions. It does cause some limits on what the subdomains you can have under amazonaws, but would help you launch quicker and with fewer risks.

weppos:
@sleevi per https://github.com/publicsuffix/list/issues/145 `*.*.amazonaws.com` is currently formally valid, but practically not supported. Unless I missed something.

sleevi:
@weppos That's why I suggested `*.compute.amazonaws.com` :) Trying to figure out if there's a more general expression of the hierarchy of suffices, which could prevent churn and also help reduce the PSL size

lawells:
@sleevi, let me talk it over with a couple folks. re: *.compute.amazonaws.com, if we still have to add S3 subdomains then it's not really that much net gain for the PSL or Amazon since we're still having pull requests for each new region. @weppos from #145  it also looks like s3.*.amazonaws.com would not work either?

sleevi:
I'm guessing formatting screwed up for @lawells 

The benefit of `*.compute.amazonaws.com` is that it reduces the number of distinct entities (and can help browsers behave slightly faster if they're using DAFSA/Tries rather than perfect hashes), even if you'd still need to add entries for distinct `.amazonaws.com` entries.

Note that `*.compute.amazonaws.com` and `*.amazonaws.com` are both valid entries (and can be used together), indicating everything under compute is a public suffix, but also anything under amazonaws.com is as well. The only reason you would *not* want to do this is if you wanted `foo.amazonaws.com` and `bar.amazonaws.com` to share cookies/other security state.

lawells:
@sleevi I've got the question out to a couple folks. will keep you posted...

sleevi:
@lawells Would `*.elb.amazonaws.com` do what you need instead?

lawells:
@sleevi @weppos AWS would like to keep regional domains instead of wildcarding for the following reasons:
1. Our goal is to have regional independence, meaning that eu-west-1 should not affect us-east-1, whereas *.compute.amazonaws.com, *.elb.amazonaws.com and *.amazonaws.com introduces the potential for sharing cookie state across multiple regions/subdomains.
2. Not all browsers honor the * syntax in a determinable way, e.g. see the Internet Explorer note in [https://devcenter.heroku.com/articles/cookies-and-herokuapp-com] 
3. If we were to introduce another subdomain layer in the future, e.g. "<new-aws-controlled-subprefix>.<region>.amazonaws.com" then we could be susceptible to broadly scoped cookies sharing state.

sleevi:
@lawells Regarding #1, you may be misunderstanding what the wildcard rule says.

`*.elb.amazonaws.com` means that `foo.elb.amazonaws.com and `bar.elb.amazonaws.com` are both public suffices, and thus can't share state, nor can their subdomains. That seems to align with exactly what you want.

Regarding #2, the issue is not with the wildcard syntax, the issue is that IE did not honor or use the PSL *at all*. Edge changed this, and now Microsoft is also consuming the PSL, moving away from their heuristic.

Regarding #3, I'm not sure what you're saying (it looks like your example domain was clipped), but the point of the wildcard is precisely so that you can add regions at will and not have to worry about updating the PSL. As you've currently done it, you're significantly more at risk in introducing new regional zones, because you cannot 'safely' do so until all PSL consumers have updated (which can take up to 18 months).

I'm pretty sure @pzb is tuned into the PSL issues, so maybe ya'll can chat with him internally and see if there's a strategy that gives Amazon the flexibility it needs, w/o adding risk. But I'm pretty sure everything you said you're concerned about is exactly the thing wildcards help mitigate, rather than cause :)

lawells:
@sleevi talking to @pzb now...

sleevi:
Any updates?

lawells:
@sleevi sorry for the delay...I think we finally have the best possible wildcarding case for AWS. EC2 and ELB are drastically simplified, however, S3 is still growing and will need to be updated in the future. Would be great to do:
<code>
s3.\*.amazonaws.com</code>
<code>
s3.dualstack.\*.amazonaws.com
</code>

With the above two, it would only leave the s3-ap-southeast-1.amazonaws.com domains to be updated.

sleevi:
Awesome, thanks for taking a pass on this.

I think it's unlikely we'd be able to add support for your S3 case in any reasonable time, and I think it'd cause a fair number of issues for clients (all, AFAICT, assume they can stop processing at a wildcard, and I think that's probably desirable). As such, it may be worthwhile to think about how you're going to expand S3 in the future, because it may be necessary to apply limits (to organizations and additions) in the coming future, to help mitigate growth.

@gerv @weppos Mind cross-checking and making sure we're happy with authorization?

lawells:
@sleevi @pzb @weppos @gerv Any thoughts?

sleevi:
Oh, sorry @lawells I had thought I merged this.

If you can rebase, I can merge either tonight or tomorrow :)

@weppos How do we want to authenticate? @pzb indicates that @lawells is a coworker :)

lawells:
@sleevi sorry about that, did a rebase - you should be good to merge!

weppos:
> @weppos How do we want to authenticate? @pzb indicates that @lawells is a coworker :)

@sleevi I personally authenticated him at
https://github.com/publicsuffix/list/issues/105#issuecomment-180260499

It's a 👍  for me.

sleevi:
@lawells There's still some conflicts with master (and perhaps we merged something after your rebase). I can try to manually fix it up tomorrow morning, or if you're able to get it clean from the web interface, I'll squash and merge sooner :)

lawells:
@sleevi should be good now

131:
I do not understand something.
I thought that *.compute-1.amazonaws.com beeing a level 4 tld, means that ec2-23-20-71-128.compute-1.amazonaws.com is a TLD, on ONLY a tld. And therefore cannot be used a a domain name alone.
Yet, this URL  : http://ec2-23-20-71-128.compute-1.amazonaws.com/news/uk-news/2020/04/14/life-on-the-inside-10-ways-to-ease-an-anxious-mind-during-lockdown/ 
Is somewhat.. invalid ?

Please help me understanding what's going on here

sleevi:
No, it doesn’t work that way, for the same reason as appspot.com or
github.io

PR #262, opened 2016-07-07
Add celtra-test.com

jakajancar:
Our developers and QA use `<developer>.celtra-test.com` for personal testing stacks. Cookies should not be shared and, most importantly, Lets Encrypt should tread them as separate registered domains.

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -11675,4 +11675,8 @@ yolasite.com
 za.net
 za.org
 
+// Celtra Inc.: http://www.celtra.com/
+// Submitted by Jaka Jancar <jaka@celtra.com>
+celtra-test.com
+
 // ===END PRIVATE DOMAINS===


jakajancar:
Verification:

```
$ dig txt _psl.celtra-test.com

; <<>> DiG 9.9.5-9+deb8u6-Debian <<>> txt _psl.celtra-test.com
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 30738

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 512
;; QUESTION SECTION:
;_psl.celtra-test.com.		IN	TXT

;; ANSWER SECTION:
_psl.celtra-test.com.	246	IN	TXT	"https://github.com/publicsuffix/list/pull/262"

;; Query time: 23 msec
;; SERVER: 8.8.8.8#53(8.8.8.8)
;; WHEN: Wed Jul 06 21:44:25 EDT 2016
;; MSG SIZE  rcvd: 107
```

weppos:
@jakajancar we'd like to avoid adding test domains to the list. If the domain is only used by your developers or QA, what's the need of cookie separation?

jakajancar:
@weppos It's not a problem, really, should have wrote "it's OK if cookies are not shared". The reason is being able to "register" domains quickly and for free (like FreeDNS, except under our own control) and be able to generate certs for them using Lets Encrypt.

jakajancar:
@weppos Can we get an update on this? I feel our case is aligned with the description in README:

> A "public suffix" is one under which Internet users can (or historically could) directly register names. Some examples of public suffixes are .com, .co.uk and pvt.k12.ma.us.

I can make registration open to public if there is really a requirement for "*all* Internet users" to be able to register (for us the goal is only not to rely on a 3rd party to register subdomains for us). However, this is not a rule that has been applied to many current entries (even the last in the README), and there are also entries which are honest about being testing ones (beta.bounty-full.com, townnews-staging.com).

sleevi:
@jakajancar The concern here is that a large number of applications would need to become aware of this, and thus carry the binary size (think every browser and network CLI that supports cookies).

While this is, admittedly, a limitation of the PSL, and not always applied, the increase in applications to the PSL have us trying to make sure that the requestor is well-informed of the operational cost they're asking the ecosystem (the Internet) to bear.

Given that Let's Encrypt has changed policies about Rate Limiting, if the only goal is for LE use cases, would you be willing to consider discussing with LE directly? See https://letsencrypt.org/docs/rate-limits/

weppos:
Closing as lack of follow up and because Let's Encrypt now offers a specific form for lift limits.

Moreover, the PR doesn't follow the [Guidelines](https://github.com/publicsuffix/list/wiki/Guidelines).

PR #264, opened 2016-07-08
add js.org

indus:
#249

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -11383,7 +11383,7 @@ iki.fi
 biz.at
 info.at
 
-JS.ORG : http://dns.js.org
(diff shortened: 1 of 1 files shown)


indus:
I´ve updated the TXT entry in the DNS zonefile but TTL will make it take some time to propagate.

public_suffix_list.dat
```diff
@@ -11383,6 +11383,10 @@ iki.fi
 biz.at
 info.at
 
+//JS.ORG : http://dns.js.org
```
sleevi: Please fix the whitespace here

sleevi:
Actually, this one I can just fix manually.

PR #277, opened 2016-08-06
Add 2LD info for .name

pzb:
The name TLD has a number of second level domains that are registry-operated.  Add them to the PSL.


sleevi:
This is another situation where I'm curious if we shouldn't just wildcard.

pzb:
the name TLD is rather special.  Quoting wikipedia's article on .name:

> On the .name TLD, domains may be registered on the second level (john.name) and the third level (john.doe.name). 
>
> When a domain is registered on the third level (john.doe.name), the second level (doe.name in this case) is shared, and may not be registered by any individual. Other second level domains like johndoe.name remain unaffected.

https://www.verisign.com/en_US/domain-names/name-domains/name-tld/index.xhtml confirms that some .name domains are registered at the second level and some at the third level.

Given this, `*.name` would not be an appropriate entry for the PSL.

sleevi:
On Saturday, August 6, 2016, Peter Bowen <notifications@github.com> wrote:

> Given this, *.name would not be an appropriate entry for the PSL.
>

No, we can use exception records to carve out the exceptions, which will at
least be smaller than the full set, because we'll be able to omit all the
3LD names.

pzb:
There are over 130000 names currently registered at the second level in .name. (e.g. example.name.).  There are only slightly over 19000 second level registry-controlled names in .name.  19k is much smaller than 130k.

AGWA:
Note that second-level registrations for .name are completely open; it's as easy to register them as it is to register a .com.  Second-level .name registrants, who are ordinary people, shouldn't have the burden of requesting a PSL exception and then waiting for browsers and other PSL consumers to update before their domain works properly.

Third-level registries, on the other hand, are operated professionally and presumably aren't created as often.  They--not second-level registrations--should be treated as the exceptional case and be listed in the PSL.

weppos:
> Second-level .name registrants, who are ordinary people, shouldn't have the burden of requesting a PSL exception and then waiting for browsers and other PSL consumers to update before their domain works properly.

@AGWA second level registrants don't have to ask for being listed in the PSL. For example, I own a second-level .name and I have the entire management of it, including the third levels.

In other words, if you can find a second level name which is open for registration, and you register it, then you own the "namespace". You don't need to request to be included in the PSL as only you can create subdomains.

pzb:
Any update on this?  I can regenerate the delta to fix the build, but I would like clarification that this will land.

weppos:
@pzb this is a quite challenging request. Merging it will add ~19k items, which effectively is more than the existing number of rules (11k).

@sleevi proposed the use of the wildcard, that was my very first idea as well. It's true though that we have 2 cases here:

1. **FOO**.name (rule `name`)
1. **FOO**.suffix.name (rule `*.name`)

and the wildcard would stop the first from being valid. To my memory, we never had cases where `rule` and `*.rule` would make sense in parallel (in fact I believe we have a test that checks and marks this case as forbidden). It also poses some interesting challenge to have it in parallel (because at that time _if_ the rule `weppos.name` is not present BUT weppos.name is a suffix, what will happen is that the eTLD of `foo.weppos.name` would be `name` instead of `weppos.name`.

In other words, adding the possibility to have both `rule` and `*.rule` may partially workaround this problem in terms of size, but may still provide an inaccurate implementation (although probably more accurate than what we have today).

Please note I am mostly speaking out of loud while travelling, hence I haven't properly analized all the possible implications of this idea. I also don't recall if it was ever considered and/or discarded (I know we have a test in libpsl that checks the presence of `rule` and `*.rule`, but I don't think it was part of an explicit decision).

weppos:
Please disregard the previous message. I actually just realized that the idea of having both `*.rule` and `rule` doesn't make sense at all, and it would not solve the problem.

So we are probably back to the original question: is the impact of merging 19k suffixes acceptable?

gerv:
I would be inclined against. More than doubling the size of a data file shipped with every browser on the planet, just to support one TLD, seems wrong. I've looked through the list - some of them seem very weird, and not the sort of thing that the .name registry is actually selling multiple different names under ("t-1424206458-1408216799559-3-sn.name", "sun-3-h1gra.name"?).

I think that the minimum we'd need to consider this would be an active application from the .name registry, with their collaboration on identifying which domains actually do need adding - e.g. those with more than 5 or 10 sub-registrations to mutually untrusting parties.

pzb:
If there is interest, I can pull number of names registered under each suffix.

gerv:
I certainly would find that interesting.

RegistryPRO:
The NIC.br of Brazil, has a domain similar to .name!

The domain is the .nom.br, where the registration is done as follows:

Name1.name2.nom.br

In "Name2", the NIC.br starts to operate it and to share the domain with other users.

In "Name1" only the subdomain holder can register and operate it.

In the list of public suffixes, NIC.br registered the .nom.br as a wildcard domain.
Https://publicsuffix.org/list/public_suffix_list.dat

I think the best way to add 3rd level records under .name in Public Suffix is the wildcard domain.

Real examples:

Felipe.pusanovsky.nom.br
Serrano.neves.nom.br
Arvores.brasil.nom.br
Meaning.origem.nom.br

See: http://whois.registro.br

sleevi:
As explained on that issue, that would not be appropriate. It is unlikely
we would or could accept nom.br due to the stability risk it poses to the
overall ecosystem.

RegistryPRO:
@sleevi I was able to check in the .name gTLD, that the most reserved names are country names, so that anyone can register them on the third level.

Example: weppos.italy.name or weppos.italia.name

Perhaps it would be necessary to include the list initially with the names of countries reserved by .name for shared use?

GANDI.net and ResellerClub.com are two of the few companies that make 3rd level registrations under .name

RegistryPRO:
@sleevi So far, the list of names that I'm sure are operated directly by NIC.name (Verisign) are country names.

pzb:
https://gist.github.com/pzb/098d96762c53c0c2d01155b5d8c5be97 has the data for .name as of 2017-06-17.  As you can see, there are 21881 known public suffixes for .name (plus `name` itself).  Of these, 3774 have more than one known subdomain.

I think .nom.br is different from .name because everything registered under .nom.br has two more labels.  In the case of .name the following are all valid current registered domains:

* peteratkins.name.
* ryanfamilyholdings.name.
* cadillac.japan.name.
* claire.jasper.name.

Because of this inconsistency, wildcarding is not possible.  Conversely, it seems `*.nom.br` is a correct rule, as there are no registered domains in the form <label>.nom.br.

RegistryPRO:
@p2b The correct would be to initially catalog suffixes with country names under the 2nd level in .name and include them in Public Suffix because they have a higher demand?
Thank you

pzb:
@RegistryPRO the data does not support the theory that county names are more popular.  Top suffixes are:

* japan.name.
* tokyo.name.
* mail.name.
* love.name.
* family.name.
* domain.name.
* tanaka.name.
* suzuki.name.
* yokohama.name.
* smith.name.

In contrast, italy.name. does not appear on the list at all and italia.name. only has one registration.

RegistryPRO:
@pzb include the domain (suffix) br.name!

PR #282, opened 2016-08-12
Add ro.im and shop.ro

globehosting:
Add ro.im and shop.ro private domains. We allow registration of second level domains using ro.im and shop.ro.


globehosting:
We have added DNS Authentication for both zones.

PR #284, opened 2016-08-14
Add wmflabs.org

yuvipanda:
Part of [Wikimedia Labs](https://wikitech.wikimedia.org/wiki/Portal:Wikimedia_Labs), allows users to
run arbitrary code on both [*.wmflabs.org](https://wikitech.wikimedia.org/wiki/Help:Proxy)
as well as [*.*.wmflabs.org](https://wikitech.wikimedia.org/wiki/Help:Addresses).

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -11708,6 +11708,10 @@ inc.hk
 // Submitted by Simon Kissel <hostmaster@viprinet.com>
 router.management
 
+// Wikimedia Labs: https://wikitech.wikimedia.org
+// Submitted by Yuvi Panda <yuvipanda@wikimedia.org>
+wmflabs.org
+
 // Yola : https://www.yola.com/
 // Submitted by Stefano Rivera <stefano@yola.com>
 yolasite.com


sleevi:
Did you mean to also add `*.wmflabs.org` org then?

As it currently stands, this rule indicates `foo.wmflabs.org` and `
bar.wmflabs.org` are distinct, but `bar.foo.wmflabs.org` and `
baz.foo.wmflabs.org` are the same.

Also, please review the Submission Guidelines (on the Wiki page) for
instructions on how requests should be formatted and authenticated.

yuvipanda:
Hello! Thanks for the review :)

Users who control bar.foo.wmflabs.org also entirely control baz.foo.wmflabs.org, so I think that can stay. Same is not true for foo.wmflabs.org and bar.wmflabs.org, so the current rules are correct? I cam amend description to make this clearer.

I'll add the authentication for the domains shortly!

sleevi:
OK, I wanted to confirm because your original message suggested second level domains (it looks like you may have said `*.*.wmflabs.org` - but `*` is a special character in GitHub, so I wanted to confirm.

Krenair:
Per discussion with Yuvi I've created the authentication record. You should be able to `dig TXT _psl.wmflabs.org. @labs-ns0.wikimedia.org` now. ns1 should get it in about an hour.

PR #285, opened 2016-08-16
Add transurl.nl

mrbreaker:
TransIP provides different internet services, among them domain name registration and webhosting. When a customer registers a domain name and webhosting packet, we also provide an additional domain name containing the transurl{.nl, .be, .eu} suffix _(so if a customer uses the domain example.nl, we would provide example.nl.transurl.nl)_. This extra domain name will also point to the customer's website.

Many different websites are hosted on subdomains under the transurl{.nl, .be, .eu} suffix and customers have almost full control over what they will be hosting. The main reason that we want our suffixes to be added is to prevent customers to make super cookies.

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -11693,6 +11693,10 @@ sopot.pl
 bloxcms.com
 townnews-staging.com
 
+// TransIP : htts://www.transip.nl
+// Submitted by Rory Breuk <rbreuk@transip.nl>
+transurl.nl
+
 // TuxFamily : http://tuxfamily.org
 // Submitted by TuxFamily administrators <adm@staff.tuxfamily.org>
 tuxfamily.org
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -11698,9 +11698,6 @@ townnews-staging.com
 *.transurl.be
 *.transurl.eu
 *.transurl.nl
-co.uk.transurl.be
-co.uk.transurl.eu
-co.uk.transurl.nl
 
 // TuxFamily : http://tuxfamily.org
 // Submitted by TuxFamily administrators <adm@staff.tuxfamily.org>
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -11695,9 +11695,12 @@ townnews-staging.com
 
 // TransIP : htts://www.transip.nl
 // Submitted by Rory Breuk <rbreuk@transip.nl>
-transurl.be
-transurl.eu
-transurl.nl
+*.transurl.be
+*.transurl.eu
+*.transurl.nl
+co.uk.transurl.be
+co.uk.transurl.eu
+co.uk.transurl.nl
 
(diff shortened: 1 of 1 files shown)


mrbreaker:
Added the DNS TXT records for _psl.transurl.nl, _psl.transurl.be and _psl.transurl.eu

sleevi:
Could you expand more on the description, as per the [Guidelines](https://github.com/publicsuffix/list/wiki/Guidelines)? We just want to make sure the PSL is the best solution for your use case :)

mrbreaker:
@sleevi : thank you for your response, I updated the description.

sleevi:
Thanks for updating the description.

Unfortunately, your naming scheme - of using customer-domain.transurl.nl - will still not result in the correct behaviour, and may thus still present a risk. As it stands, the PSL, nor clients (such as browsers), can safely support your current naming strategy.

Consider the following:
`domain2.example.transurl.nl`
`domain1.example.transurl.nl`

In this case, `domain1.example` and `domain2.example` are two distinct customer domains you're registering. Despite adding `transurl.nl` to the PSL, the effect is such that it only applies to one level - such that `domain1.example.transurl.nl` and `domain2.example.transurl.nl` have the same effective TLD as `example.transurl.nl` and the same suffix of `transurl.nl`. As a result, they'd each be able to set cookies for `example.transurl.nl` - causing supercookies.

Even if you were to use an entry such as
`*.transurl.nl` (so that every TLD that is in the 3LD of your name is itself a public suffix), that wouldn't work for something like `ryan.sleevi.name.transurl.nl` - because `jim.sleevi.name.transurl.nl` would still be able to set cookies for `sleevi.name.transurl.nl`

What you want (and what the PSL doesn't support) is some sort of indicator that every subdomain is part of a public suffix.

If you did `example-com.transurl.nl`, then having `transurl.nl` would work, but then you have to deal with issues like IDNA (for example, is `xn--foo.com` translated to `xn--foo-com` ? That doesn't work), and you'd also run into label length limits vs domain length limits.

In short, there's no solution that offers you the security you want, and so I'm not sure that adding to the PSL provides the value you hope for.

mrbreaker:
Thank you for your extensive response. You are totally right about this issue with the transurl.nl suffix. 

We only use this naming scheme for second level domains registered with us (with the exception for domains registered under co.uk). This means that with our naming scheme, a domain name like `ryan.sleevi.name.transurl.nl` would not be possible. I believe that using the `*.transurl.nl` entries, we would sufficiently cover all the domain names that we provide, given that I also added specific entry for the subdomains of `co.uk.transurl.nl`.

sleevi:
It looks like you removed the co.uk sub-specifiers and edited the comment; was that intentional?

Otherwise, this looks good - I just want to confirm where we stand on that special case.

mrbreaker:
Yes, this was indeed intentional. We decided to find a more generic solution for the co.uk exception so it would fit the same naming scheme.

Issue #20, opened 2015-08-01
Usage example


szepeviktor:
May I put this into the [Wiki](https://github.com/publicsuffix/list/wiki)?
This script outputs only the SLD part from a list of domain names.
So if you have `subdomain.example.co.uk` on your list, the output will be `example.co.uk`.
```bash
# Prepare the list for finding second-level domains
Publicsuffix_regexp() {
    local LIST_URL="https://publicsuffix.org/list/public_suffix_list.dat"

    # Download list,
    #   remove empty lines and comments,
    #   escape dots, asterisks and add SLD regexp
    wget -qO- "$LIST_URL" \
        | grep -v "^\s*$\|^\s*//" \
        | sed -e 's/\./\\./g' -e 's/\*/.*/' -e 's/^\(.*\)$/[^.]\\+\\.\1$/'
}

Publicsuffix_regexp | grep -o -f - "$YOUR_DOMAINS"
```
I use it for registration expiry purposes.

weppos:
@szepeviktor thanks for your contributions. I'm closing this, as we don't currently add usage examples or libraries to the wiki.

Also double check your code properly reflects the [formal algorithm](https://publicsuffix.org/list/), including the exceptions. Or you may get inaccurate results.

PR #17, opened 2015-07-21
Add sandcats.io dynamic DNS service as public suffix

paulproteus:
I'm on the sandstorm.io team, where we maintain a public dynamic DNS service at sandcats.io.

@kentonv is the listed domain owner. Let me know what we should do to fully confirm the domain name to your satisfaction.

PR #293, opened 2016-08-18
Add domreg.merit.edu registry

mdkuch:
ann-arbor.mi.us
cog.mi.us
dst.mi.us
eaton.mi.us
gen.mi.us
mus.mi.us
tec.mi.us
washtenaw.mi.us

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -6584,6 +6584,17 @@ lib.wy.us
 pvt.k12.ma.us
 chtr.k12.ma.us
 paroch.k12.ma.us
+// Merit Network, Inc. maintains the registry for =~ /(k12|cc|lib).mi.us/ and the following
+//    see also: http://domreg.merit.edu
+//    see also: whois -h whois.domreg.merit.edu help
+ann-arbor.mi.us
+cog.mi.us
+dst.mi.us
+eaton.mi.us
+gen.mi.us
+mus.mi.us
+tec.mi.us
+washtenaw.mi.us
 
 // uy : http://www.nic.org.uy/
 uy


sleevi:
Are you a representative of the domain registry?

we've seen issues in the past with adding certain institutions to the PSL, so we'd likely need a representative of the registry itself to confirm that it wants to be on the PSL, given the tradeoffs.

mdkuch:
Thank you for the due diligence. If required, we can publish DNS TXT records per the publicsuffix guidelines, but I prefer not to. I can be reached by the published email registry email address: domreg@merit.edu

I am not aware of any issues with the current inclusion of k12.mi.us and lib.mi.us in the PSL. It seems appropriate to include our additional suffixes. Can you point me toward any issues other institutions have faced once included in the PSL? Is this change something that should be communicated to registrants prior to the commit?

 \- - Mitchell

sleevi:
A few past examples:
*  https://github.com/publicsuffix/list/issues/243
*  https://bugzil.la/941670
*  https://bugzil.la/940478
*  https://bugzil.la/547985
*  https://bugzil.la/984824

mdkuch:
Each of the additional suffixes listed in this pull request serves independent institutions that have registered domain names under the same program that supports k12.mi.us, lib.mi.us, and cc.mi.us. Services operated in the domains below the suffixes described within should be treated independently. Please proceed with their inclusion in the PSL.

sleevi:
Sent an email to the WHOIS contact to out-of-band authenticate the request.

mdkuch:
The out-of-band authenticate request has been acknowledged.

public_suffix_list.dat
```diff
@@ -6584,6 +6584,17 @@ lib.wy.us
 pvt.k12.ma.us
 chtr.k12.ma.us
 paroch.k12.ma.us
+// Merit Network, Inc. maintains the registry for =~ /(k12|cc|lib).mi.us/ and the following
+//    see also: http://domreg.merit.edu
+//    see also: whois -h whois.domreg.merit.edu help
```
sleevi: Unfortunately, I can't use the web-editor to suggest changes and feedback, but we've been trying to tighten up the internal format/documentation for these additions.

Since all of this discussion is available via history, would you be able to put something simpler:

// Regional and industry-related 3LDs offered by the mi.us registrar - see http://domreg.merit.edu
// Submitted by mi.us registry <email here>

@gerv @weppos - Should we be putting these under ICANN or PRIVATE? I'm not sure how we want to handle that distinction - is it the same as a `.com.ccTLD` case operated by the `.ccTLD` registry, or do we see it more like `appspot.com` - the party who is delegated `appspot.com` (e.g. Google) is then sub-delegating? I can see the argument either way, and I'll note it somewhat touches on what Peter Bowen raised recently.

sleevi:
Thanks. I've got one question for the other maintainers about how we want to handle this, and as soon as we have rough consensus, we can go forward with merging this.

public_suffix_list.dat
mdkuch: Of course. We can make any of the suggested modifications.

I wrote the initial comments with the intent to articulate that Merit is the registry and solo registrar of many, but not all, of the mi.us. 3LDs. Other mi.us 3LDs are registered directly with [nic.us](http://nic.us) under their [Locality-Based Name Registrations](http://www.neustar.us/locality-based-name-registrations/) program. It is unclear if other mi.us 3LD registrations are used by single institutions or if they offer a registry for their respective memberships.

Historically, these domains have operated in a fashion more inline with the PSL's ICANN definition then the PRIVATE definition.

sleevi:
@gerv @weppos Can you see my questions above and comment? I think we have everything we need here, other than deciding which section to place these under.

sleevi:
@gerv @weppos Can you please respond to https://github.com/publicsuffix/list/pull/293#issuecomment-253092080 ?

In the absence of any further response, I will simply do what I think is appropriate, but I would prefer consensus.

gerv:
Sorry, I must have missed your earlier request.

We should probably place them based on their desires related to the use cases where people make the distinction. For example, if they want to be able to get a *.dst.mi.us wildcard cert, they need to go in the PRIVATE section. What other use cases are distinguishing between the two halves?

gerv:
@sleevi: last comment is directed at you.

sleevi:
@gerv I defer to you to accurately maintain the full state of "Why PRIVATE" :)

My own understanding was that we'd reserved the ICANN section for those gTLDs/ccTLDs delegated, and any subdivisions done by the registrar (versus by a registry or a domain registrant). If your view is that they're equivalent, modulo wildcard certificates, I can buy that.

I would have to check through the fullness of the use cases to accurately know what other impacts (beyond wildcards) from the PRIVATE vs ICANN section. I just note we haven't seemingly added any to that section over the past several years, unless it was divided by the registrar, but perhaps I read the history wrong.

gerv:
You are right - in absence of a good understanding of the consequences of doing otherwise, we should maintain our policy of putting only registrar-controlled divisions into the ICANN section. The question is more about: which of the two situations is this one, actually? I guess Merit didn't just register these domains on the open market so they can now sell subdomains of them, but they have had them delegated to them to manage by mi.us or nic.us. Is that correct?

daftano:
@mdkuch: you may want to reply to last question.

mdkuch:
@gerv: Correct.

Merit is the nic.us assigned Delegated Manager for these domains.

dnsguru:
I am late to the discussion but I am of the opinion that these SLD under the .US states really follow more of the ICANN section, as they follow a delegation path chain that is direct authority, even though it is under a former RFC for the .US ccTLD

Although we focus on the pure intent of PSL use, the derivative community and commercial uses that flow from our efforts merit some representation here.  Many of the derivative users/integrators use the private section header as the 'ignore from here' location to keep their code simpler.

There is a potential if we change the .US sub-states listing location to move them into the private section, we could break academic or civic email systems.

The 'Private' section is more designed for 'everything else'.

sleevi:
@dnsguru One noteworthy exception is that the 'PRIVATE' section may imply certain behaviour in clients. For example, Chrome leverages the PRIVATE section to know whether a domain is navigable, or whether it's treated as a [cc/g]TLD (for example, `appspot.com` should navigate, but `com` should not). Similarly, wildcards or not are gated based on ICANN vs PRIVATE (which is also an imperfect solution).

That said, I went ahead and merged the changes as is, as it followed local file comment and the requestor was happy with it being in the ICANN section. I asked both for the reasons above, and to determine whether we had a pattern for 'similar, but not quite' requests, such as where one or more registrars within a ccTLD establish their own SLDs below the ccTLD, as we frequently see - if they aren't assigned by the ccTLD operator, my view has been that they're PRIVATE, and this reflects other DDNS services or registrar-operated SLDs.

dnsguru:
I should have clarified - This belongs within the ICANN section because the [RFC 1480](http://www.ietf.org/rfc/rfc1480.txt?number=1480) use is the only use within the State registrations, and these are managed in a direct manner by the registry and requests to the registry for new domains when they are added.

Neustar, the current registry operator for .US, took this over in 2001, assuming the existing RFC 1480 structure into their registration agreement.

Regional operations still fall under the ".US Locality Domain Name Registration
Terms and Conditions" of the .US registry operator.   [Their FAQs](http://www.about.us/faqs) list in #17-#20 list the conditions and information about the process of managing, updating, listing in whois, and other requirements of these subdomains.

> 19. Which locality domains are included in the .US WHOIS?
> Second-, third-, fourth-, fifth- and sometimes sixth-level domain names are included.
> 20. I have a third/fourth/fifth/sixth-level locality .US domain name, but I can’t find it in the .US WHOIS. How can I correct this?
> First, check in WHOIS for the delegated manager of the parent of your domain name. You can also send us (support.us@Neustar.us) an email, along with your domain name, and we will advise on whom to contact.

The list of these is fairly narrow.   What I suggest that we as volunteer maintainers do on these going forward to keep it simple is to request that the modifications of legacy RFC 1480 .US regional subdomains must come from the whois listed contact of the specific domain(s) that the request is coming from, or that of the parent TLD.    Example: foo.bar.state.us  could request a PSL update or addition if it sourced from the whois contact for foo.bar.state.us, but also the listed whois contact for bar.state.us, or state.us or .us could.

PR #295, opened 2016-08-26
DynamicDNS and Portmapping Service Feste-IP.net

Feste-IP-net:
Subdomains of these zones are used by our DynamicDNS and IPv4/6 Portmapping Service feste-ip.net

TXT Records for all zones are set.

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -10835,9 +10835,9 @@ co.no
 
 // COSIMO GmbH http://www.cosimo.de
 // Submitted by Rene Marticke <rmarticke@cosimo.de>
+dyn.cosidns.de
 dynamisches-dns.de
 dnsupdater.de
-dyn.cosidns.de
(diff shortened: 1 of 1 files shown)


sleevi:
I was only able to validate l-o-g-i-n.de and feste-ip.net.

Could you also please update the PR to include the proper company name and URL, rather than description of services and URL, and then sort accordingly based on company name?

Thanks.

sleevi:
Apologies to keep asking for changes, but this latest update still doesn't meet the [Guidelines](https://github.com/publicsuffix/list/wiki/Guidelines) - the domains should be sorted, by TLD, then 2LD, then 3LD (if any), etc.

Feste-IP-net:
Hi Sleevi,

I've updated the informations.

/Greetings Rene

Feste-IP-net:
Hello,

Is there anything else needed from our side?

/Regards René

PR #297, opened 2016-08-29
Add taifun-dns.de

BjHe:
Add domain taifun-dns.de to public_suffix_list.dat.
This domain provides dynamic dns for our customers.


BjHe:
DNS TXT record _psl.taifun-dns.de is now set for authentication.

PR #301, opened 2016-09-21
Adding new Zimbabwe TLD's

mrname:
Adding valid TLD's for Zimbabwe as sold here:

https://www.name.co.zw/


sleevi:
This is not necessary. The existing `*.zw` entry already communicates this.

PR #302, opened 2016-09-22
Add shiftedit.io

adamjimenez:
Subdomains under shiftedit.io are issued to shiftedit users.
e.g: customer.shiftedit.io

DNS verification added

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -11663,7 +11663,7 @@ pp.ua
 
 // ShiftEdit : https://shiftedit.net/
 // Submitted by Adam Jimenez <adam@shiftcreate.com>
-shiftedit.net
+shiftedit.io
 
 // Shopblocks : http://www.shopblocks.com/
 // Submitted by Alex Bowers <alex@shopblocks.com>
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -11661,6 +11661,10 @@ biz.ua
 co.ua
 pp.ua
 
+// ShiftEdit : https://shiftedit.net/
+// Submitted by Adam Jimenez <adam@shiftcreate.com>
+shiftedit.net
+
 // Shopblocks : http://www.shopblocks.com/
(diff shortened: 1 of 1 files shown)


sleevi:
To confirm: Are you aware that by placing your main domain on the PSL, you may create issues for users to navigate to your domain, and you will be limited to what cookies you can set (e.g. for analytics purposes)?

In general, cloud/hosting providers are strongly encouraged to use separate domains for 'customer-controlled' content than from their navigation domains for this reason, due to the high risk of causing issue.

adamjimenez:
Thanks - acknowledged, I've changed the domain to shiftedit.io

groundcat:
Hello @adamjimenez , it seems that the project or organization "**shiftedit**" associated with your domain, which was merged into the Public Suffix List (PSL), may no longer be active based on the observable usage of subdomains. If the service has been decommissioned and the domain is no longer linked to an active project, as outlined in your original request, please confirm whether it can be safely removed from the PSL. This will help prevent leaving unnecessary entries for PSL volunteers and future registrants, as requesters often "set and forget" and neglect to request the removal of their domains from the PSL when their project reaches its end of life (EOL).

The PSL is a globally used resource, and it's important to understand that tens of millions, if not hundreds of millions, of devices and users incorporate the list. We aim to keep it as minimal as possible for those who truly need PSL inclusion. If you no longer need your entry to remain on the PSL, please create a new pull request to remove it or reply here to confirm.

If your project or organization is still active and requires continued inclusion in the PSL, please reply at your earliest convenience. If we do not hear back from you for an extended period, **your entry may be subject to removal**. Additionally, ensure that the [required `_psl` TXT record](https://github.com/publicsuffix/list/wiki/Guidelines#validation-of-submissions--always) is maintained, as specified in the [PSL Guidelines](https://github.com/publicsuffix/list/wiki/Guidelines), and that your domain is kept renewed with an expiry date of more than 2 years at all times to avoid future removal.

Thank you.

PR #305, opened 2016-09-24
Add nom.nc - official second-level domain

ebzao:
There is another reserved extension directly attached on nc TLD: nom.nc
See http://www.cctld.nc/ind?man=2
.nc : generic extension
.asso.nc : reserved for associations
.nom.nc : reserved for the private individuals who shall wish to to deposit their surname.

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -4398,6 +4398,7 @@ name
 // nc : http://www.cctld.nc/
 nc
 asso.nc
+nom.nc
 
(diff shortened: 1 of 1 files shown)


sleevi:
Are you the official operator of these domains? If so, can you please update per the Guidelines - https://github.com/publicsuffix/list/wiki/Guidelines ?

If not, we'll need to close this out.

weppos:
@gerv I've approved the request as I was able to validate it with official documents (the site). We did that in the past, when the registry was not extremely responsive.

PR #308, opened 2016-09-29
Add opencraft.hosting

smarnach:
We provide a hosting service, and subdomains of the domain are issued to mutually untrusting parties.  Customers can choose their subdomain during [registration](https://console.opencraft.com/registration/).


smarnach:
I've created a TXT record for `_psl.opencraft.hosting` pointing to this PR.  DNS changes should propagate in less than half an hour.

PR #314, opened 2016-10-07
Add Brazilian states .leg.br

GabrielRF:


GabrielRF:
`TXT` records created.

sleevi:
Could you indicate why you're requesting to be included?

For example, this would help us determine if `*.leg.br` may be a more appropriate entry.

GabrielRF:
Interlegis.leg.br is a part of the Brazilian Government. We are responsible for domains `.leg.br` for each Brazilian State (the list that I submitted). 

`.leg.br` means these domains are for __each legislature of every Brazilian city__. 

Multiple domains are found under each `state.leg.br`.
Our goal is to assist each and every Brazilian city having it's own domain.

sleevi:
Within the set of domains of `.leg.br` - are there domains which are not part of `state.leg.br`? That is, the expression `*.leg.br` helps PSL implementations save considerable space over listing each entry, but it wouldn't work if there are non-state subdomains (e.g. if there is `foo.leg.br` and `bar.leg.br`, which aren't states). It looks like it might not work because `interlegis.leg.br` doesn't sound like a state name, but if that's the only case, then perhaps we can figure out a solution to allow `*.leg.br` and minimizing disruption to `interlegis` (such as an exception record)

That is, I'm trying to see if
```
*.leg.br
!interlegis.leg.br
```
Works :)

GabrielRF:
Yes, there are multiple domains that are not a part of `state.leg.br`.

For example:

```
camara.leg.br
interlegis.leg.br
senado.leg.br
infra.leg.br
tcu.leg.br
```

Interlegis is responsible only for `state.leg.br`. Registro.br is responsible for `.leg.br`.

sleevi:
OK, thanks for confirming. It's unfortunate that this is the case, but this ostensibly meets the bar.

GabrielRF:
@sleevi There is a typo on `submited`. I've skipped a `t`.
Should I send a PR again?

sleevi:
@GabrielRF Ah, I missed that. You can just send a new PR and ping me and I'll merge :)

Issue #271, opened 2016-07-25
Add domains from afraid.org (ddns) to PSL


josmers:
As discussed for other ddns providers (https://github.com/publicsuffix/list/pull/64) please add the domains listed by http://freedns.afraid.org/domain/registry/.
As the raw list contains over 85000 entries, this is a much greater effort than adding the fewer _noip_ domains. Hopefully this serves the purpose of the PSL and taking the automation discussion further.

Cheers,
Jan

gerv:
@josmers : I assume you are an official representative of afraid.org?

josmers:
@gerv ,
I'm sorry I'm not an official representative. I'm just an ordinary user of the services offered. If I can help somehow please let me know. I'm happy to support in this case if possible.

wrb:
I sent a mail to the owner of afraid.org, hopefully he will stop by and do what's necessary.

Afraid.org hosts dns many public domains that are not owned by him, but hopefully he can add at least the ones he owns.

wrb:
I talked to the owner and admin of afraid.org. He mentioned but there's a automated list of all domain hosted on afraid.org that are public [1] but he is not interested in helping any further.

So unless PSL owners are willing to make an exception and add the public domain list without the site owner's verification (which would also require scraping the page, because there's no API or export, AFAIK), the 84,884 public dynamic dns domains hosted by afraid.org will never get added to PSL and thus never be available for Let's Encrypt certificates.

[1]: http://freedns.afraid.org/domain/registry/

nistvan86:
Any chance of moving this forward? Maybe if someone writes a scraper to extract the domains?
I would really love to use my homenet.org address with letsencrypt.

weppos:
Unfortunately, I have to close this issue as for the others that were opened.
As per our [guidelines](https://github.com/publicsuffix/list/wiki/Guidelines), only the owner can request the inclusion.

Moreover, as the intent is to get into Let's Encrypt, this is not the right channel. You should use
https://letsencrypt.org/docs/rate-limits/

josmers:
Sadly to hear you closed this issue. For me this is the remarkable issue along several others that all has been closed, too. Actually the other linked issues just requested to add a few domains. This one is more general.

I understand you guidelines. The point is totally valid. Second, Let's Encrypt is off topic of cause. Their "Rate-Limit" has nothing to do with this issue. Moreover it is not even a workaround as the described procedures are for domain owners only.

[...]

PR #323, opened 2016-10-26
added dynamic DNS provider afraid.org

kampa133:
I wrote this e-Mail to :dnsadmin@afraid.org

Hi,
I'd like to use letsencrypt for my homeservers but i get the error message:
There were too many requests of a given type :: Error creating new cert :: Too many certificates already issued for: mooo.com

I made a pull request for the public suffix list with the domains i know you are using. But there will be quite more I guess.
https://publicsuffix.org/

link to the pull request: 

kind regard
Manuel

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -10721,6 +10721,16 @@ poivron.org
 potager.org
 sweetpepper.org
 
+// AFRAID : https://freedns.afraid.org/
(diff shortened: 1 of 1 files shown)


sleevi:
Only the owner of the domain is allowed to ask the entry to be listed (see #271). If you are interested in being able to obtain certificates with Let's Encrypt, they now offer a form to bypass request limits.

PR #326, opened 2016-11-01
Add us.to suffix of FreeDNS service.

lexologe:
I added the us.to suffix of the FreeDNS service. See http://freedns.afraid.org/


weppos:
Please see #271

PR #129, opened 2016-02-03
add freeboxos domains to public_suffic_list.

r0ro:
Freebox offer its customers the ability to reserve a custom
subdomain {yourname}.freeboxos.fr to access their router.

Issue #88, opened 2015-12-22
Add FreeDNS domains


bt90:
The letsencrypt project is using the public suffix list while checking for accidental and intentional abuse.(see https://community.letsencrypt.org/t/rate-limits-for-lets-encrypt/6769)

The result is that this prevents all users of the FreeDNS project(http://freedns.afraid.org/) from getting a free certificate due to said limits.

The problem is that we're talking about a list of ~90k domains which is being constantly updated with some being marked as 'private'(see http://freedns.afraid.org/domain/registry/). The addition of the public ones is still valid as they are in fact publicly used suffixes which are open for registration(e.g. mooo.com with ~400k subdomains). 

Not adding them would prevent ~8 million sites(check 'records' at http://freedns.afraid.org/domain/registry/) from using letsencrypt.

The sheer number of domains and their fluctuation would require automation. I'm already in contact with one of their admins(dnsadmin@afraid.org) and they would provide an export link for this purpose.

TeZet2:
I strongly support this request! Accidentially (!? ;-) ) I was able to get a LE certificate for my mailserver at home and switched over from self signed to "real" SSL. Works like a charme. Tried to do the same with my owncloud server only one day later and I ran into the LE rate limits :-(. I strongly hope that either the rate limits will fall or we will see FeeDNS domains added to the public suffix list. I must admit that I am already addicted to LE ... .

weppos:
Duplicate of #79

PR #327, opened 2016-11-03
Add hd.free.fr domains

r0ro:
Add a new suffix end user can choose to get a custom domain name

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -11307,6 +11307,7 @@ fbx-os.fr
 fbxos.fr
 freebox-os.fr
 freeboxos.fr
+hd.free.fr
 
 // Fusion Intranet : https://www.fusion-intranet.com
 // Submitted by Matthias Burtscher <matthias.burtscher@fusonic.net>


weppos:
@r0ro do you represent freebox.fr and/or are you the owner of the domain you are submitting?

r0ro:
Yes, I am. I was the one submitting the first inclusion in the list. (https://github.com/publicsuffix/list/pull/129)

Freebox and Free (and Online.net) are subsidiaries of 'Iliad' (see: http://www.iliad.fr/en/societes.html), and xxx.hd.free.fr domains are offered to Free broadband customers to get a free domain name and custom reverse dns (see: http://www.free.fr/assistance/54.html)

Due to the way this domain is managed, dns entry is not easy to setup, can you valide this domain by emailing me at rfliedel@corp.free.fr ?

weppos:
@r0ro I'd actually have to ask you to provide the DNS entry. The reason is because we are working towards automating it, hence we'd rather not proceed anymore with email validation (as it's very error prone).

Also please follow right-left sorting (which means the new entry should go before freeboxos.fr, that should be moved before freebox-os.fr)

r0ro:
Just to be sure the correct order should be:
```
freebox-os.com
freeboxos.com
fbx-os.fr
fbxos.fr
hd.free.fr
freeboxos.fr
freebox-os.fr
```

because the pslint.py says it should be

```
freebox-os.com
freeboxos.com
fbx-os.fr
fbxos.fr
freebox-os.fr
freeboxos.fr
hd.free.fr
```

I'll push on the admin to get the proper dns record ...

bcecchinato:
@r0ro Since you are part from Illiad's company, would it be possible to add dedibox.fr as well ?

Regards,

r0ro:
@bcecchinato I can ask but why ? I think it's only used for default server names

bcecchinato:
@r0ro My bad I forgot to tell why. Letsencrypt is using this is to remove limit rates when requesting ssl certificates :) (see #88 for instance).

r0ro:
I'm not sure it's something we want to encourage, and not so much people seems to be using it: https://crt.sh/?Identity=%25dedibox.fr&iCAID=16418&n=1000

sleevi:
Any update on the proper DNS record?

Chris12345678900:
Why this domain is always in pending validation?
Will it be added one day to the list?
Tanks

gerv:
@Chris12345678900: it needs the proper DNS record put in place, as requested in December. See https://github.com/publicsuffix/list/wiki/Guidelines .

PR #329, opened 2016-11-09
Update .JP suffix as requested

weppos:
> The correction is as follows;
>
> please update the "jp" section of PUBLIC SUFFIX LIST.
> (remove "hitoyoshi.kumamoto.jp","kashima.kumamoto.jp")
>
> The attached file is correct.
> Please confirm it.


weppos:
Reviewed and approved via email.

PR #331, opened 2016-11-15
Added dynamic domains for Yombo

yombo:
These domains are used by users for dynamic DNS.  These domains need to be added since Yombo Gateway software will host websites. Requesting these be added to this so that cookies don't transfer between 3rd level (and lower) domains.

The domains in this pull request all use ns1.yombo.net and ns2.yombo.net as the name servers to validate that Yombo controls these domains.

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -11879,13 +11879,13 @@ yolasite.com
 
 // Yombo : https://yombo.net
 // Submitted by Mitch Schwenk <mitch@yombo.net>
-homelink.one
 ybo.faith
+yombo.me
+homelink.one
 ybo.party
 ybo.review
 ybo.science
 ybo.trade
-yombo.me
 
 // ZaNiC : http://www.za.net/
 // Submitted by registry <hostmaster@nic.za.net>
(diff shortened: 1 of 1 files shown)


sleevi:
Please review the [Guidelines](https://github.com/publicsuffix/list/wiki/Guidelines) for how to have your request processed.

yombo:
This has been completed. Thanks.

public_suffix_list.dat
```diff
@@ -11877,6 +11877,16 @@ wmflabs.org
...
+// Yombo : https://yombo.net
+// Submitted by Mitch Schwenk <mitch@yombo.net>
+homelink.one
+ybo.faith
+ybo.party
+ybo.review
+ybo.science
+ybo.trade
+yombo.me
+
```
sleevi: Please review https://github.com/publicsuffix/list/wiki/Guidelines#prepare-the-change for how to properly order these.

yombo:
Fixed.

sleevi:
Validated all domains.

As a side note, I'd like to inform you that to protect the consumption of the PSL, we are considering to limit the number of per-entity private domains. At the time being, I would be ok to include the domains in this request. However, we may not be able to accommodate the request if the number should continue to increase in the future.

PR #337, opened 2016-11-25
add scsbox.com

lim-daniel:
I would like to add this domain to public suffix list because eaech device uses SSL keys provided by Let's Encrypt. Let's Encrypt doesn't limit certificate requests for domain listed here.

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -665,6 +665,7 @@ web.co
 
 // com : https://en.wikipedia.org/wiki/.com
 com
+scsbox.com
 
 // coop : https://en.wikipedia.org/wiki/.coop
 coop


sleevi:
This request does not conform to the [Guidelines](https://github.com/publicsuffix/list/wiki/Guidelines)

Please update this PR when it does, or it will be closed out over time.

lim-daniel:
I would like to add this domain to public suffix list because eaech device uses SSL keys provided by Let's Encrypt. Let's Encrypt doesn't limit certificate requests for domain listed here.

I also set TXT record.

weppos:
@lim-daniel the patch is incorrect, as the formatting doesn't follow the guidelines.

However, for Let's Encrypt limits you should use this form

PR #339, opened 2016-11-28
Add certsbridge.com to Google, Inc. section

mbwalas:
certsbridge.com is Google-owned domain. Will be used for testing.

/cc @sleevi


sleevi:
As a matter of policy, we have not been adding testing domains, and are considering limits on the number of domains associated with an organization in order to keep growth manageable. At this time, it does not seem necessary to add this domain.

Issue #206, opened 2016-04-18
Please remove saratov.ru


Paradokc:
Hello. 
Please remove saratov.ru from publicsuffix

sleevi:
Can you explain why and how to authenticate this request?

Paradokc:
I am the owner of domain. 
https://www.nic.ru/whois/?query=saratov.ru&opensearch=1
How can I confirm this for you?

Paradokc:
And this domain name is not geographical from 2011

sleevi:
It looks like the registry cancelled all geographic registrations in 2009, based on the footnote at https://cctld.ru/ru/docs/archive/detail.php?ELEMENT_ID=3446

@weppos @gerv It seems like we should update the .ru registry based on https://cctld.ru/ru/domains/domens_ru/reserved/ [[Translated]](https://translate.googleusercontent.com/translate_c?act=url&depth=1&hl=en&ie=UTF8&prev=_t&rurl=translate.google.com&sl=auto&tl=en&u=https://cctld.ru/ru/domains/domens_ru/reserved/&usg=ALkJrhhBecD5877hg48pF0LkPbL1D-tPLw) . Are there any concerns?

gerv:
We've had at least one other of these come through... Is there an existing issue about this? Anyway, it would be great to find out the situation and make the PSL match.

sleevi:
There isn't, @gerv, which is why I was hijacking this. I included more information about "the situation", so I'm trying to see what, if anything, else should be provided before we take action.

weppos:
Most of the `.ru` entries were changed/added in 2008, one year before the regional hostnames were deleted.
https://github.com/publicsuffix/list/blame/master/public_suffix_list.dat

The removal of `mosreg.ru` was requested in 2014, see [Bugzilla#1090800](https://bugzilla.mozilla.org/show_bug.cgi?id=1090800). Here's the request body:

> We are Moscow Region Government. 
> Because CCTLD cancelled all geographical domain names on September 30, 2009 ( http://cctld.ru/ru/domains/domens_ru/domaintypes/ ). We need to delete the mosreg.ru domain from public suffix list

It looks like at that point we accepted the request, but we didn't fully investigate and realize that the cancellation would have affected most of the listed geographical domain names.

To mee it seems that we should sync the .ru list to match the content of https://cctld.ru/ru/docs/archive/detail.php?ELEMENT_ID=3446

Here's the (unsorted) list:

```
// industry

[...]

PR #344, opened 2016-12-04
Correct .ru secion

police87:
Greetings! Some links for .ru and .rf sections are outdated, correct links in this commit.

Also there are no more public geographical domains, all this domains belongs to organisations or private persons now, and cannot be used by anyone else, here is the official documentations about this:
https://cctld.ru/en/news/news_detail.php?ID=634&sphrase_id=128335
https://cctld.ru/en/domains/domens_ru/reserved/


weppos:
This is connected to https://github.com/publicsuffix/list/issues/206#issuecomment-211822371
We need to agree on a removal policy.

PR #346, opened 2016-12-05
Remove ro.com from Public Suffix List

tomasen:
We are the owner of ro.com. It's a domain name and we have build a website on it. It's not a public suffix. Please merge this pr, so that we can apply https cert for our domain.

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -11380,7 +11380,6 @@ gitlab.io
 
 // GlobeHosting, Inc.
 // Submitted by Zoltan Egresi <egresi@globehosting.com>
-ro.com
 ro.im
 shop.ro
 


sleevi:
Could you please see the [Guidelines](https://github.com/publicsuffix/list/wiki/Guidelines) for how to authenticate this request?

It looks like ro.com has changed hands, but we want to verify that this request is authorized.

tomasen:
thanks. we already added the dns record as instructed. please tell us if we missed something

dig +short txt _psl.ro.com
"https://github.com/publicsuffix/list/pull/346"

tomasen:
Haven't got any reply for more than a week. Hoping we can get this fixed before holiday. Because we can't apply for https certificate if we still on that list.

PR #348, opened 2016-12-07
Add user.party.eus to PRIVATE section

marcan-euskal:
This subdomain is used to automatically assign hostnames to event attendees. They should not be allowed to set cookies on the parent domain or interfere with Let's Encrypt rate limiting on the parent domain or other subdomains. Users can effectively trigger arbitrary domain registration under this subdomain.

I am the systems administrator in charge of DNS under party.eus and am authorized by the domain owner to request this change. As proof of domain control, I've added a TXT record under the registered domain:

```
$ dig TXT party.eus
[...]
party.eus.              600     IN      TXT     "Please add user.party.eus to the PSL (marcan@euskalencounter.org / GitHub marcan-euskal)"
```

Please let me know if this is sufficient documentation.

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -10725,6 +10725,10 @@ s3-website.us-east-2.amazonaws.com
 // Submitted by Thomas Orozco <thomas@aptible.com>
 on-aptible.com
 
+// Asociación Amigos de la Informática "Euskalamiga" : http://encounter.eus/
+// Submitted by Hector Martin <marcan@euskalencounter.org>
+user.party.eus
+
 // Association potager.org : https://potager.org/
 // Submitted by Lunar <jardiniers@potager.org>
 pimienta.org


marcan-euskal:
Just saw that there is a spec for TXT validation, so I added these records:

```
_psl.party.eus.         600     IN      TXT     "https://github.com/publicsuffix/list/pull/348"
_psl.user.party.eus.    600     IN      TXT     "https://github.com/publicsuffix/list/pull/348"
```

groundcat:
Hello @marcan-euskal I have noticed that the domain `party.eus` has been returning an `NXDOMAIN` error. Is this project still active? If the service has been decommissioned and the domain is no longer linked to an active project, please confirm whether it can be safely removed from the PSL. 

Additionally, the WHOIS expiration date is 2025-03-10, which is less than two years from now. To maintain your domain in the PSL, please ensure it is renewed for more than two years at all times. 

If your project or organization is still active and requires continued inclusion in the PSL, please reply at your earliest convenience. If we do not hear back from you for an extended period (over 30 days), your entry may be subject to removal. Thank you. #1119

PR #350, opened 2016-12-09
Add ticketswap.pizza to PRIVATE section [DNS verification added]

ruudk:
This subdomain is used to automatically assign hostnames to teammembers. They should not interfere with Let's Encrypt rate limiting on the parent domain or other subdomains. 

I am the systems administrator in charge of DNS under ticketswap.pizza and am authorized by the domain owner to request this change. As proof of domain control, I've added a TXT record under the registered domain:

```
dig txt _psl.ticketswap.pizza.

;; ANSWER SECTION:
_psl.ticketswap.pizza.	257	IN	TXT	"https://github.com/publicsuffix/list/pull/350"

```

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -11875,6 +11875,10 @@ gdynia.pl
 med.pl
 sopot.pl
 
+// TicketSwap : https://www.ticketswap.com
+// Submitted by Ruud Kamphuis <ruud@ticketswap.com>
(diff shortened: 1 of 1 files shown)


sleevi:
This does not sound like it reaches the bar for inclusion within the public suffix list (in that it's internal only), and because Let's Encrypt rate limit requests should now use the Let's Encrypt form.

PR #353, opened 2016-12-14
Add inf.ua, ltd.ua and cc.ua to PRIVATE section

1gbllc:
We provide thousands of our (and not only our) users with domain names in .inf.ua, .ltd.ua and .cc.ua zones. So we belive these domains should be added to PRIVATE section of public suffix list. Thank you.


sleevi:
Validated via TXT record and the structure of domains as observed in search results.

PR #356, opened 2016-12-15
Add wdtest6.com,wdtest7.com and wdtest8.com; TXTs included in the DNS…

JungSJin:
… record


sleevi:
Please update this request to conform to the [Guidelines](https://github.com/publicsuffix/list/wiki/Guidelines)

Notably: For private domains, please include a description of why these domains are being added.

Note that given that the Public Suffix List is a finite resource with size and memory constraints on where it is used, as a matter of policy, 'testing' domains are not generally accepted or encouraged.

JungSJin:
Follow-up:
 Description of why these are being added: we provide a customer with remote access to Network Attached Storage(NAS). These domains we want to add to PSL will be used for NAS only for remote access. Yes these domains include "test" string, but they will be used for production as well(sorry for confusion, but domain name being used for NAS is not directly exposed to a customer so it should be ok even though it contains "test" string). One of major client apps for remote access is web browsers. So we want to take advantage of PSL which meets the needs of browser manufacturers as well as relaxing the limit of Let's Encrypt if possible.

JungSJin:
Hello,

I updated PR as below. I am wondering if you can review this at your earliest chance. I really appreciated your time reviewing this.

Follow-up:
Description of why these are being added: we provide a customer with remote access to Network Attached Storage(NAS). These domains we want to add to PSL will be used for NAS only for remote access. Yes these domains include "test" string, but they will be used for production as well(sorry for confusion, but domain name being used for NAS is not directly exposed to a customer so it should be ok even though it contains "test" string). One of major client apps for remote access is web browsers. So we want to take advantage of PSL which meets the needs of browser manufacturers.

sleevi:
I'm not sure this meets the bar for inclusion in the PSL.

It sounds like your goal is to get around Let's Encrypt rate limits, and based on that, I'm closing this as WontFix, as that's not an appropriate use of the PSL. For that, please contact Let's Encrypt, as per https://letsencrypt.org/docs/rate-limits/

Given that the Public Suffix List is a finite resource that needs to run in a variety of client environments, including multiple similar domains is non-ideal. A better structure would be to utilize a single domain (or a wildcard, as appropriate).

JungSJin:
Hello,
Sorry for any confusion I gave you. We are not trying to "get around" the limit but have them be correctly applied to a user's own domain (i.e. joe.wdtest6.com)  that we give them for a server(AKA NAS) running in their home. Think about a blog site, where a joe.blogspot.com is setup and blogspot.com is on the PSL, I don't see how our case is any different. 

If you are concern about we made PR for multiple domains, then we can re-request PR for ONLY one domain. Please get us wrong we do not misuse PSL.

Also Synology which already added multiple domains to PSL is in the same industry. Their case would not be different from us. I'd like to thank you for your review one more time.

JungSJin:
Hello,

I am Jung from Western Digital. Sorry for any confusion I gave you. We are not trying to "get around" the limit but have them be correctly applied to a user's own domain (i.e. joe.wdtest6.com)  that we give them for a server(AKA NAS) running in their home. Think about a blog site, where a joe.blogspot.com is setup and blogspot.com is on the PLS, I don't see how our case is any different.

If you are concern about we made PR for multiple domains, then we can re-request PR for ONLY one domain. Please get us wrong we do not misuse PSL.

Also Synology which already added multiple domains to PSL is in the same industry. Their case would not be different from us. I'd like to thank you for your review one more time.

sleevi:
If the purpose of the request is for Let's Encrypt - which it seems the request is (and your latest email doesn't provide additional reasoning) - then using the Let's Encrypt rate limit is appropriate.

Importantly, you noted earlier that the domain name is not exposed to the customer - suggesting that the only purpose for the PSL is Let's Encrypt rate limits.

I do want to make sure we're properly understanding your use case, but simply adding it to the PSL to increase the Let's Encrypt rate limits (or, if you would prefer it phrased, to have them attributed to the individual user), or because other companies are on, isn't itself a good reason.

In the event there's some miscommunication here:
Does a user directly load a URL hosted under [whatever].wdtest6.com which contains content that they themselves control (e.g. that is more than a pre-defined management interface managed by Western Digital, for example, as part of the Firmware)?

sleevi:
(Any additional documentation about your use case can also help understand and process this request)

JungSJin:
Sorry for confusion, it was my bad.  YES, a user directly load a URL hosted under designated domain to access to its own NAS in their home from either LAN or WAN(Internet). It is centralized personal cloud storage system, so all contents on NAS is controlled by a user, not by us. They can publicly share their contents or do private share over contents with anyone as well.

Due to NDA, I cannot open DOCs widely, but I found the good link( even though pretty old) to help you understand our production. 

http://www.excitingip.net/2594/entry-level-nas-storage-wd-my-cloud-personal-cloud 

Please find the connectivity arch of NAS from the link.  After review, if you think you can add our domain to PSL, I can make another PR for only one domain which would be "remotewd.com". Originally we want to add multiple domains for multiple productions, but as you explained, I agree we need to make the PSL simpler and lighter as much as we can. But I wish you can expedite our new PR if you can. I really appreciate your time reviewing our request.

sleevi:
(For other maintainers) Additional details (though limited) include https://www.wdc.com/content/dam/wdc/website/downloadable_assets/eng/user_manual/4779-705145.pdf

I suppose to help expedite/process this request, if you could share some 'sample' domains and their structure, we can help develop the ideal proposed rule to maximally secure and protect your users.

JungSJin:
The example domains would be device-123.remotewd.com, device-456.remotewd.com. As mentioned previously, this time we want to add " remotewd.com" only.. regarding its structure one of our developers will help on that. Will update this thread as soon as possible.

sleevi:
It does sound like `remotewd.com` is the desired rule, which will ensure that `foo.remotewd.com` and `bar.remotewd.com` are treated separately.

If you want to update the pull request and setup the TXT record for validation, per the [Guidelines](https://github.com/publicsuffix/list/wiki/Guidelines), we'd be happy to review during the next batch of updates.

JungSJin:
Will do. Thanks for help

JungSJin:
FYI, made the pull request for remotewd.com (https://github.com/publicsuffix/list/pull/375)

PR #358, opened 2016-12-27
Add un-ip.org to private section

llyt:
un-ip.org is a DNS wildcard and can be used by anyone.

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -11930,6 +11930,10 @@ inc.hk
 // Submitted by Ed Moore <Ed.Moore@lib.de.us>
 lib.de.us
 
+// un-IP.org : http://un-ip.org/
+// Submitted by Mounir Ahmina <hello@moun.ir>
+un-ip.org
+
 // Viprinet Europe GmbH : http://www.viprinet.com
 // Submitted by Simon Kissel <hostmaster@viprinet.com>
 router.management
@@ -11960,9 +11964,5 @@ za.org
 // Zeit, Inc. : https://zeit.domains/
 // Submitted by Olli Vanhoja <olli@zeit.co>
 now.sh
-
-// un-IP.org : http://un-ip.org/
-// Submitted by Mounir Ahmina <hello@moun.ir>
-un-ip.org
 
 // ===END PRIVATE DOMAINS===
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -11961,4 +11961,8 @@ za.org
 // Submitted by Olli Vanhoja <olli@zeit.co>
 now.sh
 
+// un-IP.org : http://un-ip.org/
+// Submitted by Mounir Ahmina <hello@moun.ir>
+un-ip.org
+
 // ===END PRIVATE DOMAINS===


weppos:
@llyt are you the owner of the domain?

llyt:
@weppos, Yes. the TXT record proof is referring to this link now (sorry I forget to change it).

llyt:
Hello @weppos 

I just wanna make sure that everything is following the guidelines...

The TXT record:

```
:~$ dig TXT _psl.un-ip.org
; <<>> DiG 9.9.5-3ubuntu0.10-Ubuntu <<>> TXT _psl.un-ip.org
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 1574

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 4096
;; QUESTION SECTION:
;_psl.un-ip.org.			IN	TXT

;; ANSWER SECTION:
_psl.un-ip.org.		86341	IN	TXT	"https://github.com/publicsuffix/list/pull/358"

;; Query time: 33 msec
;; SERVER: 127.0.1.1#53(127.0.1.1)
;; WHEN: Mon Jan 02 17:48:13 WET 2017
;; MSG SIZE  rcvd: 101
```

I've relocated the submission according to the alphabetic order too.

llyt:
Now I thought about it... should it be a wildcard (in the list)?

By the way you can see what kind of thing the service is in the home page: http://un-ip.org/.

My real goal with the service (not mentioned in the home page yet) is to easily allow servers to use WSS using only their "IP", so I will imagine it to be used on production and would need browsers and Let's Encrypt dealing with every subdomain/IP as owned by separated person, not the last part of the IP...

Interested to hear you opinions...

sleevi:
The public suffix list does not support your use case. A wildcard only applies to a single label, and presently, does not support multiple wildcards.

So you if you did `*.un-ip.org`, then both `192.168.0.1.un-ip.org` and `10.0.0.1.un-ip.org` would be able to set cookies for `0.1.un-ip.org`.

Similarly, `example.10.0.0.1.un-ip.org` and `www.10.0.0.1.un-ip.org` would be able to set cookies for `10.0.0.1.un-ip.org`, `0.0.1.un-ip.org`, `0.1.un-ip.org`

There's no support in any of the extant applications or libraries to safely support your use case. I'm sorry.

llyt:
Ahh, thanks @sleevi 

 I will change it to be in the form of `192-168-0-1.un-ip.org`... it makes more since that way too...

I will notice the maintainers when it's ready.

llyt:
It's ready and works as intended now. I hope everything is following the guidelines.

sleevi:
@gerv @weppos How do you feel about merging this?

This meets https://github.com/publicsuffix/list/wiki/Guidelines , but it's so fundamentally insecure that I'm not sure it's worth adding, because adding it to the PSL isn't going to help anyone be more secure (and may lead to all sorts of 'shared-domain' certificates, given the example).

That is, my own take is I would **not** like this to be approved, which the Guidelines call out as a distinct process, but I don't want to be the only one blocking it, and want to ensure it's appropriate for everyone.

gerv:
I am happy to make a PSL policy that we don't admit services like this. It seems to me that people in full control of their IP addresses are probably also in control of their DNS, and so can set themselves up with a domain name. People who need a service like this to get a domain name are probably not in full control of their IP address - and that makes this insecure and dangerous. 

I think that if LE want to support this use case they can add it to their exceptions list using their form. And if they don't like the idea, we shouldn't either.

weppos:
@sleevi @gerv it's fine for me, but we need to find a sort of clear definition of this use case to be added to the guideline. Something that both us and the submitters can use as a check point to determine if the request will be accepted or not.

Given both of you are native english speaker, may I ask you to come with a draft of it? Feel free to edit the guidelines directly.

gerv:
@weppos: Done.

weppos:
Rejected per https://github.com/publicsuffix/list/wiki/Guidelines#validation

PR #360, opened 2016-12-30
Add official public suffixes for .bd domain

alexott:
Per https://en.wikipedia.org/wiki/.bd, the .bd domain includes following
official subdomains: .com.bd, .net.bd, ..., while they are omitted from
existing PSL.

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -300,7 +300,7 @@ ac.bd
 com.bd
 edu.bd
 net.bd
-org.bg
+org.bd
 // gov.bd - registration not allowed
 // mil.bd - registration not allowed
 
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -295,7 +295,14 @@ store.bb
 tv.bb
 
 // bd : https://en.wikipedia.org/wiki/.bd
-*.bd
+bd
+ac.bd
+com.bd
+edu.bd
+net.bd
+org.bg
+// gov.bd - registration not allowed
+// mil.bd - registration not allowed
 
 // be : https://en.wikipedia.org/wiki/.be
 // Confirmed by registry <tech@dns.be> 2008-06-08
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -294,15 +294,18 @@ org.bb
 store.bb
 tv.bb
 
-// bd : https://en.wikipedia.org/wiki/.bd
-bd
+// bd : http://www.registry.com.bd/extensions.aspx
 ac.bd
+co.bd
 com.bd
 edu.bd
+gov.bd
+info.bd
+mil.bd
 net.bd
 org.bd
-// gov.bd - registration not allowed
-// mil.bd - registration not allowed
+tv.bd
+ws.bd
 
 // be : https://en.wikipedia.org/wiki/.be
 // Confirmed by registry <tech@dns.be> 2008-06-08


weppos:
Thanks for the report @alexott. Unfortunately, we don't consider Wikipedia (anymore) an authoritative source to perform changes to IANA TLD.

Do you have some sort of official documentation from the .BD registry?

weppos:
Nevermind, I found http://www.registry.com.bd/extensions.aspx that lists all the suffixes.
It looks like the Wikipedia page is outdated (that's an example of why we can't rely on Wikipedia for these changes).

I'll go ahead and update the list.

weppos:
I also see you added `bd`, but the registry says:

> BTCL currently does not allow registering a domain directly under Dot BD. That must be a subdomain under several Dot BD extensions. For example "abc.bd" for a certain company "abc" is not allowed. But a domain of the form "abc.com.bd" or "abc.net.bd" is valid.

alexott:
Thank you very much for quick reply! I'm new to PSL, and may not able to do everything correct yet :-)

I added `bd` because there was an entry `*.bd`, and I understood that it should match to any domain under `.bd` TLD.

Another question - I'm planning to submit more entries for "not well know domains", such as, `.kh`, `.mm`, etc. - should I make separate pull requests for every of them, or it would be ok to make one PR?  For these domains, I know that at least, `.com` subdomain exists, as we found when analyzing Google's sites

sleevi:
@alexott I appreciate you filing these pull requests, but I am very inclined to reject them all, especially without direct request from the registry operator.

As the public suffix list is a finite resource with size constraints that, unfortunately, needs to be used effectively with every HTTP client supporting cookies (i.e. virtually all of them), we should desire to be efficient whenever possible. The current `*.bd` logic is unquestionably overly restrictive, but to date, it's not caused any reported bugs. I'm much more inclined to believe we should take such requests on a case-by-case basis from the registry directly, and with routine communication with them.

Understandably, .ccTLDs are into all sorts of 'vanity' eTLDs (like these), trying to appear like first-level gTLDs within the respective gTLD. As such, the list of what a given ccTLD operator offers changes with whatever whims the registry or registrars feel like - and I don't think it's good for the ecosystem to try to chase them ourselves.

However, I'm curious @gerv and @weppos's thoughts here - I missed that you had opened a number of PRs all doing effectively the same thing (expanding wildcards), otherwise, I would have responded sooner.

alexott:
I understand your concerns regarding cookies, and http clients. I simply
confused by inconsistencies in the list - some TLDs have subdomains because
registrants took care fore them, but some aren't - for which I filed
requests...

P.S. Maybe I misunderstood the purpose of the PSL... sorry for noise

sleevi:
@alexott I think you were fine for submitting them - the inconsistencies in policies and domains is a reasonable point of concern/confusion, and we're still working out the ideal policies to maintain list growth - as well as better understand the concrete size concerns from consumers, so that we can make sure the needs of consumers are balanced with the needs of the registries. Because of this, it's much easier/safer to adopt a conservative approach towards growth, and to treat requests from registries/domain holders rather than just anyone. This is especially true with all of the Let's Encrypt-induced list growth doubling the size in the past year.

This does mean, in practice, that it does not represent a comprehensive absolute set of eTLDs - more like a rough set of heuristics, unfortunately, which we refine when necessary - largely at the request of domain holders and registries, while making sure requestors understand the concerns :)

weppos:
@sleevi I recall that a few years ago I asked you and @gerv whether the expanded suffixes was preferred over the wildcard, and you mentioned that at Google you preferred the expanded list for some auto-complete (if I remember correctly) reasons.

Has the need changed?

I don't have a specific preference between wildcard or expanded version. Expanded is more accurate, but it also requires more maintenance and without registry support it may eventually become a maintenance problem.

gerv:
I'm happy with us retaining and using wildcards unless they prove to be inaccurate.

PR #363, opened 2016-12-30
add subdomains for .fj TLD

alexott:
As per http://domains.fj/public/register.php


sleevi:
Closing as WontFix per https://github.com/publicsuffix/list/pull/360#issuecomment-269956646

PR #367, opened 2016-12-30
add subdomains for .pg TLD

alexott:
as per http://www.unitech.ac.pg/InformationAbout/Services/ICTS/DomainRegistry

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -5316,7 +5316,13 @@ org.pf
 edu.pf
 
 // pg : https://en.wikipedia.org/wiki/.pg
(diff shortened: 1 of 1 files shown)


sleevi:
I'm curious @weppos 's thoughts, but I believe we should reject this change, unless it comes from the registrar.

I can appreciate the more refined accuracy this pull request provides, however, for efficiency sake, I don't believe we should be expanding the wildcard to more specific domains.

sleevi:
Closing as WontFix per https://github.com/publicsuffix/list/pull/360#issuecomment-269956646

PR #368, opened 2016-12-30
add subdomains for .zw TLD

alexott:
.co.zw is handled by http://www.zispa.org.zw/
.ac.zw is at least exists at http://www.uz.ac.zw/
.org.zw is at least registrant itself...


sleevi:
You state it is for .zw, but you're removing .pg

Closing this out.

PR #375, opened 2017-01-05
added remotewd.com

JungSJin:


sleevi:
For historic reference, the justification for this is from https://github.com/publicsuffix/list/pull/356#issuecomment-269579177

PR #385, opened 2017-01-17
Add testurl.cz domain suffix

michalkolar:
We can use *.testurl.cz for our customers projects.

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -11849,4 +11849,8 @@ cc.ua
 inf.ua
 ltd.ua
 
+// WVG : https://www.wvg.cz/
+// Submitted by Michal Kolar <michalkolar@wvg.cz>
+testurl.cz
+
 // ===END PRIVATE DOMAINS===
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -11820,6 +11820,10 @@ remotewd.com
 // Submitted by Yuvi Panda <yuvipanda@wikimedia.org>
 wmflabs.org
 
+// WVG : https://www.wvg.cz/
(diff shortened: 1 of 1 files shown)


michalkolar:
dig TXT _psl.testurl.cz
https://github.com/publicsuffix/list/pull/385

sleevi:
In general, we've declined including "testing" domains, to ensure that the list is useful for everyone and the size does not grow unbounded - can you explain why/how testurl.cz is not a testing domain?

michalkolar:
Testurl project will, for people who do not have their domain and want to try to create a website. I see no reason why it should not go.

sleevi:
The Public Suffix List is (effectively) a finite resource, as all clients wishing to use cookies effectively need to know about all domains. So we're reticent to add domains that are not in the general good for browser users. That is, if it's just a companies testing site, it's questionable.

Beyond that, can you ensure the company name is correct? It looks like the Company Name is "WVG Solutions, s.r.o" - is that correct?

michalkolar:
I understand, we try for a year to add to the list.

Yes, the company is called WVG Solution s.r.o.

sleevi:
@weppos They left editing enabled, so I have resorted and rebased. The outstanding question is whether "testurl" is an appropriate use of the finite resource, given that we've also declined inclusion for things like 'internal testing domains'. Thoughts?

sleevi:
@weppos Ping? I would suggest we close for testing URLs.

weppos:
I agree @sleevi, at this time I don't see this URL a valid option for inclusion just for the fact of being a test URL. Test URLs may also not have the same security requirements of other URLs.

Unless further information that proves this is a production domain are provided, I would not include this domain in the list.

PR #386, opened 2017-01-18
Add onion

### LINKED ISSUE #374: Add .onion
RFC 7686 officially passed through the IETF, which lists .onion as a Special-Use Domain Name

Unlike those names specified in the IANA Special Use list ( https://www.iana.org/assignments/special-use-domain-names/special-use-domain-names.xhtml ), the set of .onion domains extend beyond local resolutions. That is, two clients, both configured to recognize .onion names, both resolve the name to the same party. This is different than, say, .localhost or .example, which will resolve to the local machine.

At present, .onion is captured by the existing implicit '*' in the algorithm, and so adding .onion would not have a meaningful change to the PSL, except in that it will recognize it as a domain 'blessed' by IANA/ICANN, for applications which use the PSL for that purpose.

As noted, "Note that the registration of .onion names does not prohibit IANA from inserting a record into the root zone database to reserve the name."

I'm hoping to use this bug to capture any reasons why it may be desirable to *not* add .onion

sleevi:
While "onion" does not appear within the root zone database, per RFC 7686, it is set aside as a special-use domain name, and explicitly called out as suitable for inclusion by IANA within the root zone database. Unlike other special-use domain names at this time, onion names that comply with RFC 7686 meet the definition of ICP-3 for both community process and a globally unique namespace (through being bound to keys in a process that ensures two clients end up with the same resolution).

Closes #374

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -5280,6 +5280,9 @@ net.om
 org.om
 pro.om
 
+// onion : https://tools.ietf.org/html/rfc7686
+onion
+
 // org : https://en.wikipedia.org/wiki/.org
 org
 


dandiko:
Baver

dnsguru:
Locking this conversation.  If there are comments on this file them as issues and reference this pull request number.

PR #387, opened 2017-01-21
Removing s3.amazonaws.com wildcard, adding AWS eu-west-2 regional end…

lawells:
…points

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -10553,7 +10553,7 @@ elasticbeanstalk.cn-north-1.amazonaws.com.cn
 
 // Amazon S3 : https://aws.amazon.com/s3/
 // Submitted by Luke Wells <psl-maintainers@amazon.com>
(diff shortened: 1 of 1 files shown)


lawells:
ping? @sleevi @weppos

sleevi:
@lawells Sorry, this ended up falling through the cracks because none of the domains were auto-verifiable using DNS (per [Guidelines](https://github.com/publicsuffix/list/wiki/Guidelines) ).

The e-mail authentication we used to authenticate your previous PRs has the downside of effectively needing to check every PR that you're still authenticated to submit patches (which is why DNS is so much better). Considering the previous validations done, can you just send an email to sleevi@google.com using your corporate email, and I'll double check the DKIM/SPF as equivalent for continued authorization?

Future changes will certainly go smoother with the TXT record :)

sleevi:
Confirmed via email with psl-maintainers

PR #391, opened 2017-02-01
Add ipna.me to the private section.

bemasc:
[ipna.me](https://ipna.me) is an experimental service whose main purpose is to maximize use of TLS on the web.  If you are deploying a server with a static IP address, ipna.me makes it easy for you to acquire a certificate from a vendor like Let's Encrypt.  (I have spoken with Let's Encrypt, and they've confirmed that ipna.me is  an acceptable use of their service.)

Here's an example (empty) test server: https://104-131-62-56.ipna.me/

Adding ipna.me to the public suffix list will help ensure that users of ipna.me are correctly isolated from each other.

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -11423,6 +11423,10 @@ se.leg.br
 sp.leg.br
 to.leg.br
 
+// IPName : https://ipna.me/
+// Submitted by Ben Schwartz <bemasc@ipna.me>
+ipna.me
+
 // Joyent : https://www.joyent.com/
 // Submitted by Brian Bennett <brian.bennett@joyent.com>
 *.triton.zone


sleevi:
While not wanting to turn the PSL into a review of peoples' services, I am curious about the necessity of your service. CAs can totally issue for IP addresses; meaning Let's Encrypt could as well. The Baseline Requirements (which CAs follow) fully permits such issuance.

Is this something that perhaps talking to Let's Encrypt would be able to resolve, by allowing such certificates to be obtained? Curious for @jsha or @bdaehlie thoughts on whether something like that might be doable without requiring this addition.

Independent of that discussion, please make sure your pull request meets the [Guidelines](https://github.com/publicsuffix/list/wiki/Guidelines)

bemasc:
I can't speak for Let's Encrypt, but they've publicly [stated](https://community.letsencrypt.org/t/certificate-for-public-ip-without-domain-name/6082/7) that they don't issue IP certs, and aren't planning to start.  I agree with your analysis; this service is only worthwhile because they don't.

As for the guidelines, `make test` passes, and I believe the TXT records are in order.

jsha:
Thanks for tagging us. We're talking about it internally. Thoughts so far:

 - Since ACME currently supports only a "DNS" identifier type, it would have to be extended to include an "IP" type, which would put any implementation out some time.
 - The main advantage in my mind to doing so would be to obviate services like `ipna.me`. A big downside of this type of service is that `ipna.me` could choose at any time to resolve `104-131-62-56.ipna.me` to their own servers, and issue their own cert, fully MITM'ing the connection, including getting cookies. It would be nice to steer people away from having an extra party in the mix that could MITM them.
 - The main disadvantage, IMO: IP ownership times are frequently measured in hours (for cloud service providers), as opposed to usually months to years for DNS names. So certs issued to IP addresses would frequently become inaccurate when the IP address owner changed.

gerv:
> The main disadvantage, IMO: IP ownership times are frequently measured in hours (for cloud service providers), as opposed to usually months to years for DNS names. So certs issued to IP addresses would frequently become inaccurate when the IP address owner changed.

That seems to be a bit "caveat emptor" doesn't it? Don't buy a cert for an IP address that isn't yours for the long term (and/or hasn't been consistently yours for some period of time beforehand). If you do, you risk someone impersonating your service.

jsha:
> That seems to be a bit "caveat emptor" doesn't it? Don't buy a cert for an IP address that isn't yours for the long term (and/or hasn't been consistently yours for some period of time beforehand). If you do, you risk someone impersonating your service.

To be pedantic: Don't tell your users to rely on an https:// URL for your service that points at an IP address, unless you have owned that IP address for longer than the max BR certificate lifetime + validation lifetime, since you risk someone else holding a certificate or validation for that IP address.

That said, you're right that it's somewhat caveat emptor, though there is some degree to which supporting the pattern encourages the behavior. (also, getting somewhat off-topic here - if you'd like to unclutter this issue, happy to move to https://community.letsencrypt.org/).

bemasc:
> The main advantage in my mind to doing so would be to obviate services like ipna.me. A big downside of this type of service is that ipna.me could choose at any time to resolve 104-131-62-56.ipna.me to their own servers, and issue their own cert, fully MITM'ing the connection, including getting cookies.

This is true, and I would welcome any change that makes ipna.me unnecessary, but it applies equally to any dynamic DNS provider, including those already in the PSL, so I hope it won't block this pull request.

sleevi:
For clarity: None of the questions were meant to block this request, just to make sure to understand the request, scope of rules, and see if there are alternative needs.

Requests are normally processed in batches as volunteer time permits, starting with the oldest request, and working through those that are conforming to the Guidelines. If you've got everything in order, then it should be processed during the next batch of reviews.

bemasc:
OK, good to know.  Thanks!

sleevi:
@gerv @weppos Similar to Issue #358 , I'm not sure how we should proceed with this.

sleevi:
Closing as WontFix, per https://github.com/publicsuffix/list/wiki/Guidelines

PR #394, opened 2017-02-06
Full list of public domains of zone .RU

MagnusVR:
Full list of public domains of zone .RU according to https://cctld.ru/en/domains/domens_ru/reserved.php


weppos:
This PR is incorrect. The domains you references were demoted from reserved in October 2009 (as mentioned in the link you references).

PR #402, opened 2017-02-19
Add landcement.com to private section.

ZE3kr:
We use foo.landcement.com for customer. The TXT record has already created.

```
$ dig TXT _psl.landcement.com.
_psl.landcement.com.	604800	IN	TXT	"https://github.com/publicsuffix/list/pull/402"
```

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -11888,7 +11888,6 @@ gdynia.pl
 med.pl
 sopot.pl
 
-
 // TlOxygen : https://tlo.xyz
 // Submitted by Guo Zeyu <ze3kr@tlo.xyz>
 landcement.com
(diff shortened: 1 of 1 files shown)


sleevi:
Please review https://github.com/publicsuffix/list/wiki/Guidelines for this request

1) Sort by company name
2) Can you provide example domains for customers to help make sure this request is correct?

ZE3kr:
I already sort the list by the company name and create an example domain: https://example.landcement.com/

Is that necessary to provide the domains for customers? It might contains sensitive data of customers.

ZE3kr:
Because thrid level domain needs to set cookie to second level domain is needed for security check, so it cannot be added to PSL.

PR #408, opened 2017-03-01
Update .ไทย domains

sarunint:
Refer to item 5.10.1 of https://www.thnic.co.th/doc/dotthai_policy_pp20110107[draft][en].pdf

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -6731,6 +6731,12 @@ yt
 // xn--o3cw4h ("Thai", Thai) : TH
 // http://www.thnic.co.th
 ไทย
+ศึกษา.ไทย
+ธุรกิจ.ไทย
+รัฐบาล.ไทย
+ทหาร.ไทย
+เน็ต.ไทย
+องค์กร.ไทย
 
 // xn--pgbs0dh ("Tunisia", Arabic) : TN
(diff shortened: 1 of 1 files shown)


sleevi:
Are you the registrar? Per our [Guidelines](https://github.com/publicsuffix/list/wiki/Guidelines) we try to only accept requests from the domain operator, requiring that they demonstrate such control.

@weppos @gerv Thoughts on handling some of the ccTLD cases?

sarunint:
No, I'm not the registrar. But according to the wiki

> Changes here need to either come from a representative of the registry (authenticated in a similar manner to below) **or be from public sources such as a registry website**.

weppos:
@sleevi it looks like the subzones are officially supported. I was able to locate them also in the whois at https://www.thnic.co.th/en/home/

> Thoughts on handling some of the ccTLD cases?

Not sure what you mean. We have currently subzones of ccTLDs.

sleevi:
Apologies @weppos for not being more specific. It was more generally the policies related to handling ccTLDs and whether they're registry-reserved or registrar-operated-and-reserved.

However, since it's a ccTLD, I suppose the best answer is "just accept it", even if the ccTLD registry itself has not reserved them.

PR #407, opened 2017-02-27
Add vapor.cloud to list

joscdk:
Adding vapor.cloud and vaporcloud.io to the list.

Clients are going to create subdomains, and we are providing Let's Encrypt certificates. I can't explain much here because of NDA, for further details, contact me directly.

Example domains:

subdomain1.vapor.cloud
subdomain2.vapor.cloud
sub1.subdomains1.vapor.cloud
sub2.subdomains2.vapor.cloud

The same goes for vaporcloud.io

PR #398, opened 2017-02-13
add private domain in public suffix

cschang2:
we will use letsencrypt to give us free domain users

submitted by whois contact email
we have added TXT records

[root@r310 kencorp]#
[root@r310 kencorp]# dig txt _psl.url.tw +short
"https://github.com/publicsuffix/list/pull/398"
[root@r310 kencorp]# dig txt _psl.mymailer.com.tw +short
"https://github.com/publicsuffix/list/pull/398"
[root@r310 kencorp]# dig txt _psl.twmail.cc +short
"https://github.com/publicsuffix/list/pull/398"
[root@r310 kencorp]# dig txt _psl.twmail.net +short
"https://github.com/publicsuffix/list/pull/398"
[root@r310 kencorp]# dig txt _psl.twmail.org +short
"https://github.com/publicsuffix/list/pull/398"

and this passed Travis-CI check  
we have set the status checks. But for the old pull request no way to show

https://travis-ci.org/publicsuffix/list/builds/201746015

Is it possible to guide us? Is there a link that is why we did not notice and lead us to fail

PR #372, opened 2017-01-04
add CloudAccess.net free domains to Private

Pawel-Panek:
Domains used for free site provisioning.

PR #410, opened 2017-03-04
we use *.nodeart.io with lots of dynamic subdomains, most of them rep…

nosovk:
We use *.nodeart.io with lots of dynamic subdomains, most of them represent staging builds or different branch builds, or some internal services (like PBX, with personal cert for each device). We issue certificate to each of used subdomains. Unfortunately we need to use different certificates for each domain (wildcard is good, but not in our case). And we were suggested to add our domain to that list to issue more certificates at letsencrypt.

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -11607,6 +11607,10 @@ sytes.net
 webhop.me
 zapto.org
 
+// NodeArt : https://nodeart.io
+// Submitted by Konstantin Nosov <Nosov@nodeart.io>
+*.nodeart.io
+
 // NYC.mn : http://www.information.nyc.mn
 // Submitted by Matthew Brown <mattbrown@nyc.mn>
 nyc.mn


sleevi:
This does not sound like a valid reason to request inclusion, based on the
description.

If your goal is solely to raise Let's Encrypt issuance limits, please
contact Let's Encrypt, as they have a form for such requests on their
issuance limits page.

nosovk:
image
Added txt verification record.
Why not a reason? Any registered client can create any subdomain.
Also it was LE request to add site here,

weppos:
> Also it was LE request to add site here,

Can you please provide a reference to this specific suggestion? Was it in the forum?

nosovk:
Nope, it's on their site
end of third paragraph
purate limits - let s encrypt - free ssl_tls certificates - google chrome 2017-03-05 20 54 01

sleevi:
And if your only reason for requesting adding is LE, there's a form at the
bottom of that page to request limits be raised.

weppos:
Please read until the "Overrides" section in the same page:

> If you are a large hosting provider or organization working on a Let’s Encrypt integration, we have a [rate limiting form](https://docs.google.com/forms/d/e/1FAIpQLSfg56b_wLmUN7n-WWhbwReE11YoHXs_fpJyZcEjDaR69Q-kJQ/viewform?c=0&w=1) that can be used to request a higher rate limit. It takes a few weeks to process requests, so this form is not suitable if you just need to reset a rate limit faster than it resets on its own.

I will get in touch with them to see if they can rephrase the line you referenced so that developers won't be confused and ask to be included here only for Let's Encrypt rate limiting.

nosovk:
There is a bunch of pull requests connected with LE for now: https://github.com/publicsuffix/list/pull/407, https://github.com/publicsuffix/list/pull/398, https://github.com/publicsuffix/list/pull/372, https://github.com/publicsuffix/list/pull/391...
We also submitted form to LE, but there is no impact for now.
Now we are checking solution with psl. If we will not receive answer from LE in two weeks, we will retry our request to add our domain to psl.
Thanks for cooperation.

nosovk:
Unfortunately there is no response from LE. And rate limits are hitting me hard. Could you please accept pull request?

olgakozoriz:
Have the same issue with LE. It seems that publicsuffix decided not to merge such pull requests @weppos any change in terms of publicsuffix? 
@nosovk have you managed to contact LE somehow? They keep silence on my requests for more then 3 month :(

nosovk:
You may ping LE in their twitter https://twitter.com/letsencrypt
They do mot respond on mails etc.

sleevi:
Closing this out as WontFix, per the discussion it's related to Let's Encrypt rate limits, which is not a valid reason per https://github.com/publicsuffix/list/wiki/Guidelines

nosovk:
Strongly disagree. Probably you know that Letsencrypt ignores that form. It means that this repo is only be a one way to solve issue with certificate amount.
You just became bit more famous because of advertisement on LE site.
In my case adding to list is correct - because subdomains used by different clients with different cookie settings, different policies etc. I mean that adding to this list is correct, for browser security purposes.

sleevi:
While you may disagree, per the Guidelines, bypassing Let's Encrypt rate limits are not an appropriate reason.

PR #411, opened 2017-03-04
Add co.vu domain suffix

artas90:
Fixes getting certificates for domains provided by service http://codotvu.co


weppos:
@artas90 please read https://github.com/publicsuffix/list/wiki/Guidelines. THis is a private suffix, hence only the owner can request the inclusion and in the PRIVATE section, not ICANN.

PR #412, opened 2017-03-10
Pull request for adding org.ru, net.ru and pp.ru names

gauzy-gossamer:
Pull request for adding org.ru, net.ru and pp.ru names

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -11892,4 +11892,10 @@ cc.ua
 inf.ua
 ltd.ua
 
+// MSK-IX : https://www.msk-ix.ru/
+// Submitted by Khannanov Roman <khannanov@msk-ix.ru>
+org.ru
+net.ru
+pp.ru
+
 // ===END PRIVATE DOMAINS===
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -11707,7 +11707,6 @@ cloudapp.net
 // Submitted by glob <glob@mozilla.com>
 bmoattachments.org
 
-
 // MSK-IX : https://www.msk-ix.ru/
 // Submitted by Khannanov Roman <r.khannanov@msk-ix.ru>
 net.ru
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -11502,6 +11502,12 @@ cloudapp.net
 // Submitted by glob <glob@mozilla.com>
 bmoattachments.org
 
+// MSK-IX : https://www.msk-ix.ru/
+// Submitted by Khannanov Roman <r.khannanov@msk-ix.ru>
+net.ru
+org.ru
+pp.ru
+
 // Neustar Inc.
 // Submitted by Trung Tran <Trung.Tran@neustar.biz>
 4u.com
@@ -11891,11 +11897,5 @@ now.sh
 cc.ua
 inf.ua
 ltd.ua
-
-// MSK-IX : https://www.msk-ix.ru/
-// Submitted by Khannanov Roman <r.khannanov@msk-ix.ru>
-org.ru
-net.ru
-pp.ru
 
 // ===END PRIVATE DOMAINS===
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -11893,7 +11893,7 @@ inf.ua
 ltd.ua
 
 // MSK-IX : https://www.msk-ix.ru/
-// Submitted by Khannanov Roman <khannanov@msk-ix.ru>
+// Submitted by Khannanov Roman <r.khannanov@msk-ix.ru>
 org.ru
 net.ru
 pp.ru


gauzy-gossamer:
I will post here later when TXT records are added.

sleevi:
@gauzy-gossamer Ping

qk4l:
@sleevi , do you have anything that stops this PR?

I see that @gauzy-gossamer mentioned some TXT records (which one?) but they should not related to the list.

_org.ru_  is common domain for non commercial organizations with thousands of sub domains ([stat](http://net.ru/stat/?stat_type=registrars&zone=org.ru)).

gauzy-gossamer:
@sleevi , as far as I can tell, the administators couldn't add these records, because the zone is generated with the registry that doesn't allow records with an underscore. I'm not sure if they'd want to make changes in this process to allow these records, I'll ask what they decided.

gauzy-gossamer:
@sleevi, the records were added:
dig _psl.org.ru -t TXT
_psl.org.ru.		345600	IN	TXT	"https://github.com/publicsuffix/list/pull/412"

dig _psl.pp.ru -t TXT
_psl.pp.ru.		345600	IN	TXT	"https://github.com/publicsuffix/list/pull/412"

dig _psl.net.ru -t TXT
_psl.net.ru.		345600	IN	TXT	"https://github.com/publicsuffix/list/pull/412"

weppos:
@gauzy-gossamer thanks, the request is validated.

However, before proceeding, per our [Guidelines](https://github.com/publicsuffix/list/wiki/Guidelines), please sort the list of suffixes properly, move the section in the proper position based on the alphabetical order of the company, and provide a reason for listing.

gauzy-gossamer:
OK, I moved it according to the alphabetic order. I'm not sure what the reason for listing should be though. Originally, we were requested to add domains to this list, so they can be used with letsencrypt.

weppos:
> I'm not sure what the reason for listing should be though.

This is actually quite concerning to us. Being in this list has several implications (including the possibility of breaking entire sites due to how the browsers will treat a domain), therefore you MUST be absolutely aware of the implications before requesting to be listed.

If the reason is purely for Let's Encrypt, then this is not the correct approach. Let's Encrypt provides an "override" form that you can use to request the override of the limits.
https://letsencrypt.org/docs/rate-limits/#overrides

gauzy-gossamer:
These domains are available for public registration, and every third level domain in these zones is used independently, so there shouldn't be any problem from this perspective. From what I understand, they were already in this list and were only removed recently - with this commit https://github.com/publicsuffix/list/pull/359/ . 
However, we only learned about it when one of the users in these domains experienced problems with letsencrypt, hence the reason I provided.

weppos:
@gauzy-gossamer as I said, this is not a valid assumption, nor interpretation. Moreover, the list exists beyond Let's Encrypt and the fact you don't seem to have received any other complain than the ones Let's Encrypt related should be a clear evidence that perhaps you may want to address the issue with Let's Encrypt instead of this list.

I am not saying the request is invalid. I am saying that for now you haven't provided any valid reason for approving the listing request, based on the purpose of the PSL itself. And yes, the domain was listed, and it was removed exactly because of possible implications based on the fact it they were no longer reserved .RU suffixes.

I want you to be fully aware of what you are requesting, as it seems to me that you are not very familiar with the implications and goal of the PSL, instead you are driven by third party requests.

Again, if you fully understand the implications, and accept the responsibility for the request, please provide an accurate reason and I will be happy to merge the request.

gauzy-gossamer:
As far as I can tell, the primary goal of PSL is to maintain the list of zones in which users can directly register domains, so that browsers can prevent setting cookies to higher level names, and some other related cases. And from this point of view org.ru, net.ru and pp.ru qualify for the inclusion - you can register domains in these zones through several russian domain registrars including www.reg.com, and nic.ru/en/ . 

I understand that there are some other usages of PSL, including Let's Encrypt, and I probably shouldn't have mentioned it, but it seemed to me that how these zones are used is more important for this request, than the motivation we had to initialize it.

a-pog:
Last year's more than a month, as the initiator commented on his decision. Is it planned to satisfy the request?

ForNeVeR:
Well, if browsers actually will treat all org.ru subdomains as children of one big "org.ru" domain, and will allow these subdomains to share, read and/or set their cookies and any other information, then **that creates a big security breach**, because that's an actual public suffix, and its' subdomains aren't related to each other.

If my understanding here is correct, I believe that's an additional reason to accept this PR rather than to decline it.

@weppos, I've read the PSL guidelines and I believe that all the technical requirements are satisfied. If you require a clearly stated reason to add these domains, then here it is: 

**The `net.ru`, `org.ru`, and `pp.ru` suffixes are public suffixes, and the domains in these zones could be delegated to mutually-untrusting parties. It would be better to consider sites in these zones as separate entities for any purposes (including, but not limited to browser usage and certificate validation).**

nosovk:
Just checked - I was able to set cookie to .org.ru domain itself from sub-domain site.
Now I know how to set cross domain cookies :)
Really didn't thought that it's possible, but yes - looking for public domains absent in that list - is new attack vector.

Also found - https://community.letsencrypt.org/t/publicsuffix-list-should-be-replaced-with-dedicated-solution/35549

gauzy-gossamer:
@nosovk, what browser did you use, btw? I think this list is only used by Firefox and a couple of other browsers, so it still should be possible to set cookies with Chrome, say, for domains in this list?

sleevi:
@gauzy-gossamer This list is used by every major browser (Edge, Safari, Chrome, Firefox, Opera) now. They may update at a different pace, but it is consistently used.

ForNeVeR:
FYI, the PSL is recommended (although not required) for use by [RFC 6265](https://tools.ietf.org/html/rfc6265#section-5.3).

gauzy-gossamer:
@sleevi, thanks for you response. Could you comment on whether this PR will be merged or not? I'm not sure what the status of this request is at this point. These domains will be used exclusively for public registrations for any foreseeable future, so I don't see any possible conflicting issues with the purpose of this list.

nosovk:
@gauzy-gossamer I was checking http://telegram.org.ru/ in Chrome Dev 

> Google Chrome	61.0.3128.2 (Official Build) dev (64-bit) (cohort: Dev (Clang))
> Revision	f4b37c652763afaf0333dbeb7bff6f06ceca0fc7-refs/branch-heads/3128@{#2}
> OS	Windows

gauzy-gossamer:
@sleevi, @weppos , Hi, could we get any feedback on this issue?

PR #413, opened 2017-03-10
Add Advisor Websites domains

jkenn99:
We provide application hosting services to customers using subdomains like `[customername].[shared server].awdev.ca` and `[customername].[shared server].advisor.ws`
For example, `james0208.force1.awdev.ca` and `james.us1.advisor.ws`


jkenn99:
TXT record added for both domains:
`_psl	1800	IN	TXT	"https://github.com/publicsuffix/list/pull/413"`

PR #414, opened 2017-03-12
Add Fedora People and Cloud domains

puiterwijk:
These domains are used for contributor web pages (e.g. https://puiterwijk.fedorapeople.org/) and services that contributors run in our OpenStack cluster (e.g. https://copr.fedorainfracloud.org/).


puiterwijk:
DNS entries for validation are in place.

PR #419, opened 2017-03-19
Added Thingdusts per-user subdomains

adrianimboden:
We create a independent webfrontend for every customer at our company. They all have a different Subdomain (in different stages of developement: dev, prod, testing or disrec)

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -11809,6 +11809,13 @@ gdynia.pl
 med.pl
 sopot.pl
 
+// Thingdust AG : https://thingdust.com/
+// Submitted by Adrian Imboden <adi@thingdust.com>
+cust.dev.thingdust.io
+cust.disrec.thingdust.io
+cust.prod.thingdust.io
+cust.testing.thingdust.io
+
 // TownNews.com : http://www.townnews.com
 // Submitted by Dustin Ward <dward@townnews.com>
 bloxcms.com


adrianimboden:
How is it going with my request? If you have further questions, I'm happy to answer them.

sleevi:
Could you provide example domains to ensure this is well-formed?

Also, can you clarify per https://github.com/publicsuffix/list/wiki/Guidelines the purpose of this request? From your description, it sounds like it's for internal purposes only.

adrianimboden:
We host our internal services (example see below) and all customer instances under the same top level domain (thingdust.io).

All domains host services which are independent of each other.
The entry in the public-suffix list would ensure, that customer domains cannot set cookies on internally used services or other customers.

The address space is used as such:
{service}.{stage}.thingdust.io, and
{customer}.cust.{stage}.thingdust.io

Examples are:
jenkins.prod.thingdust.io
monitoring.prod.thingdust.io
demo.cust.prod.thingdust.io
hq.cust.prod.thingdust.io

jenkins.testing.thingdust.io
monitoring.testing.thingdust.io
demo.cust.testing.thingdust.io
hq.cust.testing.thingdust.io

sleevi:
To confirm: You only wish to ensure that {cust}.cust is segmented, and you're not worried about {service}.{stage} - correct?

adrianimboden:
The {service} part is not really a problem, yes. But wouldn't that be reasonable too? The services are independent of each other and should not be able to fiddle with the other domains.

But yes, my primary interest is the cust.{stage} part.

PR #422, opened 2017-03-22
Private listing for Fullrate ISP.

luttermann:
Fullrate is an ISP, these domains are used for our public website (https://www.fullrate.dk/) as well as customer A, AAAA and PTR records.


sleevi:
Can you explain more about what you mean? Can you provide example domains?

Based on your description, it doesn't sound like you want `fullrate.dk` in the list, but perhaps I'm misunderstanding.

sleevi:
Closing due to no response to questions.

PR #432, opened 2017-04-07
Add uber.space

wernermo:
We're a hosting company and we plan to use the domain `uber.space` for our customers. Every account will get its own `mydomain.uber.space` domain. For cookie isolation it would be great to get included in the public suffix list.


wernermo:
```
➜ dig +short TXT _psl.uber.space
"https://github.com/publicsuffix/list/pull/432"
```

PR #433, opened 2017-04-10
Add bplaced network domains

bplaced:
hello at PSL!

as a shared hosting provider we have interest in a PSL listing for multiple reasons.

1. customers have their own subdomains (possibly for one or all of the given domains) assigned upon registration. Preventing privacy-damaging "supercookies" is a main goal here. We have more than 200k of such subdomains at the moment.
2. also we plan to use let's encrypt here, having these domains in the PSL allows great advantages for the signing process.
3. document.domain property restrictions are considered an advantage as well.

you may contact me via "security @bplaced dot n-e-t" as given by the guidelines, also the DNS records were set up.
Thanks a bunch!

kind regards,
Miroslav Bozic
bplaced CEO
https://www.bplaced.net/

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -395,6 +395,15 @@ net.bo
 mil.bo
 tv.bo
 
+// bplaced : https://www.bplaced.net/
+// Submitted by Miroslav Bozic <security@bplaced.net>
+square7.ch
+bplaced.com
+bplaced.de
+square7.de
+bplaced.net
+square7.net
+
 // br : http://registro.br/dominio/categoria.html
 // Submitted by registry <fneves@registro.br>
 br
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -395,15 +395,6 @@ net.bo
 mil.bo
 tv.bo
 
-// bplaced : https://www.bplaced.net/
-// Submitted by Miroslav Bozic <security@bplaced.net>
-square7.ch
-bplaced.com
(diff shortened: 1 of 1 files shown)


sleevi:
Note that you've placed them in the wrong section, per the [Guidelines](https://github.com/publicsuffix/list/wiki/Guidelines)

Could you both update that and the company identifying information (consistent with other entries in that section)?

Thanks!

bplaced:
hi sleevi

sure, got the wrong group, thanks for the hint!

sleevi:
Are you sure you want to add bplaced.net to the list?

That can cause all sorts of issues for your domains (e.g. `www.bplaced.net`, `forum.bplaced.net`).

In general, it's strongly recommended to use distinct domains for your subscribers (e.g. `github.com` vs `github.io`). Once done and picked up by browsers, it can take months (or more!) to remove, and thus may disrupt access to your services. I just want to make sure you're aware that the current request will likely cause issues for you.

bplaced:
hello sleevi

to my understanding possible issues you are referring to is tracking/sharing cookies between www. or forum., which is not of interest at all. Each current and future site of ours is designed to work and communicate via other secure channels, independently of the subdomain in use.
Other issues such as DMARC lookups for subdomains are not of interest as well.

Non-reserved bplaced.net subdomains are assigned to customers along with the others in the request, we have no plans to change that unless there is a particularly good reason to do so.

PR #435, opened 2017-04-17
Add .jur.pro.br in Pivate Section

RegistryPRO:
Include the suffix .jur.pro.br in the private section of PublicSuffix.org


weppos:
@RegistryPRO this patch is incorrect, it includes a commit that has nothing to do with the TLD you reference.
See https://github.com/publicsuffix/list/pull/435/files

PR #440, opened 2017-04-20
Add redpanda.zone Domain

webeindustry:
Red Panda Solutions is LLC registered in Texas SOS# 0802671198 TAXID# 32063134657. We provide SaaS with subdomains for redpanda.zone. LLC has two members Brandon Chapman and Connor Davis. 

 `dig TXT _pls.redpanda.zone` provides `_pls.redpanda.zone.     59      IN      TXT     "https://github.com/publicsuffix/list/pull/440"`

My first request was not keeping to alphabetical order. Second change corrected this.

foo.redpanda.zone & bar.redpanda.zone should be incapable of setting cookies towards each other.

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -11773,6 +11773,10 @@ rackmaze.net
 // Submitted by Tim Kramer <tkramer@rhcloud.com>
 rhcloud.com
 
+// Red Panda Solutions LLC : http://pandawins.com/
+// Submitted by Brandon Chapman <admin@pandawins.com>
+redpanda.zone
+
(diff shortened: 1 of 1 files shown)


sleevi:
Unable to access the webpage to cross-check, but the PSL record checks out.

Because the PSL is a finite resource, due to binary size, we try to make sure that the domains added are in the broader public interest. Can you ping this when the server is up?

sleevi:
Note: Still missing details from https://github.com/publicsuffix/list/wiki/Guidelines#submit-the-change

PR #441, opened 2017-04-24
Add serveo.net

trevordixon:
Customer domains will be hosted as subdomains of serveo.net. For example, foo.serveo.net or wow.serveo.net may be provisioned to customers and should have no way to share cookies.

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -11979,4 +11979,8 @@ cc.ua
 inf.ua
 ltd.ua
 
+// Serveo : https://www.serveo.net/
+// Submitted by Serveo admin <t@serveo.net>
+serveo.net
+
 // ===END PRIVATE DOMAINS===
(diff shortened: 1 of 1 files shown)


trevordixon:
```
$ dig +short TXT _psl.serveo.net
"https://github.com/publicsuffix/list/pull/441"
```

sleevi:
You probably don't want customer domains / your PSL entry to also match your corporate entity - that can cause issues with things such as navigating to your page or, for example, integrating single sign on or other forms of privilege separation (e.g. login.serveo.net)

If you're certain you understand the risks of adding your corporate domain to the list, we can do so, but realize if it causes issues, it can take 18 weeks or more before changes percolate to clients.

sleevi:
Will close this issue out within a week if no response.

trevordixon:
Holding off for now. Thanks for your help.

PR #444, opened 2017-04-29
Add UnoEuro domain

emilstahl:
UnoEuro Danmark A/S is a shared hosting provider.

This domain is used for temporary URLs for our web hosting customers.

You can find examples of this usage on Google:
https://www.google.com/#q=site:unoeuro-server.com

DNS record for authentication is in place.
```
~ esp$ dig +short _psl.unoeuro-server.com TXT
"https://github.com/publicsuffix/list/pull/444"
```

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -11934,6 +11934,10 @@ inc.hk
 // Submitted by Ed Moore <Ed.Moore@lib.de.us>
 lib.de.us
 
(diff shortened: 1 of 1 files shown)


sleevi:
@emilstahl Like #443 , it doesn't look like this will do what you want. Is that correct?

emilstahl:
@sleevi Correct. I found out, that I can't do what I wanted.

PR #446, opened 2017-05-03
RS.BA domain added

ambraspace:
Added rs.ba domain to the list.

BA domain is for Bosnia and Herzegovina. RS.BA is for Republic of Srpska - one of three constitutional entities of Bosnia and Herzegovina.

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -282,6 +282,9 @@ mil.ba
 net.ba
 org.ba
 
+// rs.ba : http://urc.rs/index.php/domen
+rs.ba
+
 // bb : https://en.wikipedia.org/wiki/.bb
 bb
 biz.bb


sleevi:
I'm sorry, but this does not appear to be a proper request.

Either the domain operator needs to request inclusion, per the [Guidelines](https://github.com/publicsuffix/list/wiki/Guidelines), or supporting documentation from the designated ccTLD operator (`nic.ba`) needs to be provided.

Further, because this is a second level domain, it should not appear in the `.ba` section if it's being privately operated by `URC`

ambraspace:
I understand the guidelines, and following them, I tried to contact both `nic.ba` and `URC`. `NIC.ba` response was that anyone can add domain suffixes to the list, and URC didn't even know how to respond to my request, so this was my best shot. :disappointed:

PR #450, opened 2017-05-09
Add onesignal.com to public_suffix_list.dat

jasonpang:
OneSignal is a push notifications service, and we provide subdomains for customers sending web push notifications. 

Web push notifications in future Chrome versions (59+) on Mac OS X will be displayed in Mac's native notification UI:

image

instead of Chrome's custom notification UI, which has been used until now:

image

[One change in beta Chrome versions is _only displaying the TLD + 1_ for the sending site](https://bugs.chromium.org/p/chromium/issues/detail?id=717725#c8), such that a notification sent from _subdomain.onesignal.com_ will only show up as _onesignal.com_:

image

^ The subdomain _wordpressdemo_ is missing in the above notification

For these notifications, it would be confusing for end-users to see notifications coming only from _onesignal.com_, instead of _subdomain.onesignal.com_. By adding our domain to the Public Suffix List, Chrome can display _subdomain.onesignal.com_ on each web push notification.

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -11675,6 +11675,10 @@ cya.gg
 // Submitted by Eddie Jones <eddie@onefoldmedia.com>
 nid.io
 
+// OneSignal : https://onesignal.com/
+// Submitted by Jason Pang <jason@onesignal.com>
+onesignal.com
+
 // OpenCraft GmbH : http://opencraft.com/
 // Submitted by Sven Marnach <sven@opencraft.com>
 opencraft.hosting


sleevi:
This doesn't sound like a valid reason to request inclusion, much like requesting to have Let's Encrypt limits raised is not a valid reason (in and of itself) to request inclusion.

If anything, it sounds like you're attempting to get around browser security UI, rather than improve browser security, but perhaps I've misunderstood.

Among other things, adding `onesignal.com` to the list can end up preventing you from being able to set cookies on (or potentially navigate to) `onesignal.com`, which from the looks of https://onesignal.com, would explicitly break your product.

jasonpang:
Hi there,

Thank you very much for your feedback! We definitely want to improve the browser UI to make it more clear to the end-user which sites they were receiving notifications from, but we're reconsidering based on the issue of not being able to set cookies on the site, as well as the potential for not being able to navigate to onesignal.com.

PR #454, opened 2017-05-14
Add lima-city/TrafficPlex domains

phillipp:
Hi,

we're a web hosting provider and cloud company and have multiple domains where customers can freely register subdomains. We have over 400k subdomains currently active on the listed domains.

We'd like to be included in the PSL for security concerns (mainly cookies).

DNS-Records will be set up soon.

Best,
Phillipp

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -11893,6 +11893,28 @@ sopot.pl
 bloxcms.com
 townnews-staging.com
 
+// TrafficPlex GmbH : https://www.trafficplex.de/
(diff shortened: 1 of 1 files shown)


phillipp:
TXT-DNS-Records are now set up for all domains.

I also wanted to add that we do recognize that it is recommended to separate the domain where we host our site and our customers sites, but for historically reasons this is not possible (we've been operating the domain in "mixed mode" for 14 years now).

We're aware of the consequences of a PSL listing and that it takes a long time to undo the listing.

PR #456, opened 2017-05-19
Add pixolino.com domain

intermetrics-admin:
[pixolino.com](https://pixolino.com) is a service where customers can build their own websites under a *.pixolino.com subdomain. We need to ensure the appropriate and secure handling of cookies.

Examples:
- [developers.pixolino.com](https://developers.pixolino.com)
- [demo.pixolino.com](https://demo.pixolino.com)

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -11507,6 +11507,10 @@ se.leg.br
 sp.leg.br
 to.leg.br
 
+// intermetrics GmbH : https://pixolino.com/
+// Submitted by Wolfgang Schwarz <admin@intermetrics.de>
+pixolino.com
+
 // IPiFony Systems, Inc. : https://www.ipifony.com/
 // Submitted by Matthew Hardeman <mhardeman@ipifony.com>
 ipifony.net


sleevi:
Note: Adding the domain of your main site can prevent navigation or access to it in clients. Are you sure you wish to do this? Once done, it's not easily removed.

intermetrics-admin:
As we understand it, listing "pixolino.com" in the Public Suffix List will not compromise our ability to set cookies for this specific domain. We are aware that we will not be able to share cookies between this domain and subdomains. We are not aware of anything that would prevent clients from accessing or navigating to pixolino.com. If this is correct, then yes, please add pixolino.com to the Public Suffix List.

sleevi:
No, PSL consumers may not allow setting on suffices (much like one would not allow setting a cookie for 'com'), and some prevent navigation (much like one would not allow navigation to 'com')

sleevi:
Can you confirm you still wish to be added?

intermetrics-admin:
Thank you for you advice. We discussed it in detail and will change the configuration of our main domain to use a subdomain. So, yes. please add our entry.

groundcat:
hi @intermetrics-admin **do you still need the domain pixolino.com to be included in the PSL?**
It appears that the project or organization associated with your domain, which was previously added to the Public Suffix List (PSL), may no longer be active based on current domain usage and availability checks. If the service has been discontinued and the domain is no longer associated with an active project, please confirm whether the entry can be safely removed from the PSL. _This helps keep the list accurate and avoids retaining unused entries when projects reach end of life. The PSL is a widely used resource relied upon by many applications and users worldwide, so entries should remain limited to projects that continue to require inclusion._ If your project is still active and requires continued PSL coverage, **please reply with confirmation** and ensure that the required verification records are maintained according to the [PSL guidelines](https://github.com/publicsuffix/list/wiki/Guidelines), and that the domain remains properly renewed. If no response is received after a reasonable period, the entry may be considered for removal. Thank you.

PR #460, opened 2017-05-29
Add info.cx

igloo22225:
info.cx offers free sub domains through afraid.org. I would like to add the domain to the suffix list to ensure cookies are handled properly.

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -11477,6 +11477,10 @@ iki.fi
 biz.at
 info.at
 
+// info.cx http://info.cx
+// Submitted by Jacob Slater <whois@igloo.to>
+info.cx
+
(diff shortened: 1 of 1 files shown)


igloo22225:
I have added a TXT record for _psl.info.cx. If you wish to validate by e-mail as well, I'll get back to any inquiries sent to the official whois e-mail swiftly.

igloo22225:
Example domains:
illuminati.info.cx
iconoclasm.info.cx
igloo.info.cx

sleevi:
Note: Adding info.cx to the PSL can prevent navigation to http://info.cx . Are you sure that's what you do?

igloo22225:
Yes - that is not an issue. The page on the root domain has only been maintained at this point for liability purposes.

PR #462, opened 2017-06-01
Add temp-dns.com

danmilleruk:
Subdomain variants of this domain name are used across our network to provide legitimate temporary staging environments to our customers.

Sub 6 Limited is the parent company that owns this domain name.

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -11862,7 +11862,7 @@ storj.farm
 
 // Sub 6 Limited: http://www.sub6.com
 // Submitted by Dan Miller <dm@sub6.com>
-*.temp-dns.com
+temp-dns.com
 
 // Synology, Inc. : https://www.synology.com/
 // Submitted by Rony Weng <ronyweng@synology.com>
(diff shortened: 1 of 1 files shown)


sleevi:
Could you provide example subdomains to ensure that this is correct. For example, it's unclear that the wildcard is necessary.

danmilleruk:
So for example, we would provide the following to our customers:

domainname-randomstring.temp-dns.com

to our customers. This is made up of a derivative of their domain name lesser dots and a random string.

The primary purpose is so our customers can use this as a staging environment prior to purchasing a domain name.

Let me know if you need any further clarification on the above.

sleevi:
Based on that description, it would appear that `temp-dns.com` would be sufficient, which would ensure each `domainname-randomstring` is distinct. As currently written, the proposed change would make `www.domainname-randomstring.temp-dns.com` and `accounts.domainname-randomstring.temp-dns.com` distinct (and make it difficult/impossible to navigate or set cookies for `domainname-randomstring.temp-dns.com`)

sleevi:
The TXT record for `_psl.temp-dns.com` still appears to be missing

danmilleruk:
Thanks for clarifying, @sleevi.

I have corrected the entry to the PSL DB and updated the PR to reflect this.

sleevi:
Apologies, that was my error. Looks like it's good here.

PR #466, opened 2017-06-13
add CloudAccess.net free domains to Private

Pawel-Panek:
to clean up https://github.com/publicsuffix/list/pull/372

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -10714,11 +10714,11 @@ c66.me
 
 // CloudAccess.net : https://www.cloudaccess.net/
 // Submitted by Pawel Panek <noc@cloudaccess.net>
-cloudaccess.net
-cloudaccess.host
-freesite.host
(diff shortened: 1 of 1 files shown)


sleevi:
I have re-sorted the list to comply with the Guidelines.

However, to ensure that these entries are correct, we still would need to see example domains. In https://github.com/publicsuffix/list/pull/372#issuecomment-307182559 it was suggested this cannot be provided, but at the same time, suggested it's possible to register example domains to show how they'd be used.

Can you please provide example domains so that we can confirm the syntax is correct and aligns with the desired results?

Pawel-Panek:
Here are example domains:

http://expltest.cloudaccess.net/
http://wsiwyg.wpdevcloud.com/
http://batan.jdevcloud.com/
http://basecloud.freesite.host/
http://sellatek.cloudaccess.host/

PR #470, opened 2017-06-18
add *.stage.nodeart.io

nosovk:
- Updated pull request from https://github.com/publicsuffix/list/pull/410 to current master of PSL
- changed pull request to avoid problems with adding whole domain to PSL (internal services use SSO, which will broke if we add root domain to PSL)
- Current request comply PSL purposes.

Under stage subdomain different clients can publish their apps, that allows 3rd party developers to steal cookies from another sub domain. Which is problem for HIPPA compatible app etc.

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -11740,6 +11740,11 @@ sytes.net
 webhop.me
 zapto.org
 
+// NodeArt : https://nodeart.io
+// Submitted by Konstantin Nosov <Nosov@nodeart.io>
+*.stage.nodeart.io
+
+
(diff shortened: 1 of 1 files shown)


sleevi:
Based on the PR being about bypassing Let's Encrypt rate limits, closing this as WontFix.

nosovk:
@sleevi I mentioned above that it also should be added because of cookie safety in browser. It easily could be used to steal auth cookies frome other app. It's basic purpose of that list - protect cookies.
Will you merge that one or I shall create one more PR with explanation how cross-domain cookies can be used in exploit?

PR #474, opened 2017-06-20
Add blor.me, blor.tech, dvps.io

eodabas:
Blue Orange Global is a UK based software and marketing company creates custom hosted projects for its customers.

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -10657,6 +10657,12 @@ betainabox.com
 // Submitted by Nathan O'Sullivan <nathan@mammoth.com.au>
 bnr.la
 
+// Blue Orange Global : http://blueorangeglobal.com
+// Submitted by Emre Odabas <eo@blueorangeglobal.com>
+dvps.io
+blor.me
+blor.tech
+
 // Boxfuse : https://boxfuse.com
 // Submitted by Axel Fontaine <axel@boxfuse.com>
 boxfuse.io


sleevi:
Please provide the additional details per https://github.com/publicsuffix/list/wiki/Guidelines#submit-the-change

eodabas:
We're creating custom "on-premises" hosted applications and providing alternative domain names (blor.me, blor.tech and dvps.io) for our customers to be used with the service. The DNS of the these provided domains are managed by our customers. It's like registering a domain name but with an application provided with it. As it's an on-premises solution for our customers, we're losing control of the environment.

sleevi:
Can you please provide example names?

Attempting to find information about these domains only reveals a wildcard *.dvps.io from Comodo, which seems inconsistent with the justification, but perhaps it was misunderstood.

Providing examples of customer domains will help ensure this request is properly formatted.

eodabas:
We've already applied for rate limit increase on let's encrypt, and for customer controlled domains we're going to provide certificates in a secure and automated way. So you may reject the pr. thanks for your valuable time.

But per you question, here is the main idea.
let's say we've provide "customer.blor.tech" for the customer.
 * customer.blor.tech dns is managed by our customer. (*.customer.blor.tech, including customer.blor.tech)
 * customer may set their servers' ip addresses for sandbox.customer.blor.tech, qa.customer.blor.tech, content.customer.blor.tech. (for branding, the application is allowed to be run under *.customer.blor.tech only)

Additionally, the customer may use custom domain name and/or we may host the application entirely, For which we've already applied for rate limit increase on let's encrypt side as i've mentioned above.

PR #451, opened 2017-05-09
Add drayddns.com to public_suffix_list.dat

drayddns:
Greeting Sir.
Our company is going to provide service to let customers register the domain name and get the certificate from Let's Encrypt.
Because the Let's Encrypt is using the Public Suffix List to calculate the domain and limit the certificates.
By adding the "drayddns.com" into Public Suffix List, the Let's Encrypt and browsers could identify our domain name correctly.

The examples of our domain name list below:
~~https://example1.drayddns.com~~
~~https://example2.drayddns.com~~
Update the list in next comment.

Thank you, sir.

PR #475, opened 2017-06-21
Added "drayddns.com" into public suffix list.

drayddns:
Greeting Sir,

First, thank you for your notification in #451 and apologize for my misunderstanding.

We're DrayTek Corp. And this is our website: https://www.draytek.com/ 

Our company is building DDNS service to provide customers register their domain, right now we are planning using *.drayddns.com for DDNS service.
By adding the "drayddns.com" into Public Suffix List, the Let's Encrypt could identify domain correctly and each customer could get their own certificate.

examples of our domain list below:
https://vkao-home.drayddns.com
https://smallkai.drayddns.com

If there anything we need to do or we can do, please inform us. 
Thank you for taking your time and have a nice day. :D

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -10836,6 +10836,10 @@ dedyn.io
 // Submitted by Norbert Auler <mail@dnshome.de>
 dnshome.de
 
+// DrayTek Corp. : https://www.draytek.com/
+// Submitted by Paul Fang <mis@draytek.com>
+drayddns.com
+
 // DreamHost : http://www.dreamhost.com/
 // Submitted by Andrew Farmer <andrew.farmer@dreamhost.com>
 dreamhosters.com


drayddns:
The TXT record now still indicate to the old one, #451. After it update, I will comment it here soon. 

Thank you. :D

sleevi:
There is no evidence in the request to suggest the purpose is anything other than to bypass Let's Encrypt rate limits, which is not a valid reason per https://github.com/publicsuffix/list/wiki/Guidelines#validation

drayddns:
Hi @sleevi, thank you for your reply.
I believed there’s a misunderstanding as this post was created by my assistant. The Draytek is a secure network equipment provider wish its service to be as secure and sound as possible. The main reason we apply for public suffix list is to provide a secured and correctly recognizable individual domain or hostname in our service suffix domain “drayddns.com” for each customer to register or purchase.

In such way, web services hosted in individual registered host such as “customer1.drayddns.com” or domain “*.customer1.drayddns.com” would not be recognized as same domain as “customer2.drayddns.com”. Therefore, cross-domain cookie would not be created, search list will not consider these two the same, zone determination or even possibly Java/ActiveX should correctly applies security setting. The DDNS service we provided should be safe from cross-domain accessing by public contributed suffix list.

If any further question, please don’t hesitate to let me know. Thank you and have a nice day.

PR #476, opened 2017-06-21
Add boomla.net domain

zupa-hu:
Boomla ([boomla.com](https://boomla.com/)) allows anyone to create and host websites on `*.boomla.net` subdomains. We want to prevent users from adding cookies to `boomla.net`.

Example customer domain: [psl.boomla.net](http://psl.boomla.net/)


zupa-hu:
Added TXT record.

zupa-hu:
Thanks @sleevi !

PR #484, opened 2017-07-11
Update AWS Elastic Beanstalk domain suffixes.

kunsha:
Updating existing entry and adding all region specific ones.
The wildcard entry is incorrect. We give our customers CNAMEs with these TLDs.

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -10582,8 +10582,23 @@ us-east-1.amazonaws.com
 
 // Amazon Elastic Beanstalk : https://aws.amazon.com/elasticbeanstalk/
 // Submitted by Luke Wells <psl-maintainers@amazon.com>
-elasticbeanstalk.cn-north-1.amazonaws.com.cn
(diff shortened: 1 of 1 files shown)


sleevi:
Can you explain why you're proposing this? It was intentional to use the wildcard syntax.

kunsha:
I updated my first comment to reflect the intention of change.

pzb:
The current wildcard entry is not correct.  Many existing <label>.elasticbeanstalk.com names are hostnames assigned to customers.  While new names are <label>.<region-name>.elasticbeanstalk.com, the old names need to work.  Therefore *.elasticbeanstalk.com does not work correctly.

daftano:
@pzb @kunsha could you give some examples of new names (..elasticbeanstalk.com)?
how many existing .elasticbeanstalk.com names you have?

sleevi:
Confirmed out of band the authorization and correctness of this, unfortunately (due to removing the wildcard).

PR #487, opened 2017-07-13
Add new .br city domains listed at https://registro.br/dominio/categoria.html

rainest:
It looks like some of these are not fully available yet but will be in the near future--the latest is natal.br which will become available on 2017-07-31.

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -399,9 +399,11 @@ tv.bo
 // br : http://registro.br/dominio/categoria.html
 // Submitted by registry <fneves@registro.br>
 br
+abc.br
 adm.br
 adv.br
 agr.br
+aju.br
 am.br
 arq.br
 art.br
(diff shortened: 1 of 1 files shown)


rubenskuhl:
@rainest is 99.9% correct. These city domains are already available at this point:
abc.br
bsb.br
macapa.br
osasco.br

The announced release dates for the others are:
maceio.br (July 17)
aju.br (July 20)
cuiaba.br (July 24)
udi.br (July 27)
natal.br (July 31)
riobranco.br (Aug 3)
rio.br (Aug 7)

I do believe is indeed more effective to add all of the above in a single update instead of waiting until their launch dates, since they are already defined and reserved. 

Registro.br will keep launching 2 cities a week until Sep or Oct, trying to bundle the announcements.

daftano:
pull request #507 adds more domains (but doesn't include rio.br).

@rainest @rubenskuhl does it make sense to you to close this one in favour of the aforementioned one?

rubenskuhl:
.rio.br has been postponed, that's why adding it at this point is
premature. The other ones are either already active or on road to be
activated.


Rubens

rainest:
@rubenskuhl @daftano sounds good to me. The submitter of #507 is more familiar with the registry than I am; I trust they'll add .rio.br when it's no longer postponed.

daftano:
Thank you @rainest.

PR #490, opened 2017-07-18
Add byen.site

yaslama:
P.P.COM (ppcom.fr via byen.site) offers our customers a website hosted in a domain *.byen.site
E.g.:  dupond-pierre-taxi.byen.site

Reason for inclusion
Cookie isolation. Each customer can partially modify himself his site so cookie isolation is needed for security reason.

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -11953,6 +11953,7 @@ protonet.io
 // Publication Presse Communication SARL : https://ppcom.fr
 // Submitted by Yaacov Akiba Slama <admin@chirurgiens-dentistes-en-france.fr>
 chirurgiens-dentistes-en-france.fr
+*.byen.site
 
 // QA2
 // Submitted by Daniel Dent (https://www.danieldent.com/)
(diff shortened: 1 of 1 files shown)


yaslama:
$ dig +short TXT _psl.byen.site
"https://github.com/publicsuffix/list/pull/490"

tmm1:
Based on your example domain, the entry should not contain *. at the beginning.

The list should also be sorted alphabetically.

yaslama:
Hi @tmm1,
1. I don't understand why do you think that the entry should not contain *.
2. In https://github.com/publicsuffix/list/wiki/Guidelines, I can read "Sort first by the TLD, then the first label to the left of the TLD, and so forth". So I think the order is right.

tmm1:
If your example subdomain is `dupond-pierre-taxi.byen.site`, then the suffix should be `byen.site`. By adding `*.byen.site` to the suffix list, you are stating that no cookies can be set on `dupond-pierre-taxi.byen.site` or any other subdomain.

Please review the documentation on the list format.

tmm1:
The order needs to be fixed now to be alphabetical.

yaslama:
It's already sorted: site > fr

tmm1:
Ah ok, you're right. 👍

weppos:
Thanks @tmm1 for the support.

tmm1:
Sure, reviewing all these PRs seems like a tedious and thankless job, so I'm happy to do my small part to help support the great work you all do.

PR #492, opened 2017-07-19
Add keybase.pub

golyalpha:


tmm1:
Your entry in the file must be inserted alphabetically. You also need to provide the reason you wish to add this domain to the public suffix list, and verify your ownership of this domain by adding the required TXT records.

golyalpha:
I do not own the domain, but it is a public suffix. If you wish, I can try
to send an email to Chris from Keybase. And the reason to add this one
would be that it's a public suffix.

sleevi:
Closing for not following the Guidelines. Only the domain owner may request inclusion.

golyalpha:
I'm sorry in that case, never would have thought that to prove a domain was
a public suffix, it was a requirement to own said domain. I am sorry for my
ignorance.

PR #495, opened 2017-07-24
Add channelsdvr.net to the PSL

tmm1:
We are setting up a dynamic dns service for users of our server software, which is installed by the user on a computer in their home. For example I might register a domain like tmm1.remote.getchannels.com, for easy remote access to the server I setup at home. 

Since we use the getchannels.com domain for marketing and subscription purposes, the intent of this PR is to ensure that cookies are not shared between user sub-domains and our root domain, nor across different user domains (by using a `.remote.getchannels.com` supercookie).

My understanding is that a cookie set on `.getchannels.com` would _not_ be made available to `tmm1.remote.getchannels.com`, but I would appreciate if someone can confirm my understanding of how the PSL works.

We also plan to use Let's Encrypt to issue SSL certificates for these user domains.

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -11400,7 +11400,7 @@ vologda.su
 
 // Fancy Bits, LLC : http://getchannels.com
 // Submitted by Aman Gupta <aman@getchannels.com>
-remote.getchannels.com
+channelsdvr.net
 
 // Fastly Inc. : http://www.fastly.com/
 // Submitted by Fastly Security <security@fastly.com>
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -11398,6 +11398,10 @@ vladikavkaz.su
 vladimir.su
 vologda.su
 
+// Fancy Bits, LLC : http://getchannels.com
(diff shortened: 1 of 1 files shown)


tmm1:
```
$ dig +short TXT _psl.remote.getchannels.com
"https://github.com/publicsuffix/list/pull/495"
```

tmm1:
After further consideration, we've decided to use a separate domain entirely to prevent any confusion and unforeseen security issues in the future. I have updated this pull request accordingly.

```
$ dig +short TXT _psl.channelsdvr.net
"https://github.com/publicsuffix/list/pull/495"
```

I think it still makes sense for us to be on the PSL, to prevent cookie sharing between different users' domains. If not, then I am happy to apply for a Let's Encrypt rate limit exemption instead.

tmm1:
Our DDNS system on this new domain is up and running. We are assigning randomized domains to each user, for example 4990c9ef96fa.channelsdvr.net

We also plan to host a management portal for users at my.channelsdvr.net. But before we do so we need to ensure that the `my.` sub-domain will be secure against cookies set via other user-controlled subdomains.

tmm1:
cc @weppos does this look sane?

PR #497, opened 2017-07-30
ru : List of Reserved second-level Domain Names from CCTLD.RU

bekreyev:
Official list of reserved second-level domains in .RU, mostly public geographical names.


sleevi:
These were intentionally removed because they were retired by the registry. The page you link references the old (historic) list which has been discontinued by the registry.

PR #500, opened 2017-08-05
Add netlib.re, public domain name registrar

KaneRoot:
netlib.re is a public domain name registrar for non experienced users willing to have a name on the Internet.

Reason for inclusion
Cookie isolation. Each user can partially modify himself his site so cookie isolation is needed for security reason. Also, certificates are a problem, they all need to be able to own their own.

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -5697,6 +5697,7 @@ sch.qa
 re
 asso.re
 com.re
+netlib.re
 nom.re
 
 // ro : http://www.rotld.ro/
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -5697,6 +5697,8 @@ sch.qa
 re
 asso.re
 com.re
+// Association Alsace Réseau Neutre: https://www.arn-fai.net
+// Submitted by Philippe Pittoli <admins@arn-fai.net>
 netlib.re
 nom.re
 


weppos:
Are you the domain owner?

KaneRoot:
Yes.

weppos:
Then please reformat/update the request according to the guidelines
https://github.com/publicsuffix/list/wiki/Guidelines

weppos:
Also, your inclusion explanation is copied from other requests, we need a specific explanation for your case.

Please note that the inclusion is not the appropriate way to request higher Let's Encrypt limits. You should use the form included in this page

KaneRoot:
Hello again, the file has been updated. This time with the company infos + submitter email.
DNS respond correctly, if you want to test _psl.netlib.re.  and I don't know what to add to my first message.

Netlib.re is a registrar, it is designed to help people get a name on the Internet so they can host whatever they want without requiring to pay for it. The website and the DNS servers are hosted by the Alsace Réseau Neutre association (https://www.arn-fai.net) which is a French ISP. You can check the netlibre website if you want:  https://netlib.re

Since netlib.re provides a domain name to each user, people should have a cookie isolation in their browser to avoid security problems. It is not just to circumvent the let's encrypt limitation.

KaneRoot:
Hello, it's been 4 months now… can someone please respond? We've done everything asked.

sleevi:
Closed for non-action.

PR #504, opened 2017-08-08
Update 2nd level city domain for publicsuffixlist

kutpenco:
Please, update the public suffix list, to add the 2nd level city domain GRU.BR. @fnevesbr is a representative of the registry and he can confirm the PR from your authoritative email.

This domain was released on 2018-08-07 12:00 GMT -03:00 Brasil.

Best Regards, Lucas Penco.


weppos:
@fnevesbr please follow up if the change is correct.

PR #519, opened 2017-08-28
Nic.it

daftano:
Update TLDs from nic.it in compliance with version 7.1 of [Italian Regulation assignation](http://www.nic.it/sites/default/files/docs/Regulation_assignation_v7.1.pdf).
In particular:
- regions are listed in Appendix A
- provinces are listed in Appendix and B
- municipalities are listed in [Appendix C](http://www.nic.it/sites/default/files/docs/List%20of%20Italian%20municipalities.pdf)


weppos:
@daftano do you represent the nic.it?

weppos:
/cc @sleevi FYI give the limits and constraints, unless the nic.it explicitly requests the inclusion and provide valid motivations, I'm actually inclined to not approve the addition of 39k entries representing reserved municipalities.

In fact, I would probably purge the existing entries as well that were added ONLY based on official requirements (and not direct requests). Regions and Provinces are a smaller number instead.

This is a very tough decision and I also understand it can be controversial. I think that it's somehow also connected to @pbz request about including all .NAME and .US suffixes.

Part of me really wants to have those names. But I also understand the constraints and implications.

I wonder... perhaps we should have a third section for IANA (althought we call it ICANN so far for historical reason...) registry reserved suffixes?

daftano:
@weppos no, I don't.
Changes come from the registry website (as per provided URLs).

weppos:
Thanks @daftano. I currently stand on my decision to not approve the list at the time being, due to the space constraint (and not being a direct request from the registry).

I am actually even considering to purge the existing ones, at least municipalities. I'll keep the discussion open for a while to see 1) what @sleevi thinks about it 2) whether @pzb has some feedback and 3) whether the idea of a separate section (that could be stripped on pre-processing) makes sense (although in that case we will probably move to a separate ticket).

daftano:
I understand and share @weppos feelings about adding municipalities due to mentioned previous discussions. As I couldn't make a decision myself, I welcome a discussione about them and any decision that will come.

It is hard, for me, to quickly number how many 4th level domain involving municipalities are currently registered, but I found many more than I expected; that surprised me.

I split the changes in three different commits exactly to allow cherry-picking those considered worth to be included in this list.

pzb:
Similar to the suffixes for `.us` and `.name`, I think it probably makes sense to include those that are actually in use rather than the complete reserved list.  I took a quick look at the nic.it website but did not see a way to request the full zone file.  Is there a process to do so?

daftano:
@pzb none automatic that I am aware, but I already requested that information to the nic. I will update this thread when I'll get an answer.

daftano:
@weppos and @pzb: have you considered another file (instead of another section)?
The benefit would be that current workflows won't be affected, and those wishing such extra domains (e.g. `.names` and `.us`) should include that extra file.

daftano:
@weppos  and @pzb: I tried to get the list from the registry, but at the end they come back saying that it's their "policy to provide no lists of any type with the exception of official Registrar and exclusively for the domains of their competence".

Given that I reckon we should not include the municipalities. I'm gonna close this PR to open a new one including regions and provinces only.

PR #525, opened 2017-09-19
add schokokeks.net

hannob:
This is a domain we use to host customer subdomains.

Each customer has a username and can create [username].schokokeks.net and [something].[username].schokokeks.net. Thus for security reasons it makes sense to separate them cookie-wise.


hannob:
DNS record created:
```
host -t txt _psl.schokokeks.net
_psl.schokokeks.net descriptive text "https://github.com/publicsuffix/list/pull/525"

PR #527, opened 2017-09-25
Update .FR sectorial domains

t1st3:
fix outdated link for `.fr` sectorial domains

https://www.afnic.fr/fr/produits-et-services/le-fr/les-domaines-sectoriels-en-fr-11.html

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -921,7 +921,7 @@ nom.fr
 prd.fr
 presse.fr
 tm.fr
-// domaines sectoriels : http://www.afnic.fr/obtenir/chartes/nommage-fr/annexe-sectoriels
+// domaines sectoriels : https://www.afnic.fr/fr/produits-et-services/le-fr/les-domaines-sectoriels-en-fr-11.html
 aeroport.fr
 assedic.fr
 avocat.fr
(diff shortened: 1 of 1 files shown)


sleevi:
@gerv @weppos I updated this with a second commit, but based on the documentation I can find, `presse.fr` and `assedic.fr` are not expressly part of the registry's dedicated suffices, but merely part of the names for manual review. That's a rather large list (and dynamic), and so I wanted to get your thoughts as to whether we should leave them for historic reasons or remove them for consistency reasons.

a455bcd9:
I've just made a similar PR, I didn't see yours before, indeed those sector-based domains should be removed from the list as they were removed in 2009: https://github.com/publicsuffix/list/pull/530

Assedic does not exist anymore ( https://en.wikipedia.org/wiki/ASSEDIC ).

`presse` is just a keyword in the manual review list: https://www.afnic.fr/fr/ressources/documents-de-reference/chartes/termes-soumis-a-examen-prealable/

weppos:
I tried to review the PR, but for some reason I can't reach www.afnic.fr today. It also means I was not able to read the links at #530

@sleevi I am fine to remove them, especially if they are privately owned. I think they were originally added because in the list.

If they have to stay in the list, we should require the explicit inclusion (as we did for the .RU TLDs we removed few months ago).

That means I'm also positive to merge #530 (althought again, I was not able to validate the docs yet, hence I can't approve it as it should require an explicit review first).

a455bcd9:
@weppos Did you manage to reach afnic.fr finally?

PR #530, opened 2017-09-29
Removed sector-based .fr domains

a455bcd9:
Sector-based .fr domains are now simple domains owned by private users/companies. The change was operated in 2009:

- https://www.afnic.fr/en/products-and-services/the-fr-tld/sector-based-fr-domains-4.html (English)
- https://www.afnic.fr/fr/produits-et-services/le-fr/les-domaines-sectoriels-en-fr-11.html# (French)
- https://www.afnic.fr/en/about-afnic/news/operations-news/8172/showOperational/sector-based-domaines-migration-avocat-fr-2-1.html (French)
- https://www.afnic.fr/fr/l-afnic-en-bref/actualites/actualites-operationnelles/5493/showOperational/migration-des-sous-domaines-sectoriels-avoues-fr-1.html (French)
- https://www.afnic.fr/fr/l-afnic-en-bref/actualites/actualites-operationnelles/3333/showOperational/migration-des-sous-domaines-sectoriels.html (French)

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -922,23 +922,7 @@ prd.fr
 presse.fr
 tm.fr
 // domaines sectoriels : http://www.afnic.fr/obtenir/chartes/nommage-fr/annexe-sectoriels
-aeroport.fr
-assedic.fr
-avocat.fr
-avoues.fr
-cci.fr
-chambagri.fr
-chirurgiens-dentistes.fr
-experts-comptables.fr
-geometre-expert.fr
 gouv.fr
-greta.fr
-huissier-justice.fr
-medecin.fr
-notaires.fr
-pharmacien.fr
-port.fr
-veterinaire.fr
 
 // ga : https://en.wikipedia.org/wiki/.ga
 ga
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -915,14 +915,12 @@ fo
 // fr : http://www.afnic.fr/
 // domaines descriptifs : http://www.afnic.fr/obtenir/chartes/nommage-fr/annexe-descriptifs
 fr
-com.fr
 asso.fr
+com.fr
+gouv.fr
 nom.fr
 prd.fr
-presse.fr
 tm.fr
-// domaines sectoriels : http://www.afnic.fr/obtenir/chartes/nommage-fr/annexe-sectoriels
-gouv.fr
 
 // ga : https://en.wikipedia.org/wiki/.ga
 ga


sleevi:
@a455bcd9 Thanks for the very informative links! I'd like to confer with the other PSL maintainers - @dnsguru and @weppos - as how best to handle this. As best I can tell from the information provided, these domains were previously reserved directly by the registry themselves, but have now transitioned to third-party management - but are still registerable at the 3LD.

That is, directly removing from the PSL would be damaging to the stability and security of these domains for historic reasons, but at the same time, they're no longer managed by the NIC themselves, and should potentially move to the PRIVATE section.

Did I understand the translated documents correctly?

weppos:
In the past, we had a similar case #359 and we went with removing the suffixes altogether, forcing the owner to request again the reinclusion in #412

The discussion in #359 was actually triggered because of a specific removal request, which is probably the main difference compared to this ticket here.

For this reason, I would be fine to grandfather the inclusion, and move the suffixes to the private section if we have clear evidence that they are still offering 3LD.

However, I noticed that not all suffixes were migrated to Promopixel ([SmalRegistry](https://www.smallregistry.net/)). Some were assigned to other entities.

The listing in the PRIVATE section should reflect this, and we should move them grouped by the new owner entity.

Furthermore, suffixes not in the Afnic list (e.g. assedic.fr) should likely be removed, so that only the ones officially mentioned in the sector based document https://www.afnic.fr/en/products-and-services/the-fr-tld/sector-based-fr-domains-4.html will be granfathered.

Any objection?

a455bcd9:
Sorry for the late answer @weppos , OK for me 👍

dnsguru:
@a455bcd9 
Am trying to clear this one up.  As it has sat for quite some time in the queue.

Weppos was suggesting that you revise the patch to migrate the names into the private section under their respective operators, and remove any that should be removed from under the Afnic section.

Could you please revise the patch?

(cc @weppos, @sleevi )

dnsguru:
@weppos did I summarize this correctly?

dnsguru:
I will close this for now as it needs a revision due to the age and being in conflict with the more recent PSL dat file.  It should be submitted again with the appropriate _PSL TXT records accordingly to validate

dnsguru:
I found another PR with the request for remove an entry for a stub zone from the ICANN section of a ccTLD   https://github.com/publicsuffix/list/pull/530  for removal of sector-based domains in .FR.  Still, it quotes policy changes, directly at the registry, that can be validated, which is why that PR proceeded where this did not.

PR #531, opened 2017-10-01
pcloud.host

pzduniak:
[Nucleos Inc](https://nucleos.com) is a company behind [PortableCloud](https://portablecloud.net), a "local cloud solution" (currently something like a platform with an app store so that non-technical users can easily set up whatever software they want on our custom server devices), which is currently expanding its featureset to remote access, supporting SSL certificates, even in LANs and so on. Currently our software is deployed at various schools in Zimbabwe, India and evaluated in countries like Mauritius, rural Mexico and so on.

I can't _really_ give you proper examples due to the nature of our platform (no proper authentication yet and so on), but every device is given its own domain under the pcloud.host suffix (like piotr.pcloud.host, which is currently routed to a tunnelling server, which does not do anything useful _yet_). The domain is currently mostly used for local access to our devices, upon proper configuration to the network, you can access applications hosted on the device by going to something.[yourdevice].pcloud.host, which has its own HTTPS certificate generated by querying our tunnelling server with a CSR, which is then passed on to Let's Encrypt to generate a cert that contains all DNS names in the SAN field.

Other than that there's the obvious issue of cookies, but we believe that for now it should be enough to only add pcloud.host to the list.

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -11889,7 +11889,7 @@ nodum.co
 nodum.io
 
 // Nucleos Inc. : https://nucleos.com
-// Submitted by Piotr Zduniak <piotr@zduniak.net>
+// Submitted by Piotr Zduniak <piotr@nucleos.com>
 pcloud.host
 
 // NYC.mn : http://www.information.nyc.mn
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -11888,6 +11888,10 @@ stage.nodeart.io
 nodum.co
 nodum.io
 
+// Nucleos Inc. : https://nucleos.com
+// Submitted by Piotr Zduniak <piotr@zduniak.net>
+pcloud.host
+
 // NYC.mn : http://www.information.nyc.mn
 // Submitted by Matthew Brown <mattbrown@nyc.mn>
 nyc.mn


pzduniak:
``` pzduniak@Piotrs-Mac-Pro  ~/  dig TXT _psl.pcloud.host
;; Warning: Message parser reports malformed message packet.

; <<>> DiG 9.8.3-P1 <<>> TXT _psl.pcloud.host
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 45344
;; flags: qr rd ra; QUERY: 1, ANSWER: 1, AUTHORITY: 0, ADDITIONAL: 0

;; QUESTION SECTION:
;_psl.pcloud.host.		IN	TXT

;; ANSWER SECTION:
_psl.pcloud.host.	300	RESERVED0 TXT	"https://github.com/publicsuffix/list/pull/531"```

groundcat:
Do you still need this entry in the PSL? The domain you initially requested to add no longer serves its original purpose and now shows a "for sale" page.

The _psl record no longer exists. (Reminder: you need to keep the _psl records.)

If the domain has been retired and is no longer linked to an active project, it would be helpful if you could submit a PR to roll it back. This would prevent leaving debris for PSL volunteers and future registrants to deal with.

Could you please comment on PR #2052 to confirm if it can be removed?

PR #533, opened 2017-10-09
Add Fedora Openshift app domains

puiterwijk:
These domains are used by the production and staging OpenShift instances run by Fedora Infrastructure and run apps by various application owners.


puiterwijk:
DNS entries for validation are in place.

PR #540, opened 2017-10-20
Add fastvps-server.com for FASTVPS EESTI OU

lilalkor:
FASTVPS EESTI OU (https://fastvps.ru/) provides technical domains for it's customers.
fastvps-server.com - for the server itself, so customer can use domain like s056579f2.fastvps-server.com
fastpanel2.direct - for our hosting panel, that can be installed on the server

These 3-rd level domains are not related with each other and should not trust each other.
Every domain is used by different server and typically by different customer.

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -11461,6 +11461,11 @@ a.ssl.fastly.net
 b.ssl.fastly.net
 global.ssl.fastly.net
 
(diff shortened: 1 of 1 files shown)


lilalkor:
Sorry, realised that we have an issue with hosting panel domain right now. Closing for now.

PR #542, opened 2017-10-23
Remove nic.in

arjunnkn:
nic.in is considered as an eTLD which makes it diificult for us to monitor the domains under nic.in.

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -1217,7 +1217,6 @@ net.in
 org.in
 gen.in
 ind.in
-nic.in
 ac.in
 edu.in
 res.in


arjunnkn:
nic.in is considered as an eTLD which makes it diificult for us to monitor the domains under nic.in.

public_suffix_list.dat
```diff
@@ -1217,7 +1217,6 @@ net.in
 org.in
 gen.in
 ind.in
-nic.in
```
ayende: Don't you need to remove the comment about nic.in on line 1211 as well?

arjunnkn: The line 1211 was already commented out, that is why we did not change it. Also, the line says that nic.in is not an official eTLD .

weppos:
No follow up.

arjunnkn:
Hi,

Due to Nic.in and gov.in are ETld  I am unable to create new registration
with let's encrypt acme server as it replies that email domain is TLD

All subdomains of gov.in and nic.in have email IDs on apex domain @gov.in
and @nic.in

PR #543, opened 2017-10-24
Add Names.of.London Domains

Names-of-London:
Names.of.London sells domain names in the form of keyword natural language phrases


Names-of-London:
Sorry - not sure if I should leave this "open" or "closed"

public_suffix_list.dat
```diff
@@ -11769,6 +11769,47 @@ net.ru
 org.ru
 pp.ru
 
+// Names.of.London : https://names.of.london/
+// Submitted by James Stevens <regsitry@names.of.london> or <james@jrcs.net>
```
robinwhittleton: Misspelled first email address?

Names-of-London:
Thanks for getting back to me.

Sorry - yes (as you can see in this From:)

Do you need me to fix it?


James

public_suffix_list.dat
Names-of-London: Yes - sorry - registry@names.of.london

Names-of-London:
Email address typo corrrected

Names-of-London:
I am more than happy to provide examples and purpose, but I have diligently searched the guidelines document and was unable to find any reference to this, hence it was not provided. I am also therefore a little unclear on what is expected.

Names.of.London runs a second level registry service on the domains listed - selling sub-domains to trusted third parties for use in whatever way they see fit - a similar business model to CentralNIC with their "<cctld>.[COM|NET]" domains, which are also listed in here.

As most registrar do not support our subdomains, we also provide registrar style services to help users make use of their subdomains - this includes DNS hosting or web & email forwarding.

Only those users who use our DNS hosting, or who use external DNS servers, can be truly considered as an external subdomain and therefore require listing here. Where all the client subdomains in a domain are exclusively used for web & email forwarding only, it would not be necessary to list it here.

Hence, although in theory, any of our domains might be listed here, in practice most do not need to be. In addition, where domains are deemed "commercially unviable" (e.g. no customers have used that domain) we will drop it. This would mean more frequent changes to our entry here.

As the time it takes to get changes implemented in this list is quite considerably more than we had expected, we have tried to plan ahead and consider which domains need to be listed and which might end up getting dropped, over the next 12 to 18 months.

Hence, the ones left are the ones we consider have a longer term viability.

Examples... (as of now)

delegated subdomains, owned by an untrusted third party, using external nameservers - realestate.for.sale , privatejets.for.sale

subdomains, owned by an untrusted third party, using our internal registrar services - land.for.sale , bitcoin.for.sale , orgy.for.one

I hope this helps. If you have any further questions, if you could be more specific it would really help us give more specific answers.

dnsguru:
@Names-of-London of.football appears to no longer be registered and we're clearing out NXD
Can you submit a PR to clear unregistered names of yours out please?

james-stevens:
> @Names-of-London of.football appears to no longer be registered and we're clearing out NXD
> Can you submit a PR to clear unregistered names of yours out please?

Sure no probs

PR #544, opened 2017-10-25
Adding own.pm

MontiMcMannus:
Hi,

I change type of company (GmbH now) and added own.pm, now at the correct place

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -11952,9 +11952,10 @@ operaunite.com
 // Submitted by Duarte Santos <domain-admin@outsystemscloud.com>
 outsystemscloud.com
 
-// OwnProvider : http://www.ownprovider.com
+// OwnProvider GmbH: http://www.ownprovider.com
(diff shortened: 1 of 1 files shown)


MontiMcMannus:
Hi weppos,

Explanation:
We are giving away third level domain for customers of our boxes, and we know that www.ownprovider.com is totally seperated from any other domain like login.ownprovider.com. But for a company which cares about privateness of data and does NOT follow the current "putting all into clouds" and "merging all web apps into a big monster without understanding security side effects or data leakage", this is ok for us.  Ownprovider.com is too long for some customers, so we provide also own.pm for customers. Each customer can request a subdomain like cust1.own.pm for free and we don't want to allow change cookies of cust1 for cust2, these 3rd level subdomains should be as seperated as possible. Ownprovider.com should resolve just for safety or short links. It will be redirected to www.ownprovider.com, so cookies are not necessary, and we don't find anything which doesn't allow the top level to resolve. 

Validation records:
I missed it too, but "Pull requests are automatically checked by our test suite," --> can I enforce it?

Best

PR #545, opened 2017-10-26
.br new 2nd level city domains

fnevesbr:


fnevesbr:
I'll send a clean request without the old data already merged

PR #518, opened 2017-08-28
Remove altoadige.it from list

gsmember:
altoadige.it is not a TLD

PR #549, opened 2017-10-30
remove altoadige.it

FreddyFY:
on behalf of altoadige.it


daftano:
That's a duplicate of #518

PR #551, opened 2017-10-31
Adding cloud66.ws customer domain for Cloud66

khash:
cloud66.ws is a domain used by Cloud 66 (www.cloud66.com) customers to power their websites.


khash:
Added record `_psl.cloud66.ws`

PR #292, opened 2016-08-18
Add alwaysdata.net.

cbay:
[alwaysdata](https://www.alwaysdata.com) is a web hosting company. Each of our customers has a alwaysdata.net subdomain.

PR #555, opened 2017-11-15
Fix alwaysdata.net.

cbay:
I made a mistake when I added alwaysdata.net in #292: our customers have their own subdomain foo.alwaysdata.net, so the suffix is actually alwaysdata.net, not *.alwaysdata.net.


cbay:
```
$ dig _psl.alwaysdata.net TXT +short
"https://github.com/publicsuffix/list/pull/555"
```

PR #561, opened 2017-11-19
BoltN NIC semi-TLDs

Moritz30:
On these domains subdomains are delegated down to customers using NS DNS records.


Moritz30:
The records for DNS verification have been added to the DNS zones.

PR #562, opened 2017-11-20
Add 1Password domains

roustem:
1Password service creates separate subdomain for every team and family account. For example, `https://agilebits.1password.com` and `https://anotherteam.1password.com` should be treated as different sites.

1Password has three geographical regions: 1password.ca, 1password.com, and 1password.eu.


weppos:
PS. Thanks for the amazing software @roustem, I'm an happy customer since 10y 👏

PR #567, opened 2017-11-27
add fvds.ru in public_suffix_list.dat

Shatalin89:
add fvds.ru


frju365:
Pls correct this and add a description. Just description like "add ...domain.tld" is not enough for validation.

PR #573, opened 2017-12-05
Add linkyard.cloud

msiegenthaler:
We are hosting Atlassian products -as-a-Service under the linkyard.cloud domain.
Since each customer is run on his own instances I'd be great to have cookie isolation. The customer sites are run as xyz.linkyard.cloud.


msiegenthaler:
Sorry about that, the TXT records got lost on a DNS provider change. Should be fine now.

PR #586, opened 2018-01-05
remove wildcard from compute-1.amazonaws.com

shapor:
z-{1,2}.compute-1.amazonaws.com was consolidated to *.compute-1.amazonaws.com but since new instances are being assigned names directly under compute-1.amazonaws.com, this is not exactly accurate.


shapor:
@breadtk is this acceptable?

sleevi:
Please do not open pull requests unless you are the domain owner. From other comments, it seems you are not.

PR #605, opened 2018-02-20
Add mozilla-iot.org

benfrancis:
Mozilla offers free subdomains for Web of Things gateways in people's homes as part of an experimental "Project Things". See https://iot.mozilla.org

Each subdomain points at a different web server under a different user's control so cookies should not be shared across subdomains.

Please use email authentication if possible.

Thanks!

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -11834,6 +11834,10 @@ azurewebsites.net
 azure-mobile.net
 cloudapp.net
 
+// Mozilla Corporation : https://mozilla.com
+// Submitted by Ben Francis <bfrancis@mozilla.com>
+mozilla-iot.org
+
 // Mozilla Foundation : https://mozilla.org/
 // Submitted by glob <glob@mozilla.com>
 bmoattachments.org


sleevi:
Hi @benfrancis - It would definitely be better to use DNS-based authentication, in part because we want to ensure that owners are maintaining that record (e.g. in the event of domain transfer)

Is it possible to set that up? If not, can you explain why?

benfrancis:
Hi @sleevi, yes that's possible, just slightly inconvenient.

Please bear with me while I discuss with our team.

Thanks

benfrancis:
The TXT record should now be live for _psl.mozilla-iot.org

Please can you confirm?

weppos:
@benfrancis thanks, I was able to validate it. However, as per https://github.com/publicsuffix/list/wiki/Guidelines I will have to kindly update the PR to keep the entries in order by company.

If you follow that order you will see there is already a Mozilla block. You can append your changes there:

```
+// Submitted by Ben Francis <bfrancis@mozilla.com>
+mozilla-iot.org
```

benfrancis:
Ah, sorry about that. I added a new section for Mozilla Corporation in the correct place (technically this is a request from the Mozilla Corporation, not Mozilla Foundation).

PR #616, opened 2018-03-08
add *.uberspace.de

wernermo:
We're a hosting company and we're using *.uberspace.de for our customers content. Every account gets its own mydomain.*.uberspace.de domain. For cookie isolation it would be great to get included in the public suffix list.


wernermo:
```
➜  ~ dig +short TXT _psl.uberspace.de
"https://github.com/publicsuffix/list/pull/616"
```

PR #622, opened 2018-03-20
cloud.muni.cz cloud subdomains

Mamut3D:
Arranging for appropriate TXT records now.

This is a public IaaS cloud service for academic use, provided by cooperation of https://www.metacentrum.cz/en, http://www.ics.muni.cz/ and https://www.cerit-sc.cz/.

The DNS structure is flat. Domain names used for virtual machines provisioned in the cloud are for instance:

snowflake252.flt.cloud.muni.cz
snowflake215.flt.cloud.muni.cz

We will also deploy a Dynamic DNS service, where users will be able to set up DNS names for their virtual machines. These names will go to the usr zone, i.e., the names will be:

foo.usr.cloud.muni.cz
bar.usr.cloud.muni.cz

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -11824,6 +11824,11 @@ hb.cldmail.ru
 cloud.metacentrum.cz
 custom.metacentrum.cz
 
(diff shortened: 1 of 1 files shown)


Mamut3D:
_psl TXT records are in place.

PR #632, opened 2018-04-04
Remove 1password domains

roustem:
We had a few issues and for now it would be easier if we do not have 1password domains listed in PSL.

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -10640,12 +10640,6 @@ cc.ua
 inf.ua
 ltd.ua
 
-// AgileBits Inc : https://agilebits.com
-// Submitted by Roustem Karimov <roustem@agilebits.com>
-1password.ca
-1password.com
-1password.eu
-
 // Agnat sp. z o.o. : https://domena.pl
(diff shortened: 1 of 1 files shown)


sleevi:
Can you indicate more about the issues you’ve run into?

roustem:
One of the issues was related to DMARC where emails were not delivered. I believe that FastMail got them resolved but I am worried about other providers running into similar issues. Here is the quote from the email:

> The DKIM alignment check it failing because 1password.com appears on the public suffix list.
>
> RFC7489 (DMARC) section 3.3.1 talks about not allowing domains listed in the psl to align in relaxed mode, and the library we are using for DMARC validation (Mail::DMARC) detects this domain is on the list and returns a fail result for alignment.


We are also not sure if PSL has any effect on Google search results and domain rankings because subdomains are treated differently.

roustem:
```
$ dig TXT _psl.1password.com +short
"https://github.com/publicsuffix/list/pull/632"
$ dig TXT _psl.1password.ca +short 
"https://github.com/publicsuffix/list/pull/632"
$  dig TXT _psl.1password.eu +short
"https://github.com/publicsuffix/list/pull/632"
```

PR #633, opened 2018-04-04
Add cnpy.gdn as private domain

angelogladding:
`cnpy.gdn` will serve as a namespace for customers' personal websites.

    alice.cnpy.gdn
    bob.cnpy.gdn
    carol.cnpy.gdn

Each site should stand independent. Cookies should not be shared across subdomains.

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -10940,6 +10940,10 @@ cloudns.us
 // Submitted by Stefan Dimitrov <contact@cloudeity.com>
 cloudeity.net
 
+// CNPY : https://cnpy.gdn
+// Submitted by Angelo Gladding <angelo@lahacker.net>
+cnpy.gdn
+
 // CoDNS B.V.
 co.nl
 co.no


angelogladding:
> dig TXT _psl.cnpy.gdn +short
    "https://github.com/publicsuffix/list/pull/633"

sleevi:
I can't find any reference to this domain, nor does it appear to be hosting any meaningful content. Further, the registration seems to expire 4 months from now. Can you speak more about this domain and its purpose?

angelogladding:
I extended the domain registration to 2021.

As for the timing it seems there's a little bit of a chicken and egg problem here. I may have jumped the gun but I really just wanted to get ahead of future problems.

I'm building a decentralized social network. It leverages the IndieWeb and thus personal websites to host the identities and content of the network. The webapp I'm writing is called `canopy` and can be obtained at https://understory.live while https://cnpy.gdn will be a near-future simplified host of canopy trees.

At first CNPY.GDN (suffix in question) will pair a hosted instance to a subdomain (eg. alice.cnpy.gdn). This leaves me in control of the software provisioned at my subdomains. HOWEVER, to prevent creating yet another silo I want to offer my customers the ability to download their data, host it somewhere else and have their subdomain point to the new IP. At this point their canopy instance will no longer be under my control and would leave my remaining customers exposed to cookie manipulation.

Let's Encrypt has already given me headroom for TLS cert signing and I fully understand the consequences of addition to the PSL. I can setup a static site at https://cnpy.gdn if that'd help or do I need to have a fully functioning system before I proceed?

PR #635, opened 2018-04-05
These domains are used to host member subdomains.

jmcclelland:
Members can create their own subdomains using either
[member].mayfirst.org or [member].mayfirst.info.

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -11826,8 +11826,8 @@ barsy.support
 
 // May First - People Link : https://mayfirst.org/
 // Submitted by Jamie McClelland <info@mayfirst.org>
-mayfirst.org
 mayfirst.info
+mayfirst.org
 
 // Mail.Ru Group : https://hb.cldmail.ru
 // Submitted by Ilya Zaretskiy <zaretskiy@corp.mail.ru>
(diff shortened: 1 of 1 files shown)


sleevi:
That's not sufficient information to determine if addition to the PSL is necessary. As per the [Guidelines](https://github.com/publicsuffix/list/wiki/Guidelines), can you please provide additional detail explaining.

Note, it appears you're hosting pages on `www.mayfirst.org`, `support.mayfirst.org`, `members.mayfirst.org`, `mayfirst.org`, and possibly others. Are you aware that addition to the PSL may impact these services by segmenting them out?

jmcclelland:
Thanks for the response.

We are a membership organization and have been allowing our members to create subdomains of mayfirst.org. In addition, we run our own services on this domain (such as the ones you have identified).

Our long term plan is to phase out member subdomains using .mayfirst.org and transition member sites to mayfirst.info - however, until then, we want to protect our members and our own sites from domain-wide cookies using the mayfirst.org domain.

We do not set any domain-wide cookies on our own services (support, www, etc.) and are prepared for any adverse consequences of this policy (we have determined that the security trade-off is worth it).

Please let me know if there is any other information we can provide to help with this application.

PR #637, opened 2018-04-09
add azurecontainer.io

jluk:
Users may create an arbitrary amount of subdomains for off of this one.

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -11839,12 +11839,12 @@ eu.meteorapp.com
 // Michau Enterprises Limited : http://www.co.pl/
 co.pl
 
-// Microsoft : http://microsoft.com
+// Microsoft Corporation : http://microsoft.com
 // Submitted by Justin Luk <juluk@microsoft.com>
+azurecontainer.io
 azurewebsites.net
 azure-mobile.net
 cloudapp.net
-azurecontainer.io
 
 // Mozilla Corporation : https://mozilla.com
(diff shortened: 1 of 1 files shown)


sleevi:
It looks like you're waiting for the TXT record to be set. Could you explain why you didn't combine this with the previous Microsoft entry that @blowdart added?

jluk:
@sleevi wasn't sure how to designate a different contact. I could slate it under the previous Microsoft entry with a dedicated comment with my info if that's preferred.

blowdart:
Yea, I've moved on from Azure security now, so the removal of my name is a bonus :)

blowdart:
I'll start up a discussion internally with the azure folks to get a DL set up anyway, but @jluk you should head the DNS route anyway. That came in long after I did my additions, so ... I didn't have to do that work *grin*

jluk:
Update TXT record is in place.

```
:~$ dig TXT _psl.azurecontainer.io +short
"https://github.com/publicsuffix/list/pull/637"
```

sleevi:
@blowdart Just since you were the last to touch, and because of https://github.com/publicsuffix/list/pull/637#issuecomment-379899227 , I just wanted to make sure y'all are good landing as is, as it seems to be in order (now that I resorted)

blowdart:
Yea that discussion has started, and is now out of my hands. Thank goodness :) I'll let azure mull over what they want to do. But yea, it's fine by me.

PR #642, opened 2018-04-17
Add nic.ro TLDs

DDuarte:
> For example local or global Business with .CO.RO where 'co' is for co-mmercial.
> Or a srl Company "Societate cu răspundere limitată" with our .SRL.RO extension.
> Or a Corporation with .SA.RO ("Societate pe Actiuni").
> An or-ganisation with the .OR.RO suffix

Source: http://co.ro/ http://www.nic.ro/

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -5790,16 +5790,20 @@ asso.re
 com.re
 nom.re
 
-// ro : http://www.rotld.ro/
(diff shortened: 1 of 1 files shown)


sleevi:
nic.ro is not the ccTLD operator - roTLD is, and the authorative document is http://rotld.ro/reguli-de-inregistrare/

Instead, it appears nic.ro is just like any other registrar selling subdomains, and as such, nic.ro would need to make the request.

PR #610, opened 2018-02-28
Add MyDNS.JP suffixes

disco-v8:
Hello!

My Service, https://www.mydns.jp is a Dynamic DNS Service in JAPAN.
I providing subdomains for users.
Please add these domains to PSL. :-)

Regards.

PR #648, opened 2018-04-25
fix syntax in public_suffix_list.dat

mattn:
I guess that why many contributor mistake syntax in public_suffix_list.dat is you don't enable travis for pull-request. Current public_suffix_list.dat is already broken.

If you enable travis for pull-request, most of mistakes in PR will be solved. (like #610)

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -7204,7 +7204,7 @@ auspost
 // author : 2014-12-18 Amazon Registry Services, Inc.
 author
 
-// auto : 2014-11-13 Cars Registry Limited 
+// auto : 2014-11-13 Cars Registry Limited
 auto
 
 // autos : 2014-01-09 DERAutos, LLC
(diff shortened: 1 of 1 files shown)


sleevi:
Are you suggesting the white space is causing trouble for you? Implementations should be ignoring any characters after a comment.

weppos:
This section is generated automatically, hence I'm prone to reject this change. It will be re-introduced at the next section update (unless the script is updated to strip the spaces).

Do you want to include that change in your PR @mattn?

sleevi:
@weppos I’m not sure we should take that change even if they did. We should work to understand why the whitespace causes any issues (if that is the case), so as to determine if it’s a bad implementation.

mattn:
>Do you want to include that change in your PR @mattn?

No, this is just notification for that you need to enable travis for pull-request.

weppos:
Travis was enabled in the past. I need to investigate why it stopped. However, should we re-enable it, it will completely ignore those spaces.

@mattn those are comments, spaces in there are not significant as @sleevi pointed out.

weppos:
https://travis-ci.org/publicsuffix/list

Travis is enabled by default. This PR passed the build (as expected).
https://travis-ci.org/publicsuffix/list/builds/370889334

mattn:
I thought that why #610 is not merged is that it doesn't comply with guidelines.

https://community.letsencrypt.org/t/limitation-of-jp-tld-suffix/53063/7?u=mattn

PR #650, opened 2018-04-30
Add *.otap.co

govert-co-co:
Co & Co is a bespoke software developer for a wide variety of clients.
We use otap.co as a domain for test environments, acceptance environments and on the fly hosting of short and long lived experiments. We would like to add stronger (cookie) isolation between these domains, as well as benefit from any additional effects the PSL might bring us in terms of browser UI, etc.

Domain names on otap.co always have four labels total in the form of {project}.{client}.otap.co, hence the wildcard.

Examples:
a.x.otap.co should be isolated from b.x.otap.co. Both should also be isolated from c.y.otap.co.
inner1.a.x.otap.co is not isolated from a.x.otap.co.

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -10924,6 +10924,10 @@ cloudcontrolapp.com
 // co.ca : http://registry.co.ca/
 co.ca
 
+// Co & Co : https://co-co.nl/
+// Submitted by Govert Versluis <govert@co-co.nl>
+*.otap.co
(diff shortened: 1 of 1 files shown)


govert-co-co:
TXT record has been added:
```
;; ANSWER SECTION:
_psl.otap.co.           3594    IN      TXT     "https://github.com/publicsuffix/list/pull/650"
```

PR #656, opened 2018-05-11
Add cloudycluster.net

exiledcole:
CloudyCluster is used to quickly spin up HPC Clusters and manage them from mobile or desktop devices.   We assign subdomains of cloudycluster.net to clusters that our customers start.  For security reasons, different clusters should not share cookies and other data.  We use cloudycluster.net only for our customers' subdomains.

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -12227,6 +12227,10 @@ nom.vg
 // Submitted by Andrew Sampson <andrew@ulterius.io>
 cya.gg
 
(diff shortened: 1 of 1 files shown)


sleevi:
Note you will need to add a TXT record to validate, as per the [Guidelines](https://github.com/publicsuffix/list/wiki/Guidelines)

exiledcole:
We've changed the DNS servers and it has to propagate, it should be good tomorrow.  Thanks!

exiledcole:
The TXT record is added and is ready for you to validate.

PR #649, opened 2018-04-25
Add funkfeuer.at delegations

wnagele:
FunkFeuer is an association with various initiatives throughout Austria. These initiatives all have their own sub-domain within the funkfeuer.at namespace. These sub-domains are fully delegated out to the initiatives and allocate their member domains underneath as they see fit.

PR #658, opened 2018-05-13
Add wien.funkfeuer.at

wnagele:
FunkFeuer is an association with various initiatives throughout Austria. These initiatives all have their own sub-domain within the funkfeuer.at namespace. These sub-domains are fully delegated out to the initiatives and allocate their member domains underneath as they see fit. This adds the community sub-domain for Vienna.

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -11664,6 +11664,10 @@ freeboxos.fr
 // Submitted by Daniel Stone <daniel@fooishbar.org>
 freedesktop.org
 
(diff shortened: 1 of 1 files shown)


wnagele:
Replacement for #649. Each initiative will have to make their own pull request as they are delegated out and we cannot add the _psl records from the parent zone.

wnagele:
Verification record added:
```
$ dig _psl.wien.funkfeuer.at txt +short
"https://github.com/publicsuffix/list/pull/658"
```

PR #668, opened 2018-05-22
Adds hasura.app to the list

shahidhk:
Similar to `hasura-app.io`, with the introduction of new `app` TLD, Hasura allocates sub-domains on `hasura.app` for the clusters users create on it's platform. A user should not be able to set cookies across these subdomains, apart from their own.


shahidhk:
Here is the authentication:
```
dig TXT +short _psl.hasura.app
"https://github.com/publicsuffix/list/pull/668"
```

shahidhk:
Thanks for the quick merge @weppos :+1:

PR #669, opened 2018-05-23
Update gu ccTLD suffixes

eadanzo:
Update gu ccTLD suffixes

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -1114,8 +1114,18 @@ mil.gt
 net.gt
 org.gt
 
-// gu : http://gadao.gov.gu/registration.txt
-*.gu
+// gu : http://gadao.gov.gu/register.html
+// University of Guam : https://www.uog.edu
+// Submitted by uognoc@triton.uog.edu
+gu
+com.gu
+edu.gu
+gov.gu
+guam.gu
+info.gu
+net.gu
+org.gu
+web.gu
 
 // gw : https://en.wikipedia.org/wiki/.gw
 gw


sleevi:
Thanks for submitting this. Do you have any links to details about these policies?

The current entry would totally work and would be both equivalent and more efficient than your proposed entry, so trying to understand a bit more about the registration policies and goals.

Thanks!

eadanzo:
Hi, we've just inherited the responsibilities for the .gu ccTLD.
The previous domain operator didn't have a formal documentation. 

We are currently working on updating and standardizing our policies.

sleevi:
Do you believe the current entry is wrong or causing some compatibility issues?

eadanzo:
The current entry does not allow us to add the suffixes into our Internet2 InCommon SSL Certificate Manager because the suffixes does not exist in the public suffix list. Per Comodo Support.

Updating the .gu suffixes will allow the University of Guam to manage the suffixes in the cert manager.

sleevi:
@eadanzo Thank you, this is very interesting, and a use case that we were not previously aware of! It would be really helpful to understand more about this problem or use case, including example domains that you're having trouble with getting certs for.

I'd also be happy to personally follow-up with Comodo, with your permission, to better understand the challenges you're facing. If you have a support ticket with them, I can reach out to their PKI compliance team to better understand how they've implemented these PSL changes.

eadanzo:
We have added these domains/suffixes into our cert-manager:

com.gu
edu.gu
gorex.gu
gov.gu
guam.gu
mariix.gu
net.gu
org.gu
web.gu

We were not able to validated them because PSL only listed *.gu per comodo support.
Which led us to update the .gu ccTLD PSL.

sleevi:
Thanks. I'll follow-up with the Comodo team to understand more about their implementation and how they consume the PSL.

One question - looking at the domains you listed, `gorex.gu` and `mariix.gu` are currently being treated as public suffices (under the `*.gu` rule), but it sounds like they may not be intended to. I noticed `gorex.gu` is just a redirect, but `mariix.gu` looks like it's actually hosting content, so your change is probably the correct one :)

Probably the easiest way to expedite the approval of this process if you would be able to add a note to http://gadao.gov.gu/register.html that just indicates that you will allow registration of top-level `.gu` domains. That's not clear from the Item 1 and Item 2 on your informative list, and is the only thing preventing me from merging this in :)

"Registration requests may be made for second-level domains under `.gu` or as third-level domains under the following domains: `.com.gu`, `.edu.gu`, `.gov.gu`, `.guam.gu`, `.info.gu`, `.net.gu`, and `.web.gu`"

If that's possible to make, that'd be amazing! Thanks!

sleevi:
Thanks so much!

PR #676, opened 2018-06-07
Add pages.sh

wjh:
Pages.sh is the domain used to provide GitLab pages access for members of the fork.sh community. As they each get a subdomain, I made this request - it is also recommended by the GitLab documentation.

(sorry for the second PR, managed to mess up the first one)


wjh:
It is not, I'll close this - thank you though!

PR #268, opened 2016-07-18
Add gitlab.io

ahanselka:
GitLab.io is used for GitLab Pages and thus allows arbitrary users to have subdomains. You can see the official request to add GitLab.io on [our issue tracker](https://gitlab.com/gitlab-com/infrastructure/issues/230).

PR #679, opened 2018-06-10
Added *.rhcalab.app

dean1012:
For GitLab pages as per https://docs.gitlab.com/ee/administration/pages/ to avoid issues related to supercookies

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -11807,6 +11807,10 @@ hashbang.sh
 hasura.app
 hasura-app.io
 
+// RHCALAB Gitlab : https://gitlab.rhcalab.net
+// Submitted by Jerry Smith <jdsmith1012@me.com>
+*.rhcalab.app
+
 // Hepforge : https://www.hepforge.org
 // Submitted by David Grellscheid <admin@hepforge.org>
 hepforge.org


dean1012:
Verification `dig TXT _psl.rhcalab.app` shows #679

sleevi:
I'm sorry, that's not a sufficient explanation. Could you please explain more?

dean1012:
@sleevi Members of the public will be able to create subdomains on the rhcalab.app domain for the purpose of hosting static web pages for projects.

GitLab is the software that will facilitate this. It is just like GitHub pages.

There is a domain that already hosts static sites like this on your list: gitlab.io

dean1012:
@sleevi Please see #268 for the inclusion of gitlab.io - my reasons are exact to theirs with same product in use.

I hope this and my previous response will serve as sufficient explanation but please let me know if I need to clarify further.

sleevi:
Mere creation of subdomains does not appear to be a clear need. I'm hoping you can expand a bit more about your use case.

dean1012:
@sleevi The documentation for installing and configuring GitLab Pages which I've already linked specifies a strong recommendation for adding the wildcard domain to the Public Suffix List as they did in #268 

With full respect to your project, if one domain has already been added for this reason I fail to see why this one cannot be.

It is my understanding that each subdomain must be treated separately by browsers for security and privacy purposes.

A.rhcalab.app should not be able to view, set, or modify cookies for B.rhcalab.app or vice versa. They should be treated as entirely separate entities by browsers which as I understand your project facilitates.

Subdomains under this domain will be created automatically by user request through the GitLab software and pages served under those subdomains will be controlled and managed by those customers.

As another example of a domain added for similar reasons, please check #656 

Thank you again for your time and consideration. I hope the above is enough justification.

sleevi:
We've seen quite a few projects inappropriately document steps to request inclusion in the PSL, and so understanding the use case and intent is a necessary part. For example, this may prevent rhcalab.app from being navigated to, and thus its necessary to ensure that submissions have done more than read some projects documentation, but understand and acknowledge the implications of the addition - as inclusion (within browsers) can take 6 months to a year, and the same for removal. As such, regardless of the existence of the GitLab documentation, each request - and its justification - is treated independently. This is especially true given the size constraints of the PSL and the need to ensure that requests have appropriate justification.

The technical details will be reviewed during the next review batch, which may permit inclusion. The Guidelines attempt to document the necessary steps and information to provide, which your most recent comment seems to have provided.

dean1012:
@sleevi Thank you.

I actually did NOT know about most of the things you mentioned (the ~6 month timeframe, the fact that rhcalab.app itself may not be navigable, ...). However, I do understand now thanks to your explanation and I also understand why we went back and forth so much.

I now understand what's going on and I believe this is still the right thing to do. rhcalab.app itself will not be used in any way, only subdomains created by customers as stated before.

I will await your decision and will be ready to respond if further clarification or acknowledgement is needed.

Have a good day.

dean1012:
Cancelling.

PR #680, opened 2018-06-12
Adding missing No-IP Domains to the PSL.

noipcom:


sleevi:
These do not comply with the [Guidelines](https://github.com/publicsuffix/list/wiki/Guidelines)

noipcom:
@sleevi Can you please elaborate on what does not comply? I read through the guidelines and the only issue I am seeing is the suffix list is not sorted properly. Is this the issue you're referring to? If so I will be glad to correct it. 

Please note I've also spoken with another project maintainer via email to verify the domains in the past as well.

sleevi:
Sorting, motivation, examples, and domain validation, all documented in the
Guidelines.

PR #681, opened 2018-06-14
Add run.app and a.run.app to the psl

apolito:
both *.run.app and *.a.run.app will run user code

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -11713,6 +11713,8 @@ goip.de
 
 // Google, Inc.
 // Submitted by Eduardo Vela <evn@google.com>
+run.app
+a.run.app
 *.0emm.com
 appspot.com
 blogspot.ae


apolito:
DNS entries are in
dig TXT +short _psl.run.app  
"https://github.com/publicsuffix/list/pull/681"
dig TXT +short _psl.a.run.app  
"https://github.com/publicsuffix/list/pull/681"

<projectname>.run.app and <projectname>.a.run.app will be serving user code separated by project.  This is very similar (but a different use case than) appspot.com or cloudfunctions.net.  Currently there is no public usage, but since it takes a while to get into the PSL, we would like to have it available before release.

apolito:
ping?

frju365:
You will have to wait until @weppos or @sleevi wheck and merge your request.

Can we have a short description of your orga too ? What is the goal of this orga ? (it's just to legitimate your PR).

apolito:
Apologies if my comment in on Jun 15 wasn't clear.

This is not a private domain, this is a public google cloud platform offering that will be used to run multiple customer projects, routed via the label that comes before the run.app or a.run.app domain suffixes.  (So foo.run.app will route to on project, but bar.run.app to another) We want to be in the PSL to prevent cross project cookies being set.

This is very similar to the use case for appspot.com and cloudfunctions.net, both in usage and reasons for inclusion (which are in the same section of the PSL)

The reason we need run.app and a.run.app is because there will be two ways to address a project, foo1.a.run.app and foo2.run.app may end up routing to one project, but bar1.run.app and bar2.a.run.app will route to another.  It's outside of the scope of the PSL but does mean we need both suffixes in the PSL.

PR #683, opened 2018-06-18
Add ssl.fun

berstend:
ssl.fun is a free service allowing users to register subdomains programmatically.

The wildcard is set in accordance with the allowed format:

```bash
# valid
example.com.ssl.fun
foobar.de.ssl.fun
baz.foobar.de.ssl.fun

# invalid
example.ssl.fun
com.ssl.fun
```

DNS ownership proof:
```bash
❯❯❯ dig TXT +short _psl.ssl.fun
"https://github.com/publicsuffix/list/pull/683"
```

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -12432,6 +12432,10 @@ bounty-full.com
 alpha.bounty-full.com
 beta.bounty-full.com
 
+// ssl.fun : https://ssl.fun
+// Submitted by Markus Hänel <ssl@ssl.fun>
+*.ssl.fun
+
 // staticland : https://static.land
 // Submitted by Seth Vincent <sethvincent@gmail.com>
 static.land


sleevi:
Can you provide more details about your service? I believe this is
something we will need to close as WontFix.

berstend:
Of course :-) I submitted the request early (before all documentation is available) as I didn't expect such quick response.

The service offered by ssl.fun is quite unique (I didn't see anything comparable).

- It allows users to programmatically register a subdomain which follows their production domain closely (e.g. example.com > example.com.ssl.fun)
- It integrates with common acme clients and responds to the DNS-01 ownership challenge
- By default all A records point to `127.0.0.1`

As a result the user (target audience developers) are able to test things locally with a valid https certificate.

Please note:
- This differs quite a lot from previous attempts to solve this problem
- The users themselves do the certificate request directly and only they have access to the private key
  - Hence the certificate is not compromised, which was the main issue with spiritual successors in the past :-)
- I've read the Let's Encrypt guidelines closely and as long as the private key is not compromised there shouldn't be a problem
- I'm aware of alternative solutions (e.g. self-signed certs) but unfortunately those bring their own set of problems

sleevi:
Thanks for clarifying. We will not be proceeding forward with this request.
I’m sorry, your use case is not supported, as such designs are
fundamentally insecure.

If you’re willing to accept that insecurity, you can contact Let’s Encrypt
to see about raising your rate limits. However, the PSL will not be able to
help you improve that.

sleevi:
Closing as WontFix, as DNS mirrors are not supported.

berstend:
No worries @sleevi, thanks for your time :-)

I realized I might've been a bit quick in this regard, I instead did a RFC for this approach at LetsEncrypt to validate and discuss the idea beforehand:
https://community.letsencrypt.org/t/rfc-a-way-to-use-valid-https-with-local-hosts/64596

PR #684, opened 2018-06-21
Add pub.cloud.scaleway.com

jtherin:
Scaleway is a cloud provider, we assign a subdomain under pub.cloud.scaleway.com for each instance. For security reasons, different instances should not share cookies and other data.

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -12378,6 +12378,10 @@ sandcats.io
 logoip.de
 logoip.com
 
+// Scaleway : https://www.scaleway.com/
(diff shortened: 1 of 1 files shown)


jtherin:
$ dig TXT _psl.scaleway.com +short
"https://github.com/publicsuffix/list/pull/684"

frju365:
normally th cloud record title must be : _psl.complete.prefix, so in your case it mus be  : _psl.pub.cloud.scaleway.com 
The reason is that we validate dns txt record per prefix.

weppos:
Closing due to no follow up.

PR #688, opened 2018-07-04
Added antiquariatssoftware.com for our customers pages

beepab:
Hello,
we host a software (online-shop) for our customers and they get a subdomain (e.g paket1.antiquariatssoftware.com) for their shop hosted by us. We ran into problems with the letsencrypt rate-limit for subdomains so that some of our customers can't secure their pages with SSL now. Please consider adding antiquariatssoftware.com to the list.

Kind regards
Pasqual B.

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -12658,4 +12658,8 @@ now.sh
 // Submitted by Su Hendro <admin@zone.id>
 zone.id
 
+// Guthschrift-Systems : http://guthschrift-systems.com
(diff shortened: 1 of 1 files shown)


weppos:
The correct way to request a limit override is via LE form
https://letsencrypt.org/docs/rate-limits/#a-name-overrides-a-overrides

beepab:
Could you please elaborate what makes us different to a hosting provider giving subdomains to customers webpages they host? Cookie data and such should not be shared across this subdomains.
I thought that was exactly what this list is for?

PR #272, opened 2016-07-26
Add stackspace.space

linahe:
Stackspace allows customers to launch and use big data tools hosted on subdomains of stackspace.space.

PR #691, opened 2018-07-12
Remove stackspace.space from PSL

dbkegley:
We would like to remove our domain's entry from the PSL.

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -12451,10 +12451,6 @@ spacekit.io
 // Submitted by Stefan Neufeind <info@speedpartner.de>
 customer.speedpartner.de
 
-// Stackspace : https://www.stackspace.io/
-// Submitted by Lina He <info@stackspace.io>
-stackspace.space
-
 // Storj Labs Inc. : https://storj.io/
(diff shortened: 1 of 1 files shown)


dbkegley:
```
$dig TXT +short _psl.stackspace.space
"https://github.com/publicsuffix/list/pull/691"
```

weppos:
Related to https://github.com/publicsuffix/list/pull/272

Out of curiosity, can you please share the reason of the removal?

dbkegley:
We've been having trouble renewing our wildcard cert for our domain from namecheap and thought this may be related

dbkegley:
COMODO is the CA for our cert through Namecheap. @weppos if you have any ideas as to why our request for a new wildcard is being denied we'd like to hear your thoughts. Our wildcard was issued initially before we added our domain to the PSL but it has since expired so we need to provision a new one.

weppos:
Sorry for the delay @dbkegley . I went ahead and removed the entry. It's not the first time I hear about issues with Comodo. If I recall correctly @sleevi got the same feedback in another PR.

@sleevi did you have the chance to follow up with Comodo?

PR #703, opened 2018-08-13
Update public_suffix_list.dat for Indonesia

nfrozi:
Adding ponpes.id, new TLD in Indonesia. Ref: https://pandi.id/en/domain/registration-requirements


weppos:
Thanks.

PR #708, opened 2018-09-11
Add public suffix entries for dapps.earth

burdakovd:
# Reason of this pull request

[dapps.earth](https://dapps.earth/) is an HTTP gateway into IPFS/Swarm.

Anyone can effectively create websites at the following subdomains (by uploading content into IPFS or Swarm/ENS):
1) `*.ipfs.dapps.earth` (example: `bafybeiemxf5abjwjbikoz4mc3a3dla6ual3jsgpdr4cjr3oz3evfyavhwq.ipfs.dapps.earth`)
2) `*.bzz-immutable.dapps.earth` (example: `h4cpab3mz443iehtiipfi5vj46pnytrspvm5peu2u2wz7rz7m4vq.bzz-immutable.dapps.earth`)
3) `*.*.bzz.dapps.earth` (example: `theswarm.eth.dapps.earth`)

Therefore, to isolate cookies of all hosted websites, I suggest adding the following public suffix records:
1) `*.dapps.earth`
2) `*.bzz.dapps.earth`

Similarly for `staging.dapps.earth`, which runs testing version of the same website.

# DNS verification

```
dig +short TXT _psl.dapps.earth
"https://github.com/publicsuffix/list/pull/708"
dig +short TXT _psl.bzz.dapps.earth
"https://github.com/publicsuffix/list/pull/708"
dig +short TXT _psl.staging.dapps.earth
"https://github.com/publicsuffix/list/pull/708"
dig +short TXT _psl.bzz.staging.dapps.earth
"https://github.com/publicsuffix/list/pull/708"
```

make test
=========
Passed: http://termbin.com/ym4m

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -11084,6 +11084,11 @@ firm.dk
 reg.dk
 store.dk
 
+// dapps.earth : https://dapps.earth/
+// Submitted by Daniil Burdakov <icqkill@gmail.com>
+*.dapps.earth
+*.bzz.dapps.earth
+
 // Debian : https://www.debian.org/
 // Submitted by Peter Palfrader / Debian Sysadmin Team <dsa-publicsuffixlist@debian.org>
 debian.net


sleevi:
Many PSL implementations do not support multiple wildcards (`theswarm.eth.dapps.earth` / `*.*.bzz.dapps.earth`), so that sounds like it will potentially be a deal breaker.

burdakovd:
@sleevi but PSL itself is just single wildcard, e.g. `*.dapps.earth`. Or do you mean that implementations do not even support a single wildcard in PSL?

burdakovd:
If even a single wildcard is not supported, I could fall back to the following PSLs:
```
ipfs.dapps.earth
bzz-immutable.dapps.earth
eth.bzz.dapps.earth
```

It would limit ENS domains to be in `.eth` zone, which is the only one in use right now. If another zone appears, the list would need to be updated, but this is OK, since adding another zone is manual work anyway due to the need of requesting extra wildcard SSL certificate.

sleevi:
Single wildcards are supported, multiple wildcards are not.

sleevi:
Apologies, I misread the earlier explanation as describing multiple wildcards, rather than the single wildcard.

The second part - we try very hard not to include 'staging' or 'test' domains which are specific to a particular user/use case. Can  you explain more as to why they're needed?

burdakovd:
I guess including staging version is not critical. `staging` is needed to test new versions of gateway before they go live.

It is marginally safer if the same protection against cookies being set would apply there (in case someone opens a page on staging website). Also it is somewhat more convenient to do automated testing, if you know staging and prod domains behave the same (whereas if we submit just one to PSL, they may start behaving differently in browsers).

However I understand your reluctance to add staging/test domains (i.e. I did not include staging-2), and can remove staging from pull request if needed.

Looking at the list, I see just one precedent of staging/test domains being included:
```
cust.dev.thingdust.io
cust.disrec.thingdust.io
cust.prod.thingdust.io
cust.testing.thingdust.io
```

burdakovd:
I have updated the pull request to exclude staging subdomain.

burdakovd:
Any update here? It's been open for three months.

PR #46, opened 2015-11-06
Add mybluemix.net

IvanGoncharov:
See this article: https://developer.ibm.com/bluemix/2014/06/16/introducing-mybluemix-net-new-domain-application/

PR #720, opened 2018-10-28
Convert list data from NFKD to NFKC

### LINKED ISSUE #715: Use normalized non-ASCII names
Some international domain names / rules use composed characters, others use decomposed chars.

E.g. `läns.museum` is composed while `südtirol.it` is decomposed.

For comparison of strings it would be nice to have just one *or* the other. Since older entries are composed, I suggest NFKC as described at https://en.wikipedia.org/wiki/Unicode_equivalence#Normalization.

rockdaboot:
Fixes #715

As described in #715, decomposed characters break certain configurations of libpsl.
Decomposed characters will result in wrong punycodes if not composed before.

<!--


rockdaboot:
The suggestion is based on standards which say (in very short) first convert to NFKC then to punycode.

Just quickly found [here](https://unicode.org/faq/idn.html):
> How are label delimiters handled in implementations of IDNA?
>
> The processing of UTS #46 matches what is commonly done with label delimiters by browsers, whereby characters containing periods are transformed into the NFKC format before labels are separated.

And from the [spec](https://tools.ietf.org/html/rfc5894) (3.1.3):
> There is provision for exception cases but, in general, characters
   are placed into DISALLOWED if they fall into one or more of the
   following groups:
> The character is a compatibility equivalent for another character.
      In slightly more precise Unicode terms, application of
      Normalization Form KC (NFKC) to the character yields some other
      character.

weppos:
Last ping for @sleevi 😉

sleevi:
There were no particular objections so I didn't say anything. Sorry ;)

PR #741, opened 2018-12-23
Update .TR

ilhansubasi:
Add tsk.tr


weppos:
https://www.tsk.tr/ is reachable via www hence this merge will not introduce issues to the root domain.

PR #758, opened 2019-01-23
Proposal to remove bv.nl

mdavids:
I represent the registry for the .nl ccTLD (SIDN) and would like to draw your attention to bv.nl.
To the best of our knowledge that private domain has never properly qualified for the public suffix list, even though there might have been good intentions when the initial request was made. We are not the owner of the private domain bv.nl. We did try to contact the current owner and discuss the matter, but we received no response. Since we are not sure what the correct procedures are in this case, I'd like to ask for extra attention (and help) from the maintainers of this list.

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -4659,9 +4659,6 @@ web.ni
 //      ccTLD for the Netherlands
 nl
 
-// BV.nl will be a registry for dutch BV's (besloten vennootschap)
-bv.nl
-
 // no : http://www.norid.no/regelverk/index.en.html
 // The Norwegian registry has declined to notify us of updates. The web pages
 // referenced below are the official source of the data. There is also an


sleevi:
Thanks! As you represent the .nl registry, we've got a couple options:

1) If you can point to documentation about any policies you have (about reserved 2LDs), on the registry site, then we can use that to vet that bv.nl is not a registry-operated 2LD, and remove it entirely

2) Alternatively, we could move bv.nl to the PRIVATE section. However, I'd want to do some source archaeology on understanding the original addition of bv.nl (which probably goes back some time), as it may make sense to remove it entirely.

Do you have any resources we could refer to for SIDN's operation of .nl? You can see examples in some of the other ccTLDs within the ICANN section, as we point to various TLD's policies as the sources for the included entries.

mdavids:
1) Here is some documentation:
https://www.sidn.nl/a/nl-domain-name/registering-a-domain-name?language_id=2
(scroll all the way down). Since we do not offer registry-operated 2LDs and never have, there is not a whole lot of information that explicitly states that we don't. It is all rather 'implicitly stated'.

The only reserved and excluded 2LDs are: ftp.nl, mail.nl, nl.nl and www.nl
(https://www.sidn.nl/a/nl-domain-name/what-is-a-domain-name-?language_id=2)

There are officially no registry-operated 2LDs. There is one privately owned .nl domain presenting itself as such that we know of, which is co.nl (https://domain.co.nl/en/about-conl/). 

2) bv.nl was added in 2011. It is still owned by the same company. We tried to contact them, but so far without luck. We could try again. We strongly feel that their initial plans to act like some kind of 'co.nl' have long seized to exist (fyi: BV in the Netherlands roughly translates to Ltd, so we can imagine the value of their ideas). Their website is outdated and does not mention any options for 3LDs. There are no bv.nl certificates in the transparency log. We believe it is safe to remove bv.nl from de PSL. We can probably officially inform them of this with a letter.

Please let me now if you require further information. Thank you.

sleevi:
@weppos Can you act as a sanity check? I think that the info in https://github.com/publicsuffix/list/pull/758#issuecomment-457548771 is sufficient to allow us to wholly remove it, rather than move it to the PRIVATE section.

mdavids:
> @sleevi I agree to remove it.
> 
> I tried to locate the original discussion, but the Bugzilla ticket referenced in the diff ([621707](https://bugzilla.mozilla.org/show_bug.cgi?id=621707)) is actually referring to something different.

This is the right one:
https://bugzilla.mozilla.org/show_bug.cgi?id=621710

PR #384, opened 2017-01-17
Add .RU and .SU geographic suffixes owned by FAITID

weppos:
Follow up to GH-359. See also GH-380.

This commit is also related to
https://bugzilla.mozilla.org/show_bug.cgi?id=1134993 https://github.com/publicsuffix/list/commit/601e0b094957f3289be53e50a5dbdcad6d744e6a
However, I need to follow up with FAITID because it's not clear the
boundary between the current list of .SU suffices, and the provided one.

PR #808, opened 2019-04-26
Refactor and add all missing Reserved domain names from cctld.ru

unreturned:

Reason for PSL Inclusion
====

Refactor and add all missing reserved names from https://cctld.ru/en/domains/domens_ru/reserved.php

DNS Verification via dig
=======

No. Infromation from official site of cctld.ru

make test
=========

Done. Passed https://pastebin.com/raw/Fdqe7gKk


peterthomassen:
Is the list on the linked page current? The title says "Reserved domain names (till October 01, 2009)". My understanding of that is that the page expired nearly 10 years ago.

sleevi:
I don't think we want to merge this. You'll need to speak more about the motivations for reservation. The current approach was correct, per #384 

That is, by proposing to add to a registry section, you need to demonstrate explicitly the registry policy by the registry operator. Based on https://www.iana.org/domains/root/db/ru.html , this is cctld.ru

I cannot find information indicating the relationship between faitid.org and ccTLD.ru that would indicate there is any specific relationship. Based on https://cctld.ru/ru/docs/ and the translated https://cctld.ru/files/pdf/docs/new/rules_ru-rf.pdf further support https://github.com/publicsuffix/list/pull/808#issuecomment-491953400

If there is missing information, please highlight it. Otherwise, we plan to continue as-is, consistent with the policies, and at the explicit request of the operators of these 2LD registrations.

unreturned:
Thanks for information!

Issue #831, opened 2019-05-30
The NewGTLDs script should use the ICANN JSON feed


sleevi:
As captured in https://github.com/publicsuffix/list/pull/822#discussion_r289142418 , ICANN has switched to providing the gTLD list via JSON at https://www.icann.org/resources/registries/gtlds/v2/gtlds.json and the CSV list at https://www.icann.org/resources/registries/gtlds/v1/newgtlds.csv

We should update the[`newgtlds` script](https://github.com/publicsuffix/list/blob/master/tools/newgtlds) to use the new feeds.

cpu:
I wrote a small Go tool that uses the v2 JSON feed for [`zlint`](https://github.com/zmap/zlint) called [`zlint-gtld-update`](https://github.com/zmap/zlint/blob/master/cmd/zlint-gtld-update/main.go)

On top of that I have an [automation repository](https://github.com/cpu/zlint-autopull) (based on athe approach @weppos uses with [publicsuffix-go](https://github.com/weppos/publicsuffix-go)) configured with a once-a-day Travis CI cron build to make automatic PRs when TLD data changes.

My Python skills aren't as refined as my Go skills but I think with some review help I can arrive at a solution that works for you folks if it would be helpful! I would probably have time to start on this late next week.

weppos:
Thanks for the offering @cpu. You know I'm an happy Go dev too, so I would be totally fine with using Go. Python was used in the past because previous contributors/maintainers (notably @gerv) were familiar with that.

I am actually not using Python at all, I believe the last time I've used it was to adjust the code in this repo. 😅

cpu:
>  I would be totally fine with using Go. 

That would certainly make my life easier :+1: :rocket:

weppos:
@sleevi any preference?

sleevi:
I’m perfectly happy with Go :)

cpu:
Great! 

Would you folks prefer to see the existing tool ported to Go using the same deprecated data source in one PR, and then separately updated for the new JSON data source in a follow-up or is it acceptable to do both in one fell swoop?

sleevi:
One fell swoop is fine

cpu:
I started working on this today. Hope to have a PR up for review shortly.

cpu:
A first version is ready for review: https://github.com/publicsuffix/list/pull/834

PR #822, opened 2019-05-24
Update new gTLDs to 2019-05-24

sleevi:

PR #834, opened 2019-06-13
Update newgtlds tooling & associated gTLD data.

cpu:
This branch resolves https://github.com/publicsuffix/list/issues/831 by replacing the `newgtlds` utility written in Python with a `newgtlds.go` utility written in Go that uses ICANN's updated V2 JSON gTLD registry.

This branch also replaces https://github.com/publicsuffix/list/pull/822 by running the updated tooling to refresh the gTLD data in `public_suffix_list.dat`.

If you'd prefer to see the tooling merged first and the data update done in a follow-up PR I'd be happy to split them. I thought it might be useful to see a proposed diff generated by the new tooling when reviewing the tooling so I've included the update in this branch to start.

I included some unit tests for the updated `newgtlds.go` code. They can be run with `go test -v tools/*.go`. If the repo maintainers are interested I would be happy to update the `.travis.yml` to run these unit tests as part of the regular CI process.

The existing `patchnewgtlds` script was using `/bin/sh` - I really wanted to add `set -e` and `set -o pipefail` to ensure that if one of the commands in the patch pipeline fail the script fails. I wasn't 100% sure if these `set` options were portable so I changed this script to use `/bin/bash`. If you'd prefer not to see that change in shell I'm happy to oblige. On a related note, the `tools/` directory now has: Go (`newgtlds.go`), BASH (`patchnewgtlds`), and Perl (`replace-between`). Replacing more of the `tools/` with Go code could be good follow-up work.


cpu:
It looks like my data update in 3ca081a is failing the `doublette/ambiguity` tests [in CI](https://travis-ci.org/publicsuffix/list/builds/545480761). I'm out of time for tonight but I'll aim to iterate on that in the next few days. Advice welcome if you spot the bug before me :-)

sleevi:
Thanks for doing this!

> Advice welcome if you spot the bug before me :-)

You can see duplicates like `com` and `biz`. It appears to be because the JSON is pulling all TLDs, and not just the 'new' gTLDs.

I see a couple options here, but there may be more:
- Restructure the PSLs `ICANN` section to separate out ccTLDs (which will presumably never be within the gtlds.json list) from gTLDs, and merge 'new gTLDs' and 'gTLDs' within the PSL into an identical section
- Load the PSL, and don't include entries in the gtlds.json list if they're already present and outside of the newGTLDs section
- Allowlist/blocklist certain 'known TLDs' in the script. I suspect this will predominantly/exclusively be the gTLDs that existed prior to the ICANN new GTLD program (such as `com`, `biz`, `aero`, etc)

I think I'm slightly preferential to the latter two options, because AIUI, some of the domains (and `aero` is a particular example) may have registry-reserved 2LDs that the PSL already tries to capture, and thus trying to wholly replace `aero` using the JSON would potentially remove the record of that structure.

Does that help?

cpu:
> Thanks for doing this!

Ofc! 

> You can see duplicates like com and biz. It appears to be because the JSON is pulling all TLDs, and not just the 'new' gTLDs.

Ahhh, I see what you mean. I didn't realize the old `.csv` differed here. Thanks for calling that out for me.

> I think I'm slightly preferential to the latter two options, because AIUI, some of the domains (and aero is a particular example) may have registry-reserved 2LDs that the PSL already tries to capture, and thus trying to wholly replace aero using the JSON would potentially remove the record of that structure.
>
> Does that help?

Yes, thank you! I'll take a look at the latter two options and see which looks like it will be the best fit.

weppos:
This is definitely outside the scope of this change, but since @sleevi mentioned potential changes to the list, I actually wonder: what would happen if a registry requires extra suffixes to be added to the list?

For instance, there is `foo` and for some reason we need to add `gov.foo`. Since the list is auto-generated, we are potentially screwed.

This came to my mind because I was wondering why we need to separate newGTLD and we can't for instance pull all ICANN TLDs directly. And then I realized we have second/third level suffixes for other TLDs.

sleevi:
On Fri, Jun 14, 2019 at 4:58 PM Simone Carletti <notifications@github.com>
wrote:

> For instance, there is foo and for some reason we need to add gov.foo.
> Since the list is auto-generated, we are potentially screwed.
>

I think we’d be fine. Today, for no other reason than it makes it easier
for us to maintain the file, and because of how it originated, we group
secondary and tertiary registry-reserved domains with the ICANN-assigned
TLDs.

For simplicity sake, we dropped the CSV into this file, in a fixed marker.
There’s no inherent need for that either.

This is similar to how we sort entries within the PRIVATE section - they
help us maintain them, but they don’t serve a semantic purpose.

Since there were only 27 or so TLDs prior to the new gTLD program, I was
trying to offer small solutions that don’t boil the ocean. At the end of
the day, all of the structure that exists is just to make it easier for us
as maintainers, and if things need to change, I’m sure we can come up with
ideas :)

>

cpu:
> Allowlist/blocklist certain 'known TLDs' in the script. I suspect this will predominantly/exclusively be the gTLDs that existed prior to the ICANN new GTLD program (such as com, biz, aero, etc)

This was the solution I ended up going with in 82fde5b. The number of legacy gTLDs was small enough that it made more sense to me to handle them specially than to add the complexity of having to parse the PSL to this tool.

I regenerated the data with the fix from 82fde5b in 5a21a3f and all of the tests are passing now :tada:

I did a little bit of unrelated cleanup while I was at it:
* 82d8fbe - consolidates the filtering logic
* 0c571fd - formats the fetch date better (and 9a13950 updates the `.dat`)
* 34c4259 - adds a few more tests

PTAL when you folks have a chance. Thanks!

tools/newgtlds.go
```diff
@@ -0,0 +1,254 @@
...
+	}
+
+	// pslTemplate is a parsed text/template instance for rendering a list of pslEntry
+	// objects in the format used by the public suffix list.
+	//
+	// It expects the following template data:
+	//   URL - the string URL that the data was fetched from.
+	//   Date - the time.Date that the data was fetched.
+	//   Entries - a list of pslEntry objects.
+	pslTemplate = template.Must(template.New("public-suffix-list-gtlds").Parse(`
```
cpu: There might be an argument to be made that this template is simple enough that it would be better to replace it with string manip/`fmt` usage similar to the Python version.

I already had code based around templating so I went this direction. Happy to change if folks would prefer.

sleevi: I'm fine with this. Granted, I don't do much Go readability :)

tools/newgtlds.go
```diff
@@ -0,0 +1,254 @@
...
+{{- end }}
+`))
+)
+
+// pslEntry is a struct matching a subset of the gTLD data fields present in
+// each object entry of the "GLTDs" array from ICANN_GTLD_JSON_URL.
+type pslEntry struct {
+	// GTLD contains the gTLD name. For internationalized gTLDs the GTLD field is
+	// expressed in punycode.
+	GTLD string
```
sleevi: nit: For consistency with ULabel, should this be called ALabel? I'm fine with either, but the latter seems slightly more descriptive of format - but at the cost of being less-descriptive of content.

tools/newgtlds.go
```diff
@@ -0,0 +1,254 @@
...
+	if e.ULabel == "" {
+		e.ULabel = e.GTLD
+	}
+}
+
+// Comment generates a comment string for the pslEntry. This string has a `//`
+// prefix and includes:
+//   * the ASCII representation of the gTLD (punycode for internationalized
+//     gTLDs).
+//   * the date of the contract signature (may be empty).
```
sleevi: The description here says "may be empty", but then it's listed as unconditionally included.

The reason this reads a bit weird to me is that, by the same logic, RegistryOperator is also always present, but may be empty as well. That is, it seems the three forms are:

```
<gTLD> :
<gTLD> : <Contract Signature Date>
<gTLD> : <Contract Signature Date> <RegistryOperator>
```

I suppose the only distinction is that, without the RegistryOperator check, we could end up with two trailing spaces instead of one, right?

cpu:
>  I was thinking we may want to Squash-and-Merge here. WDYT?

That would be my pref :-) I'm also happy to squash my commits within this branch.

tools/newgtlds.go
cpu: TBH I was trying to exactly match the logic from the previous tool to avoid output diffs without putting much thought into the "why". I think my textual comment probably misrepresented the result.

https://github.com/publicsuffix/list/blob/6f03f42a65d006c8ae657f125f14fb8f9d3337f4/tools/newgtlds#L38-L40

> I suppose the only distinction is that, without the RegistryOperator check, we could end up with two trailing spaces instead of one, right?

Yes, I think that's what the original code was trying to avoid. I'll update my comment to try and make that clearer.

tools/newgtlds.go
cpu: The consistency argument makes sense to me. Implemented in ae671bb

tools/newgtlds.go
cpu: I took a crack at writing a better comment in 9ef22c3 - I think it better represents the possible output now.

cpu:
@sleevi @weppos Do you folks want to give @dnsguru another week to give this a :mag: or can we merge with two positive reviews?

sleevi:
@cpu Thanks again! I just went ahead and squash-and-merged it :)

cpu:
@sleevi Perfect, thanks!

dnsguru:
Giving this a look over

PR #842, opened 2019-06-27
Add lodz.pl domain

dszymczuk:

Description of Organization
====

Organization Website: https://www.man.lodz.pl/man/

Reason for PSL Inclusion
====

Domain `lodz.pl` exists on https://en.wikipedia.org/wiki/.pl#Regional_domains list, but it is not on public suffix list like another Polish regional domain.
Can't add domain to Cloudflare. They are not recognized as domain.

DNS Verification via dig
=======

```
dig +short TXT katedra.lodz.pl
"https://github.com/publicsuffix/list/pull/842"
```

```
dig +short ns lodz.pl
"https://github.com/publicsuffix/list/pull/842"
```

make test
====
```
# TOTAL: 5
# PASS:  5
# SKIP:  0
# XFAIL: 0
# FAIL:  0
# XPASS: 0
# ERROR: 0
```

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -5680,6 +5680,7 @@ lebork.pl
 legnica.pl
 lezajsk.pl
 limanowa.pl
+lodz.pl
 lomza.pl
 lowicz.pl
 lubin.pl


sleevi:
I'm not sure what DNS verification commands you're doing, or if you're editing by hand, but those are clearly not correct :)

For Registry-defined domain namespaces, we require the request either be opened by the registry or be supported by official registry documentation.

Working from IANA, we can see the registry information is defined at  https://www.iana.org/domains/root/db/pl.html , and points to https://www.dns.pl/en/

I have trouble finding a clear Registry-defined policy with respect to a reservation. The closest I can see is https://www.dns.pl/en/registration_domain_pl/technical_conditions , which seems to support using https://www.dns.pl/en/list_of_regional_domain_names as a reserved list (and not simply using https://www.dns.pl/en/list_of_functional_domain_names )

From that list, I do not see `lodz.pl` on either. Indeed, it appears to be a separate registration, to `POLITECHNIKA ŁÓDZKA CENTRUM KOMPUTEROWE` based on WHOIS.

As such, this is a private registration, not a registry-defined registration, and would need to be updated and with the appropriate DNS verification.

I'm closing this as WONTFIX, because this PR is not that.

PR #843, opened 2019-07-01
Removed k12.wa.us

danheisel:
Hi. So this is certainly a new one for me. My team launched a new site for k12.wa.us and our host is unable to apply an automated public cert for the site. They pointed us here and asked that we remove k12.wa.us from the list so the cert can apply. I've made the change and hope that nothing is out of order. Please be kind if this little change is not correct. Thanks!

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -6506,7 +6506,7 @@ k12.ut.us
 k12.vi.us
 k12.vt.us
 k12.va.us
-k12.wa.us
+// k12.wa.us - Removed at the request of Dan Heisel <dan.heisel@watech.wa.gov>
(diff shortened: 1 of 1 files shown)


weppos:
@danheisel the change can only be requested by a representative of the k12.wa.us suffix. Do you represent the suffix?

Also, removing this suffix from the list will have security implications related, for example, to the ability to set cookies from outside k12.wa.us. Please only proceed if you are absolutely aware of the security impact.

danheisel:
@weppos Thanks! We're going to hold off for now then. I'd like to cancel this pull request as we explore another path before taking a chance with this change. Thanks a bunch for your help.

PR #847, opened 2019-07-15
Add *.webhare.dev

unilynx:

Description of Organization
====

Organization Website: https://www.webhare.com/

We are WebHare bv; we develop a CMS named WebHare which can be self-installed or is offered as a SaaS. We are a commercial organization but working on open-sourcing (parts of) our CMS - which will be done through https://www.webhare.dev/

The reason we're requesting the PSLs for webhare.dev and not webhare.com is that we're trying to separate the open source/product bits and the company bits.

Reason for PSL Inclusion
====

We want to use subdomains under `test.webhare.dev` for continuous integration tests (we have a few automated tests that specifically test cross-domain logins, and require separate cookie jars) and for test and acceptance versions of websites for ourselves and our customers (which require logins to be separated for the sites).

We want to offer subdomains under `user.webhare.dev` to developers and partners who use our CMS to set up their own (possibly publicly accessible) test and development instances and to be able to set up DNS-validated (single or wildcard) SSL certificates for them. If we have a 'safe' subdomain to allocate to users, we can automate the entire provisioning process for them.

Adding `*.webhare.dev` allows both and keeps the PSL a slight bit smaller

DNS Verification via dig
=======

```
Arnolds-MBP:list arnold$ dig -t txt _psl.webhare.dev +short
"https://github.com/publicsuffix/list/pull/847"

```

make test
=========

I ran the test - https://pastebin.com/zmQXR0vZ

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -12913,6 +12913,10 @@ voorloper.cloud
 // Submitted by Masayuki Note <masa@blade.wafflecell.com>
 wafflecell.com
 
+// WebHare bv: https://www.webhare.com/
+// Submitted by Arnold Hendriks <info@webhare.com>
+*.webhare.dev
+
 // WeDeploy by Liferay, Inc. : https://www.wedeploy.com
 // Submitted by Henrique Vicente <security@wedeploy.com>
 wedeploy.io


sleevi:
We're keen to avoid adding testing domains. I'm curious with your domain setup, whether `*.webhare.dev` would be a viable rule, and would mark every subdomain of `webhare.dev` as an effective public suffix. Would that work, as it would help clients significantly in saving space and processing?

unilynx:
`*.webhare.dev` would work as well. I'll update the PR

sleevi:
Thanks! Apologies, I looked at webhare.com and didn't realize you were also hosting webhare.dev. I didn't spot anything that would be a compatibility risk, as long as you're comfortable with `www.webhare.dev` being treated as itself a public suffix. Sorry for not having noticed that when making the suggestion.

If that works for you, this LG. The only thing missing (and again, apologies as this was due to my suggestion) is the TXT record for `_psl.webhare.dev`. I already verified for user & test, but would need to verify for the apex with this change.

Thanks for understanding, and thanks for helping keeping the bloat of the PSL in check.

unilynx:
sorry, got distracted by a squirrel on my way to the DNS settings

```
Arnolds-MBP:list arnold$ dig -t txt _psl.webhare.dev +short
"https://github.com/publicsuffix/list/pull/847"
```

PR #824, opened 2019-05-25
Remove zone.id (!urgent)

hdrxs312:
this domain will be expired.
reason:
- the new owner may dont know about publicsuffix. it will be dangerous.
- keep the list from inactive domain, if there is no one buy this domain

* [x] Description of Organization
* [x] Reason for PSL **Removal**
* [x] DNS verification via dig
* [x] Run Syntax Checker (make test)

Description of Organization
====

Organization Website: https://www.zone.id

Reason for PSL _Removal_
====

DNS Verification via dig
=======

make test
=========

PR #575, opened 2017-12-06
Add zone.id

hdrxs312:
My website https://zone.id is a subdomain provider, which let someone to register a subdomain and park it to their website / blog.
I am worrying about cookie stealing throug XSS (Cross Site scripting) or other method.
Please help me.

PR #874, opened 2019-09-04
Remove zone.id from list

hdrxs312:
Domain is no longer accessible (expired).
===

You can remove from the list.
Thankyou.

.

* [ ] Reason for PSL Inclusion

Description of Organization
====

Organization Website: 

Reason for PSL Inclusion
====

DNS Verification via dig
=======

make test
=========

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -12995,8 +12995,4 @@ virtualserver.io
 site.builder.nu
 enterprisecloud.nu
 
-// Zone.id : https://zone.id/
-// Submitted by Su Hendro <admin@zone.id>
-zone.id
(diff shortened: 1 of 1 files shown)


sleevi:
Removing on the basis that requestor is same user as requested in https://github.com/publicsuffix/list/pull/575

@dnsguru @weppos We don't really have a good way to validate removal requests (e.g. see past issue https://github.com/publicsuffix/list/pull/824 ). The domain is still registered, so I can see an argument going either way. I'm most worried about using someone's prior authorization in the event they sell or transfer the domain to a new entity, but we don't have a good way to resolve this.

PR #779, opened 2019-03-01
Add on-rio.io and on-rancher.cloud

vincent99:
* [x] Description of Organization
* [x] Reason for PSL Inclusion
* [x] DNS verification via dig
* [x] Run Syntax Checker (make test)

Description of Organization
====

Organization Website: https://rancher.com

Rancher Labs develops an open-source Kubernetes distribution and management platform called Rancher and similar container product called Rio.

I am [Vincent Fiduccia](https://github.com/vincent99), software architect at Rancher.

Reason for PSL Inclusion
====

We are adding a feature to automatically provide each user cluster with a `<service name>.<random user-id>.on-rancher.cloud` (and `on-rio.io`) DNS entry, so each `user-id` is a mutually untrusting third-party that should not be able to set cookies readable by each other.  We also plan to provide Let's Encrypt wildcard certs for each `user-id`.

DNS Verification via dig
=======

```
dig +short TXT _psl.on-rancher.cloud
"https://github.com/publicsuffix/list/pull/779"
```

```
dig +short TXT _psl.on-rio.io
"https://github.com/publicsuffix/list/pull/779"
```
make test
=========

Ran and passed..

<details>
cd linter;                                \
  ./pslint_selftest.sh;                     \
  ./pslint.py ../public_suffix_list.dat;
test_NFKC: FAILED
--- test_NFKC.expected	2019-03-01 18:29:20.000000000 +0000
+++ log/test_NFKC.log	2019-03-01 18:59:38.000000000 +0000
@@ -1,2 +1,10 @@
-9: error: Rule must be NFKC: 'südtirol.it'
-11: warning: No PRIVATE section found
+Traceback (most recent call last):
+  File "./pslint.py", line 276, in <module>
+    sys.exit(main())
+  File "./pslint.py", line 270, in main
+    lint_psl(infile)
+  File "./pslint.py", line 174, in lint_psl
+    error('Rule must be NFKC')
+  File "./pslint.py", line 44, in error
+    print('%d: error: %s%s' % (nline, msg, ": \'" + orig_line + "\'" if orig_line else ""))
+UnicodeEncodeError: 'ascii' codec can't encode character '\u0308' in position 32: ordinal not in range(128)
test_allowedchars: OK
test_dots: OK
test_duplicate: OK
test_exception: OK
test_punycode: OK
test_section1: OK
test_section2: OK
test_section3: OK
test_section4: OK
test_spaces: OK
test_wildcard: OK
[...]

PR #879, opened 2019-09-13
Update on-rancher.cloud

vincent99:

Description of Organization
====

Organization Website: https://rancher.com

Rancher Labs develops an open-source Kubernetes distribution and management platform called Rancher.

I am [Vincent Fiduccia](https://github.com/vincent99), software architect at Rancher.

Reason for PSL Inclusion
====

This entry already exists (#779), but we would like to add another "*." onto it so that it can be split into multiple DNS zones to control the number of records that will live in each one.

(There are no records in on-rancher.cloud yet so there is no concern for current records becoming public.  on-rio.io is in use and is not changing.)

We are adding a feature to automatically provide each user cluster with a `<service name>.<random user-id>.<shard>.on-rancher.cloud` DNS entry, so each `user-id` is a mutually untrusting third-party that should not be able to set cookies readable by each other.  We also plan to provide Let's Encrypt wildcard certs for each `user-id`.

DNS Verification via dig
=======

```
dig +short TXT _psl.on-rancher.cloud
"https://github.com/publicsuffix/list/pull/879"
```

make test
=========

Says `12591: error: Illegal character: '*.*.on-rancher.cloud'`, but the [specification](https://publicsuffix.org/list/) specifically says this should be allowed?

`(I.e. *.*.foo is a valid rule: *bar.foo is not.)`

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -12600,6 +12600,7 @@ rackmaze.net
 
 // Rancher Labs, Inc : https://rancher.com
 // Submitted by Vincent Fiduccia <domains@rancher.com>
+*.on-k3s.io
 *.on-rancher.cloud
 *.on-rio.io
 


vincent99:
The linter has a case that would specifically check and allow this, but no explanation of why it's commented out and replaced with an if that only allows one `*.` prefix.

https://github.com/publicsuffix/list/blob/master/linter/pslint.py#L182-L183

sleevi:
That’s because no known implementation supports that syntax.

PR #796, opened 2019-04-01
Add Handshake zones

mikedamm:
* [X] Description of Organization
* [X] Reason for PSL Inclusion
* [X] DNS verification via dig
* [X] Run Syntax Checker (make test)

Organization Website: https://handshake.org/

Handshake is a decentralized, permissionless naming protocol compatible with DNS where every peer is validating and in charge of managing the root zone with the goal of creating an alternative to existing Certificate Authorities. Its purpose is not to replace the DNS protocol, but to replace the root zone file and the root servers with a public commons. The Handshake protocol maintains the root zone file in a decentralized manner, making the root zone uncensorable, permissionless, and free of gatekeepers.

I am a developer for this project.

Reason for PSL Inclusion
====

Names registered in Handshake are reflected in to the hs.zone and hs.run zones, allowing clients not running Handshake to look up these names. Because multiple parties can indirectly register names under these zones, cookie isolation is highly desirable.

DNS Verification via dig
=======

```
bash-3.2$ dig +short IN TXT _psl.hs.zone.
"https://github.com/publicsuffix/list/issues/796"
bash-3.2$ dig +short IN TXT _psl.hs.run.
"https://github.com/publicsuffix/list/issues/796"
```

make test
=========

Tested. https://pastebin.com/raw/iHMUfmMJ

```
============================================================================
Testsuite summary for libpsl 0.20.2
============================================================================
# TOTAL: 5
# PASS:  5
# SKIP:  0
# XFAIL: 0
# FAIL:  0
# XPASS: 0
# ERROR: 0
============================================================================```

PR #884, opened 2019-09-18
Add additional Handshake resolver (hsd.zone)

jacobhaven:
* [x]  Description of Organization
* [x]  Reason for PSL Inclusion
* [x]  DNS verification via dig
* [x]  Run Syntax Checker (make test)

Organization Website: https://handshake.org/

Handshake is a decentralized, permissionless naming protocol compatible with DNS where every peer is validating and in charge of managing the root zone with the goal of creating an alternative to existing Certificate Authorities. Its purpose is not to replace the DNS protocol, but to replace the root zone file and the root servers with a public commons. The Handshake protocol maintains the root zone file in a decentralized manner, making the root zone uncensorable, permissionless, and free of gatekeepers.

I am a developer for this project.

# Reason for PSL Inclusion
Names registered in Handshake can be resolved as a prefix of the hsd.zone, allowing clients not running Handshake to look up these names. Because multiple parties can indirectly register names under these zones, cookie isolation is highly desirable.

hsd.zone is being added in addition to hs.zone / hs.run added in #796 for independent testing of different resolver configurations.

# DNS Verification via dig
```
$ dig +short IN TXT _psl.hsd.zone.
"https://github.com/publicsuffix/list/issues/884"

```

# make test
Tested. https://pastebin.com/raw/dmyjDjLv

```
# TOTAL: 5
# PASS:  5
# SKIP:  0
# XFAIL: 0
# FAIL:  0
# XPASS: 0
# ERROR: 0
Making check in msvc
```

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -11891,6 +11891,8 @@ conf.se
 // Submitted by Mike Damm <md@md.vc>
 hs.zone
 hs.run
+// Submitted by Jacob H. Haven <jacob@jhaven.me>
+hsd.zone
 
 // Hashbang : https://hashbang.sh
 hashbang.sh


sleevi:
Thanks. This seems to violate [ICP-3](https://www.icann.org/resources/pages/unique-authoritative-root-2012-02-25-en), which software using the PSL respects and expects.

This is why, for example, the .onion TLD was long rejected, until the IETF and IANA decided to reserve it.

It seems that this request should be rejected on similar grounds.

jacobhaven:
@sleevi I agree that including names from the Handshake alternate root would not work, but this is just a regular ICANN-root DNS gateway to that system for backwards compatibility (similar to tor2web), not a new TLD like `.onion`.

It's the same as `hs.zone`/`hs.run` which [were added (by you) in July](https://github.com/publicsuffix/list/pull/796/commits/c3e72f8880d2ef9cd3b936e7af9e14ce63c302bd). Has your opinion changed?

jacobhaven:
@sleevi Have you looked at this again? This is just an alternative to the existing `hs.zone` public suffix that will hopefully be more stable a usable.

PR #904, opened 2019-10-28
Add gov.scot to Public Suffix List

martiell:

Description of Organization
====

Organization Website: https://www.gov.scot

The Scottish Government is the devolved government for Scotland. There is no ccTLD for Scotland, but there is a geographic gTLD: `.scot`. The gov.scot domain is used by the Scottish Government and other public sector organisations in Scotland.

The lack of ccTLD for Scotland means that this proposed entry appears in an unusual place in the list for a government domain. From the commit message:

> This entry for gov.scot appears in the private domains section of the list.  This is in contrast to most `gov.*` domains in the list, which appear in the ICANN section of the list, near or alongside the relevant country code.  This is because the .scot domain is a (geographic) gTLD and the gTLD section of the list is auto-generated, so the entry for `gov.scot` cannot be added alongside the entry for `.scot`. Use of the private domains section is consistent with, for example, `service.gov.uk`.

Please see Section 4 of this "Gov.scot domains" publication for information about which organisations are eligible to register gov.scot domains:
https://www.gov.scot/publications/govscot-domains/

Registration of gov.scot domains is handled by JISC:

* Instructions:
https://community.jisc.ac.uk/library/network-and-technology-service-documentation/register-govscot
* Form: https://www.jisc.ac.uk/forms/govscot-domain-registration-form

My role is as a software engineer in the Digital directorate at the Scottish Government (possibly devops or infrastructure engineer, depending on who you ask). My responsibilities include the hosting of gov.scot and www.gov.scot web sites (the former redirecting to the latter), and coordinating DNS changes for both www.gov.scot and the gov.scot apex domain with JISC.

This change follows consultation with:

* the Technical Director in the Digital Directorate;
* the Information Security Officer in the Digital Directorate;
* Digital Communications in the Scottish Government;
* JISC; and
* Demys (who provide DNS hosting for many (all?) gov.scot domains).

Reason for PSL Inclusion
====

*Cookies*: We'd like to avoid super-cookies being set on the gov.scot domain. Subdomains of gov.scot are owned by other organisations and we have concerns about the security of subdomains sharing the same cookie namespace. In particular, because some public sector services will be handling sensitive user information. Our team is responsible for web hosting of the gov.scot apex domain, so I'm also keen that we don't inadvertently set any cookies that could cause issues with the behaviour of other sites.

*Consistency*: The PSL currently includes many entries of the form `gov.tld`, such as `gov.uk`. This change ensures that the gov.scot domain is handled by browsers in a way that is consistent with domains used by other governments. This helps us be more confident about security on the domain, and I understand may also have benefits browser UIs.

*Let's Encrypt*: We're aware that PSL maintainers are concerned about use of the PSL to circumvent rate limits for this service. We do use Let's Encrypt at the Scottish Government, but our usage is not such that we need increased rate limits. However, we are concerned that if another public sector organisation in Scotland had a bug in their Let's Encrypt automation, then it could cause problems for us as we currently share the same usage limits with that organisation. We believe this change ensures that Let's Encrypt service limits are applied as intended, rather than something that circumvents PSL policy.
 
DNS Verification via dig
=======

```
dig +short TXT _psl.gov.scot
"https://github.com/publicsuffix/list/pull/904"
```

make test
=========

```
# TOTAL: 5
# PASS:  5
# SKIP:  0
# XFAIL: 0
# FAIL:  0
# XPASS: 0
# ERROR: 0
```

👍

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -12668,6 +12668,10 @@ logoip.com
 // Submitted by Hanno Böck <hanno@schokokeks.org>
 schokokeks.net
 
+// Scottish Government: https://www.gov.scot
+// Submitted by Martin Ellis <martin.ellis@gov.scot>
+gov.scot
+
 // Scry Security : http://www.scrysec.com
 // Submitted by Shante Adam <shante@skyhat.io>
 scrysec.com


weppos:
Thanks @martiell . I was able to confirm the request.

I have only one concern, regarding the placement of this entry. Given it's a semi-official TLD, I would see it better placed in the main section after the `.scot`, similar to what we did for `.gov.uk` (which is also managed by Jisc).

@sleevi WDYT?

martiell:
Hi @weppos,

Thanks for taking a look at this.

I've addressed the unusual placement in the list in the pull request description (see the excerpt from the commit message and preceeding paragraph).

The .scot entry is in an auto-generated part of the list, so I assumed that trying to add gov.scot nearby will cause it to be removed from the list the next time the gTLDs are imported.

This is my understanding of the comment here: https://github.com/publicsuffix/list/blob/b566870aaebab56ebf12f153482335c769715707/public_suffix_list.dat#L7081-7082

```
// List of new gTLDs imported from https://www.icann.org/resources/registries/gtlds/v2/gtlds.json on 2019-10-29T17:00:12Z
// This list is auto-generated, don't edit it manually.
```

Is there a way to make gov.scot appear in the main section of the list, without it being later removed by an update to the gTLDs? As per my comment above, I don't see how it can be done in the same way as gov.uk, because uk is in the ccTLD section of the list (which isn't auto-generated).

Thanks
Martin

weppos:
> The .scot entry is in an auto-generated part of the list, so I assumed that trying to add gov.scot nearby will cause it to be removed from the list the next time the gTLDs are imported.

You are absolutely right. This is actually an issue I brought up with @sleevi a few weeks ago while he and @cpu worked on automating the update. We did not have cases so far. I guess we just found one.

martiell:
Sure. For what it's worth, it seems reasonable that the Welsh Government might want to add gov.wales and llyw.cymru to the PSL. But I think they'd have the same issue.

cpu:
> This is actually an issue I brought up with @sleevi a few weeks ago while he and @cpu worked on automating the update. We did not have cases so far. I guess we just found one.

The real world has a preference for bubbling up corner cases doesn't it? :-) I'm not sure what the best solution is to this particular case but I can turn around updates to the automation pretty quickly if folks have thoughts on the best way to tackle this.

sleevi:
> Thanks @martiell . I was able to confirm the request.
> 
> I have only one concern, regarding the placement of this entry. Given it's a semi-official TLD, I would see it better placed in the main section after the `.scot`, similar to what we did for `.gov.uk` (which is also managed by Jisc).
> 
> @sleevi WDYT?

Sorry for the delay in responding.

This isn't an IANA ccTLD, and I think that is and should remain the criteria for the ICANN section. So this naturally seems to fit with the PRIVATE section, the same as we do other restricted-use generic TLDs.

weppos:
@sleevi I wonder, given this statement, how is .GOV.UK different than .GOV.SCOT?

PR #913, opened 2019-11-06
Add regnumneapolitanum.com

daftano:

Description of Organization
====

Organization Website: https://daftano.com

I am the founder.
The domain will provide hosting and domain names to business and individuals related to the Neapolitan culture.

Reason for PSL Inclusion
====

**_Cookie Security_**
The main purpose is to allow each assignee to set its cookies securely and keep them separate from those of the main server (www|api)

The * is required because each assignee will belong to specific geographical regions, e.g.:

**Cookies may not be set for:**
- ac.regnumneapolitanum.com
- asia.regnumneapolitanum.com
- north.regnumneapolitanum.com

**Cookies may be set for:**
- regnumneapolitanum.com
- api.regnumneapolitanum.com
- www.regnumneapolitanum.com
- www.south.regnumneapolitanum.com
- business.ad.regnumneapolitanum.com managed by Alice
- business.xyz.regnumneapolitanum.com managed by Bob

DNS Verification via dig
=======

```
dig +short TXT _psl.regnumneapolitanum.com
"https://github.com/publicsuffix/list/pull/913"
```
make test
=========

All passed:

```
PASS: test-is-public-builtin
PASS: test-is-public
PASS: test-registrable-domain
PASS: test-is-public-all
# TOTAL: 5
# PASS:  5
# SKIP:  0
# XFAIL: 0
# FAIL:  0
# XPASS: 0
# ERROR: 0
```

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -11114,6 +11114,12 @@ cupcake.is
 cyon.link
 cyon.site
 
+// Daftano: https://www.daftano.com
+// Submitted by Davide Fiorentino lo Regio <public-suffix@daftano.com>
+*.regnumneapolitanum.com
+!www.regnumneapolitanum.com
+!api.regnumneapolitanum.com
+
 // Daplie, Inc : https://daplie.com
 // Submitted by AJ ONeal <aj@daplie.com>
 daplie.me


sleevi:
Note, as currently implemented by a number of user agents, the wildcard rule has the effect of saying that the 'base', `regnumneapolitanum.com`, is also a public suffix.

The negation rules - `!www.regnumneapolitanum.com` - simply means that subdomains of `www.regnumneapolitanum.com` are treated as part of the same public suffix; however, it will not allow cookies to be set for `www.regnumneapolitanum.com` because of the wildcard being seen as meaning that `regnumneapolitanum.com` is itself a public suffix (thus treating `www` and `api` as separate)

I don't know how well this explanation makes sense, but the TL;DR: is this won't do what you want, and there's currently no interoperable way to do what you want, unfortunately. Even though there ideally would be.

daftano:
Thank you for the useful note. 
So ideally it should have a third level domain (eg.: `*.area.regnumneapolitanum.com`) 
to share the cookies between `www` and `api` and separated for 
ac.area.regnumneapolitanum.com
asia.area.regnumneapolitanum.com
south.area.regnumneapolitanum.com

would that work, or there is any other implementation surprise?

sleevi:
Correct, if you wish to ensure you have functional second levels, then assigning at the third-level (or, as you propose, fourth level) is generally better.

dnsguru:
This request appears to need revision to the PR, if it is still being requested.  @daftano please update/respond

daftano:
We’re still deciding what would be the best approach. I closed this request until a decision is made.

Thank you for your support.

PR #915, opened 2019-11-10
add localzone.xyz

yahesh:

Description of Organization
====

Organization Website: https://yahe.sh

**localzone.xyz** is meant as a domain to be used in internal networks to circumvent the problems layed out [by the ICANN](https://www.icann.org/en/system/files/files/name-collision-mitigation-05dec13-en.pdf). To date there is only the **home.arpa** domain as specified in [RFC 8375](https://tools.ietf.org/html/rfc8375) but which is not widely used.

Reason for PSL Inclusion
====

The reason for inclusion is Cookie Security.

DNS Verification via dig
=======

```
yahe@YaheBookAir:~$ dig +short TXT _psl.localzone.xyz
"https://github.com/publicsuffix/list/pull/915"
```

make test
=========

Executed `make test` successfully:
https://gist.github.com/yahesh/cd9e116b2a7af577cf796fe4f7cc485c

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -12121,6 +12121,10 @@ nodebalancer.linode.com
 // Submitted by Victor Velchev <admin@liquidnetlimited.com>
 we.bs
 
+// localzone.xyz
+// Submitted by Kenny Niehage <hello@yahe.sh>
+localzone.xyz
+
 // Log'in Line : https://www.loginline.com/
 // Submitted by Rémi Mach <remi.mach@loginline.com>
 loginline.app


dnsguru:
@yahesh can you please explain the rationale a bit better so it is documented here for future eyes?

yahesh:
@dnsguru The rationale behind this is to have a domain that can be used a local domain suffix within different organizations that will not collide with publicly available services, especially for organizations that do not intend to have a separate internal domain of their own and that do not wish to use their public domain internally.

dnsguru (approved):
* DNS Validates
* Rationale (expanded) sufficient
* passes tests
* no conflict with codebase

**APPROVED**

dnsguru:
@yahesh MERGED

PR #945, opened 2020-01-24
Add sorrell.us to list of private domains

SvenAelterman:

Description of Organization
====

Organization Website: https://sorrell.us. We currently do not have a public-facing website.

Sorrell Solutions, L.L.C. is a business service firm. 

I am the Cloud Software Architect for the projects, and secretary of the LLC.

Reason for PSL Inclusion
====

We create subdomains for different customer projects. Cookies from different projects should not leak up to the parent domain or other child domains.

DNS Verification via dig
=======

Pending PR #

make test
=========

Haven't been able to complete yet.

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -12768,6 +12768,10 @@ static.land
 dev.static.land
 sites.static.land
 
+// Sorrell Solutions, L.L.C. : https://sorrell.us (no current website)
+// Submitted by Sven Aelterman <security@sorrell.us>
+sorrell.us
+
 // SourceLair PC : https://www.sourcelair.com
 // Submitted by Antonis Kalipetis <akalipetis@sourcelair.com>
 apps.lair.io


SvenAelterman:
DNS verification:
````
dig +short TXT _psl.sorrell.us
"https://github.com/publicsuffix/list/pull/945"

weppos:
@SvenAelterman I see you (plan to?) host the main site on the same domain you are proposing as a suffix. Adding sorrell.us will likely prevent you from being able to use the naked domain in the future.

Do you confirm you want to proceed?

SvenAelterman:
@weppos I did not expect that. I am trying to find some resources as to the exact implications. It appears that there are two, based on https://wiki.mozilla.org/Public_Suffix_List/Uses:

1. Chrome will reject wildcard certificates (*.foo.bar) if foo.bar is a Public Suffix. 
2. Issuance of wildcard TLS certifications.

However, the caveat mentioned is that this is ignored for entries in the PRIVATE section, which is where our domain would be.

Can you help me understand what I am missing?

weppos:
> @weppos I did not expect that. I am trying to find some resources as to the exact implications. It appears that there are two, based on https://wiki.mozilla.org/Public_Suffix_List/Uses:
> 
> 1. Chrome will reject wildcard certificates (*.foo.bar) if foo.bar is a Public Suffix.
> 2. Issuance of wildcard TLS certifications.
> 
> However, the caveat mentioned is that this is ignored for entries in the PRIVATE section, which is where our domain would be.
> 
> Can you help me understand what I am missing?

That document refers to Mozilla behaviors, doesn't necessarily apply to other consumers. That's not binding.

Other consumers may handle the PRIVATE section in the same way of the ICANN one. In fact, we've seen that happening in the past.

dnsguru:
I am sweeping through the back-log of issues and Pull Requests and would like to close this as it the patch, as-is **fails** testing - @SvenAelterman can you re-submit if you still need this

SvenAelterman:
I will re-evaluate. We want to be able to use the naked domain, but we also don't want to publish some of the subdomains on this list.

PR #948, opened 2020-01-28
Add envoy.rocks domain

Rijk:

Description of Organization
====

Organization Website: https://www.envoy.rocks

Envoy is an online feedback tool for creative agencies. It is a SaaS platform where each agency can set up a branded instance for their team under a subdomain.

I am the owner and developer of the platform.

Reason for PSL Inclusion
====

The different subdomains (agency1.envoy.rocks, agency2.envoy.rocks, etc) should not share cookies. Also, for people using the tool to give feedback it is important that the complete domain is highlighted in the address bar, as they will mainly recognise the name of the agency which is the subdomain.

DNS Verification via dig
=======

```
dig +short TXT _psl.envoy.rocks
"https://github.com/publicsuffix/list/issues/948"
```

make test
=========

Tests run.

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -11524,6 +11524,10 @@ staging.onred.one
 enonic.io
 customer.enonic.io
 
+// Envoy : https://www.envoy.rocks/
+// Submitted by Rijk van Wel <rijk@envoy.rocks>
+envoy.rocks
+
 // EU.org https://eu.org/
 // Submitted by Pierre Beyssac <hostmaster@eu.org>
 eu.org


weppos:
@Rijk I see you host the main site on the same domain you are proposing as a suffix. While you correctly use `www` hence you are currently not affected, I just want to warn you before proceed. Adding `envoy.rocks` will likely prevent you from being able to use the naked domain in the future.

Do you confirm you want to proceed?

Rijk:
Thanks for the heads up Simone. Do you mean browsers will no longer accept the naked domain? Right now, when you navigate to https://envoy.rocks, you get a 301 to https://www.envoy.rocks/. Will this still work after adding?

weppos:
It will. But consumers of the PSL are not just browers. As an example, since `envoy.rocks` will become a suffix, some CAs may refuse to issue a SSL certificate (which will prevent you from getting a certs for the apex and hence perform HTTPS-redirects in the future, effectively breaking `https://envoy.rocks`).

Rijk:
Thanks for pointing this out! Knowing this, I think it’s best to cancel this request and leave it as is.

PR #964, opened 2020-02-11
Update fj entry

dnsguru:
Registry .fj provided direct list of stub zones after being informed by CA that registry could not apply for SSL due to presence of *.fj for pending upgrade to registry system

The registry also supplied the explicit list of direct zones
fj
ac.fj
biz.fj
com.fj
gov.fj
info.fj
mil.fj
name.fj 
net.fj
org.fj
pro.fj

Description of Organization
====

Organization Website: https://domains.fj

Reason for PSL Inclusion
====

Already listed, this is the ccTLD for Fiji, replacing the *.fj listing with the list of tld and direct sld stubs within .fj, furnished from the registry operator

DNS Verification via dig
=======
_psl TXT records returning pull request ID correctly and verified 

make test
=========

Tested ok

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -982,7 +982,7 @@ fi
 // TODO: Check for updates (expected to be phased out around Q1/2009)
 aland.fi
 
-// fj : Updated 2020-02-11 
+// fj : Updated 2020-02-11  by domains.fj
 // Confirmed by registry garth.miller@cocca.org.nz
 fj
 com.fj
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -985,16 +985,16 @@ aland.fi
 // fj : http://domains.fj/
 // Submitted by registry <garth.miller@cocca.org.nz> 2020-02-11
 fj
+ac.fj
+biz.fj
 com.fj
+gov.fj
+info.fj
+mil.fj
+name.fj 
 net.fj
 org.fj
-ac.fj
-info.fj
 pro.fj
-biz.fj
-name.fj 
-mil.fj
-gov.fj
 
 // fk : https://en.wikipedia.org/wiki/.fk
 *.fk
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -982,8 +982,19 @@ fi
 // TODO: Check for updates (expected to be phased out around Q1/2009)
 aland.fi
 
-// fj : https://en.wikipedia.org/wiki/.fj
-*.fj
+// fj : Updated 2020-02-11 
+// Confirmed by registry garth.miller@cocca.org.nz
+fj
+com.fj
+net.fj
+org.fj
+ac.fj
+info.fj
+pro.fj
+biz.fj
+name.fj 
+mil.fj
+gov.fj
 
 // fk : https://en.wikipedia.org/wiki/.fk
 *.fk
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -982,8 +982,8 @@ fi
 // TODO: Check for updates (expected to be phased out around Q1/2009)
 aland.fi
 
-// fj : Updated 2020-02-11  by domains.fj
-// Confirmed by registry garth.miller@cocca.org.nz
+// fj : http://domains.fj/
+// Submitted by registry <garth.miller@cocca.org.nz> 2020-02-11
 fj
 com.fj
 net.fj


dnsguru:
I have submitted this and have the ccTLD administrator adding the _psl entry to the .fj zone for validation with the pull request number

dnsguru:
@weppos because the .fj is above com.fj etc stubs, can we skip the _psl TXT records for those to save the admin the hassle?

dnsguru:
TXT Record confirmed

sleevi:
@dnsguru could you explain more about the problem trying to be solved? This is much less efficient, and so understanding the registry policies seem very useful here, as does a better understanding of the problem to solve.

dnsguru:
Test good

sleevi:
Based on http://domains.fj/public/faq.php#14 I think we may want to reject this. At the least, we should better understand the problem to see the best way to solve it.

dnsguru:
The *.fj is blocking the issuance of certs from Sectigo, Thawte, digicert, GoDaddy - they will just not issue any ssl for fj because of the existing entry.

dnsguru:
The registry system is being updated, and the new system requires an ssl cert, which can't be issued,  

They have put in place the correct record to validate the _psl authority here, for this and the other stubs.  I have known the person submitting it since 1995 they are reputable and are connected to the project.

public_suffix_list.dat
```diff
@@ -982,8 +982,19 @@ fi
 // TODO: Check for updates (expected to be phased out around Q1/2009)
 aland.fi
 
-// fj : https://en.wikipedia.org/wiki/.fj
-*.fj
+// fj : http://domains.fj/
+// Submitted by registry <garth.miller@cocca.org.nz> 2020-02-11
+fj
```
weppos: The sorting needs adjustment.
https://github.com/publicsuffix/list/wiki/Format

For instance, `ac.fj` needs to be before `com.fj`

sleevi:
Thanks!

The current approach has the benefit of both failing closed and accurately
reflecting the current stated policy as efficiently as possible, which is
certainly good for embedders of the PSL.

Broadly speaking, I don’t think we want to replace wildcards with more
specific entries willy-nilly. If there are limited exceptions, we’d likely
want to use the exception syntax. If the 2LDs are the exception, rather
than the rule, then your PR is definitely how we’d handle it.

However, if the issue is just a matter of obtaining certs, it’s potentially
something specific to a CA. Normally, a CA can issue such certificates
fine, if they do additional due diligence validating the requestor controls
the namespace (e.g. that the request is from the duly authorized registry).
That only works if the CA is willing to do that validation - e.g. by
hand/manually, which not all CAs are willing to do. Understanding more of
the context is useful to trying to understand the problem here.

If, for example, they were trying to get a cert for domains.fj, it might
make perfect sense to list, as exception rules, !domains.fj and any other
directly-registry-operated domains.

Overall, I really appreciate you reaching out to them. Certainly, I don’t
see any problems with the parent record validating any child records; after
all, that’s how DNS works. But I do want to make sure we solve the right
problem as efficiently as possible 😊

dnsguru:
Have triple-dipple checked with registry on the intentional change reflecting their desired outcome.

@weppos or @sleevi please squash / merge 

(Can submitter be one to merge, or did we establish submitter must != merger?)

weppos:
@dnsguru Ultimately, only the 3 of us currently have merge rights, hence it's an easy problem. So far both Ryan and myself came into cases where we were both submitter and merger, generally when there was implicit consensus (e.g. when pulling down updates) or no conflicting arguments.

In this particular circumstance, since Ryan raised some concerns, I'd like the two of you to come into an agreement about moving forward of this PR. I don't want to step in and click "merge", it feels to me a bit disrespectful regarding Ryan's comment.

However, I'll take the opportunity to share my 2 cents. I propose we move forward and merge this PR, given it is also backed up by an immediate registry need, with the condition to follow up with the relevant CAs as suggested by Ryan to try to mitigate the need of similar changes in the future. Ryan concern is valid, and if we can avoid the need to have each individual registry to come with its on PR for the same purpose, that would be better.

sleevi:
@weppos I’d like to request we hold off merging, until the relevant information can be shared here on this issue for public record.

That is, on this issue, we don’t have an answer why `!domains.fj` wouldn’t suffice. Normally, with PRs directly opened by the registry, it’s very easy to get these answers because the registry will reply on the bug. When one of us is acting as interlocutor for the registry, it’s harder, and that is why it makes it important to document here, on this issue, why the particular approach is used. This is the same we apply to any PR :)

dnsguru:
> @weppos I’d like to request we hold off merging, until the relevant information can be shared here on this issue for public record.
> 

@sleevi darn, I had seen your willy-nilly and raised you with a triple-dipple.  ;)

Hopefully the following addresses this:

> That is, on this issue, we don’t have an answer why `!domains.fj` wouldn’t suffice. 

OK.  I think I understand where we're disconnected...  The rationale is that although the initial trigger of the ccTLD's interest in revision of their PSL entry was motivated by being rejected by the CA under the rationale of the *.fj entry, in my discussion with them identified that the registry has  will be making it possible for third parties to register at the second level, as well as under the explicit list of stub level domains.  

As I read through all the information I passed through, I wonder if I had articulated that important factor clearly, so I suspect that you are helping me to solve the specific case of their SSL request being embargoed due to the PSL listing.  I assure you I have vetted this with the concerns you're raising, and the PR is accurate.

Adding `!domains.fj` (would also need `!registry.fj`) would be the tip of the iceburg for the various parties registering at the second level who might seek certs because it will be openly available.  It just isn't now, and their FAQ is not explicitly updated until they will, so they _currently_ state "[Q: Can I register a name directly under .FJ?  A: Not at the moment.](http://domains.fj/public/faq.php#14)", their request was to get things ready for the next state that is coming shortly.

> When one of us is acting as interlocutor for the registry, it’s harder, and that is why it makes it important to document here, on this issue, why the particular approach is used. This is the same we apply to any PR :)

Yes, I respect and understand this.  Rather than just use @weppos' answer which is the faster path, I am taking the time to respect and address your point and reticence in the open.

I still would like this to get merged and move us forward to the rest of the stack.

Hopefully I have addressed your concern

sleevi:
> in my discussion with them identified that the registry has will be making it possible for third parties to register at the second level, as well as under the explicit list of stub level domains.

Ah hah. Thanks. This is indeed the missing and important piece for why making the change _this_ way makes sense. Thanks for clarifying!

dnsguru:
@weppos Still pending your review, but I addressed the sorting issue you'd identified.  I'd like to proceed with merge, ok?

PR #980, opened 2020-02-25
Remove unnecessary trailing whitespace for name.fj

dkg:
this is a simple change, just normalizing the text.


sleevi:
Thanks! I’m not sure why C-I didn’t catch this...

PR #1000, opened 2020-03-21
add bitbridge.net to private section

abelii:
// bitbridge.net : Submitted by Craig Welch, abeliidev@gmail.com

Description of Organization
====

Organization Website: https://www.softlatch.net https://gitlab.com/cwsyncdev/SoftLatch/-/wikis/home

SoftLatch provides a way to self-host home automation and productivity apps - a simple way to install domoticz (https://www.domoticz.com/), a motion detecting webcam application, and a note taking application, with more planned.  

I'm the founder of the project and owner of the domain names, Craig Welch.

Reason for PSL Inclusion
====

The desire to be added to this list is to enable https://letsencrypt.org/ certificates to be vended for the self-hosted bridge applications for which dynamic dns services are offered on bitbridge.net (e.g. after signup for an account on softlatch.net, during installation of the bridge on macos, windows, or linux (raspberry pi included), the bridge can be linked to the site to easily give it a dns name, and it will also create a let's encrypt certificate.  This entry will allow multiple bridges to register and receive let's encrypt certificates (otherwise there is a limit at let's encrypt which will block users over a certain volume each week from successfully completing the challenge).

DNS Verification via dig
=======

```
$ dig +short -t txt _psl.bitbridge.net
"https://github.com/publicsuffix/list/pull/1000"
```

make test
=========

Looks to pass...
```
# TOTAL: 5
# PASS:  5
# SKIP:  0
# XFAIL: 0
# FAIL:  0
# XPASS: 0
# ERROR: 0

```

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -11185,6 +11185,9 @@ drud.us
 // Submitted by Richard Harper <richard@duckdns.org>
 duckdns.org
 
+// bitbridge.net : Submitted by Craig Welch, abeliidev@gmail.com
+bitbridge.net
+
 // dy.fi : http://dy.fi/
 // Submitted by Heikki Hannikainen <hessu@hes.iki.fi>
 dy.fi


dnsguru:
Scrolling to the bottom of [https://publicsuffix.org/learn/](https://publicsuffix.org/learn/)

"If you just need an exception from their rate limits, please do not request a change to the list, but instead use their form, linked from [their documentation](https://letsencrypt.org/docs/rate-limits/). **This is a faster way to achieve what you want.** "

 [Let's Encrypt Rate Limits](https://letsencrypt.org/docs/rate-limits/)

After reading that, understanding how this PSL PR would work, do you seek to proceed?

abelii:
Thank you for asking - because each of the entries is a subdomain off the main domain and I can't group them into multi-sub-domain certs (as the clients generate the cert and get it signed without my having any of the bits, all that happens centrally is the dns challenge response) it doesn't look like the kind of rate limit they will raise directly via their form submission would apply - this really does look like an open second level domain case (kind of like a duckdns.org, application specific, but still) so, afaik, this is the only/best way to make sure users don't hit limits that will block them.  Given that I would like to proceed.

dnsguru:
Tests passed, DNS validated, rationale accurate, confirmed impact and intent, no conflicts with base branch.  Merging

groundcat:
@bitsii Do you still need this entry in the PSL? Domain is sitting at a parking page. If the project has been decommissioned and is no longer associated with an active project as described in your original request, please confirm in #2064 that it can be safely removed from the PSL to avoid leaving unnecessary debris for PSL volunteers and future registrants to deal with (#1996). Thank you.

PR #1011, opened 2020-04-07
Added Wiki Link as info resource

dnsguru:
They're already here... let's get that wiki in their face in addition to the publicsuffix.org website URL


dnsguru:
Perfect ... the objective is to help educate

PR #1105, opened 2020-09-22
Add .tm.dz

romuald:
Used for trademarks registered in Algeria

Reason for PSL Inclusion
====

The .tm.dz zone seem have been around for quite some time, but is not included in the PSL

Source: http://www.nic.dz/images/pdf_nic/charte.pdf - page 3 (in French)

~~(However I'm unable to find any domain in the zone, which is not using DNSSEC so I cannot enumerate domains either)~~

I will also try to update the [Wikipedia page](https://en.wikipedia.org/wiki/.dz) (**edit:** done)

The zone was [removed from Wikipedia in 2017](https://en.wikipedia.org/w/index.php?title=.dz&diff=798042678&oldid=798041828), seemingly because the [English charter](http://www.nic.dz/en/images/pdf_nic/charte.pdf), not changed since 2012, did not include it

```
% dig +short SOA tm.dz
ns1.nic.dz. dnsmaster.nic.dz. 2020091700 3600 600 86400 3600
```

make test
=========

The tests seems to be OK

```
# TOTAL: 3
# PASS:  3
# SKIP:  0
# XFAIL: 0
# FAIL:  0
# XPASS: 0
# ERROR: 0

# TOTAL: 5
# PASS:  5
# SKIP:  0
# XFAIL: 0
# FAIL:  0
# XPASS: 0
# ERROR: 0

```

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -917,6 +917,7 @@ edu.dz
 asso.dz
 pol.dz
 art.dz
+tm.dz
 
 // ec : http://www.nic.ec/reg/paso1.asp
 // Submitted by registry <vabboud@nic.ec>
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -907,16 +907,17 @@ org.do
 sld.do
 web.do
 
-// dz : https://en.wikipedia.org/wiki/.dz
+// dz : http://www.nic.dz/images/pdf_nic/charte.pdf
 dz
+art.dz
+asso.dz
 com.dz
+edu.dz
+gov.dz
 org.dz
 net.dz
-gov.dz
-edu.dz
-asso.dz
 pol.dz
-art.dz
+soc.dz
 tm.dz
 
 // ec : http://www.nic.ec/reg/paso1.asp


romuald:
Looking at the Wikipedia article, I see that some other zones were removed but still exists and are somewhat updated (for example asso.dz serial is 2020060100)

I will try to list the zones that are missing and amend this PR

dnsguru:
@romuald changes like this require the txt record be added to the respective zone that contains reference to this PR in order to validate it is submitted by registry representative with authority over zone

romuald:
@dnsguru I'm very pessimistic that we could obtain this from this particular registry

Do you always require this? I though those kind of ccTLD zones were managed "manually" by checking the registry registrations rules (and past rules since some may now be closed to new registrations)

dnsguru:
The objective is to have the party with administrative control over zones be the ones making changes that will impact the domains within it.  

I did follow the path from IANA.ORG -> nic.dz to the reference document above, where it lists on page 3 in the final bullet point that the following are valid: .com.dz, .gov.dz, .org.dz, .asso.dz, .pol.dz, .edu.dz, .net.dz, tm.dz, .soc.dz

It would still be important to have the registry be the party making changes to the zone, which requires outreach and time when they are not the submitting party. 

Although early versions of the PSL took information from a variety of third party submissions, and those are 'evergreen', we do count on community submissions like yours to update things, as we are entirely volunteers.    ccTLD sections benefit from the legacy entries, but we have implemented the DNS validation so that it makes the registry actively involved in the changes that could disrupt their zones in a way that they were not aware.

sleevi:
@dnsguru Is this new? We've previously consistently told ccTLDs that publishing such documents, as they have, is sufficient proof.

https://github.com/publicsuffix/list/wiki/Guidelines#icann-domains still states "Changes here need to either come from a representative of the registry (authenticated in a similar manner to below) or __be from public sources such as a registry website.__"

dnsguru:
@sleevi outstanding, that will work as rationale to proceed

dnsguru (approved):
Alphabetized SLDs and revised and added soc.dz and tm.dz to align with published policies of the registry
http://www.nic.dz/images/pdf_nic/charte.pdf

romuald:
@dnsguru thanks!

Small warning, the `soc.dz` zone does not seem to exist at the moment (NXDOMAIN) (the other zones exist)

My best guess is that the definition is so narrow no one bothered to register one and the zone was never technically created

Issue #1027, opened 2020-05-04
Modify request template / documentation to include related historic Issue/PR #s


dnsguru:
In reviewing submissions that modify or update previous approved submissions, I am noticing that having the submitting party furnish the historic PR or Issue numbers provides the community volunteers a more efficient process of review.

The most recent was #1025 where I had to look up #730 that they were expanding.  If we update the template and documentation to request that the submitting party include prior references, including approved and declined PR and Issues, it will help things move more efficiently, and it seems like a simple ask.

sleevi:
Yeah, I can see no harm in asking, as long as we're not requiring. It's definitely useful to see how the example domains are changing (i.e. what doesn't work vs what they want to work)

dnsguru:
Addressed in #1116

PR #1111, opened 2020-09-30
Add wapblog's domain

fsodic:
* [x] Description of Organization
* [x] Reason for PSL Inclusion
* [x] DNS verification via dig
* [x] Run Syntax Checker (make test)

Description of Organization
====
Organization Website: https://www.wapblog.id

I'm Fajar Sodik from WapBlog Developer Team
WapBlog.ID is a platform for create a blog.

Reason for PSL Inclusion
====
WapBlog.ID is a platform blog builder who user can create a blog which our subdomain or their own domain. If they don't have a domain, we provide some excelent domain for them like wbq.me or wblog.id and many other.

It could be very usefull, if browser could correctly manage those subdomains (cookies, highlighting the subdomain, sorting by subdomain and not all the domain ). We want these services to be recognized as a known public suffix and wapblog.id services too.

DNS Verification via dig
=======
```
dig +short TXT _psl.bloger.id
"https://github.com/publicsuffix/list/pull/1111"
```

```
dig +short TXT _psl.fastblog.net
"https://github.com/publicsuffix/list/pull/1111"
```

```
dig +short TXT _psl.idnblogger.com
"https://github.com/publicsuffix/list/pull/1111"
```

```
dig +short TXT _psl.indowapblog.com
"https://github.com/publicsuffix/list/pull/1111"
```

```
dig +short TXT _psl.wblog.id
"https://github.com/publicsuffix/list/pull/1111"
```

```
dig +short TXT _psl.wbq.me
"https://github.com/publicsuffix/list/pull/1111"
```

make test
=========
```
============================================================================
Testsuite summary for libpsl 0.21.1
============================================================================
# TOTAL: 5
# PASS:  5
# SKIP:  0
# XFAIL: 0
# FAIL:  0
# XPASS: 0
# ERROR: 0
============================================================================
```

Issue #1109, opened 2020-09-29
Require that expiration be >2 years from submission date on domains submitted in Private section


dnsguru:
#1095 Submitted a large list of entries, which included an expired name, which led to additional overhead in the back and forth dialog and replaying the review of the DNS validation.

I would like to update the documentation (and the rules) to state that it is the responsibility of the requestor to ensure that any names submitted must have expiration dates at least 2 or more years from the date of submission of a PR, and that the names maintain at least 1-2 years' of registration term in them.  I would also like to have a consequence of failing this step and ongoing obligation to be that they can lose their entry for that domain and/or their entire section. 

We do not have a means (currently, as I author this) to review the expiry date(s) of the names in the PSL private section to remove them, so I'd like to suggest we at very least introduce a requirement of names in the private section having at least 2 years of registration length remaining.  It can be lightweight requirement, as an assertion / statement within the rationale "the domains included in this submission hold a term length that extends more than 2 years from today's date".

Because there is not a clear requirement in the documentation, we may be receiving strings that could expire while listed.  

The change and requirements would allow for us to remove entries through third party reports of expiries or should we include a periodic and automated sweeping audit that indicated names having insufficient expiry dates.

We need something like this to help evolve from the set and forget mode that we have submitters using currently.

sleevi:
@dnsguru I support this. I would even go further, to suggest something like 5 years, given the lifecycle.

dnsguru:
https://github.com/publicsuffix/list/wiki/Guidelines/734d6b3579a38479cad2029f29d6ef9401bf72d6#validation 
Added third bullet point

I want to update the rationale template as well to include the boiler plate statement that I put in the updated bullet point

dnsguru:
Here is the new bullet point:
```
We now require that domains submitted as private section entries have expiration dates more than 2 years beyond the submitting date of a PR. Please include a statement within the rationale that a] the domain name(s) submitted have at least 2 years of registration period left in them at the time of submitting the name, and b] include a commitment that you will maintain their registration in good standing with more than a year left in their term.
```

dnsguru:
Added Bolded text to wiki section [Please be thorough](https://github.com/publicsuffix/list/wiki/Guidelines#information-furnished-in-the-pr-template-be-thorough): (**bold text is new**)

[...]

PR #1116, opened 2020-10-07
Update pull_request_template.md to ask for related issue or PR numbers

dnsguru:
Added request for submitting party to include historic Issue # or PR # to address issue #1027

* [ ] Reason for PSL Inclusion

Description of Organization
====

Organization Website: 

Reason for PSL Inclusion
====

DNS Verification via dig
=======

make test
=========

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -42,6 +42,9 @@ Please tell us why your domain(s) should be listed in the PSL
 confirm that any private section names hold registration term
 longer than 2 years and shall maintain more than 1 year 
 term in order to remain listed.
+
+Please also include the numbers of any past Issue # or PR # 
+specifically related to this submission or section.
 -->
 
 DNS Verification via dig


dnsguru:
Fast-follow to #1109 in order to address #1027 (request for submitters to include historic Issue/PR)

fsodic:
@sleevi & @dnsguru 
I completely disagree. Because it means we have to register the domain for the next 2 years which may add to the costs that are incurred for 1 year of financial statements if from an organization that already has a legal entity.
It might seem okay if the organization is large.
But think again if these organizations only target the local market in their respective countries.
So I think it makes more sense when the domain registered is at least 2 years old from the date of the PR request and not 2 years left since the PR request.
So we know whether the domain is well controlled and not all domains can immediately ask for PR for this publicsuffix.

dnsguru:
The discussion was really amongst the core maintainers on how to address
the change as it is part of addressing expired or expiring names as a
general rule and not punitive response to the PR that included the expired
names.  That was more a catalyst to take a long-deferred action which was
to add the renewal term requirements.

The disagreement is noted.  It is understandable to seek to save money, but
the renewal terms serve to keep things efficient and protect the submitting
party while maintaining the integrity of this list.

If a name cannot be maintained with ample registration term, it really does
not belong on the list and should not get submitted.

What we often face is that submissions are 'set and forget' until a change
is needed from the submitting party, but mostly folks get what they want
and leave, and this causes some debris to clean up when renewals don't
occur.

Renewal expenses are what they are, and dodging the costs of maintaining
the submitted names pushes burden onto the volunteers of this project to
address the issues of expired names being on the list.  The objective is to
maintain quality and reliable listings and resourcing is such that dealing
with expired names (especially in new submissions) wastes volunteer time.

Volunteers are maintaining the public suffix list. Wasting their time is a
cost as well, which should not be considered to have a value of zero.
There is no entitlement to being listed that should be expected by any
party seeking entries within the private section, and requests should be
modest in nature and conform in respect of this being a voluntary global
resource.  We also see that the file size growing can introduce bandwidth
and scaling issues, which modesty of requests can help, but also
housekeeping of old records will aid in addressing.

Because the volunteer time gets wasted removing expired names or watching
for them, those maintaining the project candidly do not, but field reports
from new registrants of names.  Out of necessity we will at some point have
some form of automation that reviews expiry dates and make removals of
domains or sections.  These will depend upon expiry horizons beyond at
least a year, so the objective in entries having renewal obligations is
really a safety net for the submitting party, so they know the potential
consequences of non-renewal.

The objective is to maintain a quality list, through modest, efficient
means, and the renewal requirement aids in that objective.

fsodic:
#1111 my PR not response bro?

dnsguru:
Had to answer all this other distraction stuff first...

#lookingnow #volunteer #broke-a-f-due-to-pandemic
#volunteeringcompeteswithpaidgigs

dnsguru:
@sleevi please approve and s/m

sleevi (approved):

PR #1136, opened 2020-10-29
Removed obsoleted lists of second level domains and added URLs to authoritative domain list locations

repperille:
Description of Organization
====

Norid AS runs the registry for Norwegian domain names. All domains directly under .no are registered with us. In addition to processing applications we develop the domain name policy according to the needs of society. We are also responsible for the technical operation and development of associated services. We also handle other national tasks related to our core activities.

In order to develop and adapt the rules for allocating domain names, we co-operate with various stakeholders. We also take social responsibility by working to keep the domain name system reliable and robust, and to ensure that it is governed in an open and democratic way.

The service is regulated by the Domain Name Regulations and is supervised by the Norwegian Communications Authority (Nkom). Norid is not a public administrative agency, and the allocation of domain names is based on private law. We are a neutral party, and the business is run non-profit. Norid is a subsidiary in the Uninett Group.

Organization Website: https://www.norid.no

I am a senior systems developer at Norid AS.

Reason for PSL Inclusion
====

Our ccTLD (no) was already listed here. This PR is just an update where we removed the erroneous second level domains and added URLs to where the correctly updated lists can be found.

DNS Verification via dig
=======

N/A

make test
=========

...
Making clean in msvc
Making check in po
Making check in include
Making check in src
  CC       libpsl_la-psl.lo
  CC       libpsl_la-lookup_string_in_fixed_set.lo
  CCLD     libpsl.la
ar: `u' modifier ignored since `D' is the default (see `U')
Making check in tools
  CC       psl.o
  CCLD     psl
Making check in fuzz
  CC       libpsl_fuzzer.o
  CC       main.o
  CC       libpsl_load_fuzzer.o
  CC       libpsl_load_dafsa_fuzzer.o
  CCLD     libpsl_icu_fuzzer
  CCLD     libpsl_icu_load_fuzzer
  CCLD     libpsl_icu_load_dafsa_fuzzer
PASS: libpsl_icu_load_dafsa_fuzzer
PASS: libpsl_icu_fuzzer
PASS: libpsl_icu_load_fuzzer
# TOTAL: 3
# PASS:  3
# SKIP:  0
# XFAIL: 0
# FAIL:  0
# XPASS: 0
# ERROR: 0
Making check in tests
  CC       test-is-public.o
  CC       test-is-public-all.o
  CC       test-is-cookie-domain-acceptable.o
  CC       test-is-public-builtin.o
  CC       test-registrable-domain.o
  CCLD     test-is-cookie-domain-acceptable
  CCLD     test-is-public-builtin
  CCLD     test-is-public
  CCLD     test-is-public-all
  CCLD     test-registrable-domain
PASS: test-is-public-builtin
PASS: test-is-public
PASS: test-registrable-domain
PASS: test-is-public-all
# TOTAL: 5
# PASS:  5
# SKIP:  0
# XFAIL: 0
# FAIL:  0
# XPASS: 0
# ERROR: 0
Making check in msvc


sleevi:
> This PR is just an update where we removed the erroneous second level domains and added URLs to where the correctly updated lists can be found.

Could you clarify: Removal comes with great risk. It sounds like you're saying there are still second level domains, and so they all need to be kept here (e.g. the category second-level domains).

The updated URLs are very valuable, but I think we still want to make sure we're judicious in removal. Could you explain more?

dnsguru:
@repperille this is a rather large change 

We consider the registry the ultimate authority on their listing, of course, and we will do as directed, but feel it important to elevate to your attention that it would be irresponsible of us to blindly accept this submission without highlighting to you that it will have some consequences for registrants.

@sleevi is making a point, and I second it, that removing all of these sub-entries could have sweeping, unintended impacts on any registrants of those domains identified in those URLS.  

Updating the reference to them in the comment line is helpful, but removing these from being included within the PSL could have sweeping impacts on software libraries, browsers, search engines, cookies, SSL certificates and other consequences.

Additionally, the submission has a conflict with the base and might require a re-basing to correct.

Could I suggest that you re-base the change, and do it in two phases? 

The first PR change you submit leave the subdomains intact but update the header comments, while you make a determination of the impacts of the change and evaluate how many subdomains are present underneath those nested TLDs?

The subsequent PR change could remove the subdomain lists once you have made whatever evaluations.

fsodic:
@sleevi @dnsguru ,
Just a **suggestion**, I would rather agree with the separation of the subdomains for their respective reasons directly here. It is more rational when deleting multiple lists as I think volunteers will get burdened seeing so many lists being deleted or added at once which is time consuming to check.

like mil.no which is possible for military (in my country mil.id) sub ccTLD example.mil.no example2.mil.no will mutually link cookies etc.

Changes especially deletions like this give freshness to our suffix list but also impose big risks like cookie security as above.

repperille:
Hi @sleevi, @dnsguru and @fsodic you are right about that this PR might break things too much. I will close this PR and make a new one where only the URLs to where the authoritative data can be found.

After a discussion at work (Norid, the registry for .no) we have clarified some issues and there was a misunderstanding about the removal of all the second level domains. We will not update the lists but update the information about where the authoritative lists can be found. Next step for us might be to make the lists of second level domains available in a machine readable interface

br
Richard Persson

sleevi:
Fantastic! Thanks much!

dnsguru:
@repperille thank you for taking this conservative approach.  Will watch for the header PR .   Please say hi to Hilde for me

PR #1166, opened 2020-12-13
CI: Add Github Actions workflow for TLD updates PRs.

cpu:
## Summary

I wanted to describe the situation in detail but it made for a long PR description. Here's the important parts up front:

* Travis & the old gTLD PR automation is effectively broken.
* Let's move the gTLD automation into this repo, using Github actions.
* If you agree, stamp this PR with an approval.
* And we need to coordinate on adding the @tld-update-bot's Github "personal access token" as a secret env var in this repo's config. (_Alternatively: maintainers can make their own account and plug a PAT in._)

## Background

Since ~August 2019 the gTLD section of `public_suffix_list.dat` has been kept up to date by a combination of the `tools/newgtlds.go` command [in this repo][newgtlds-cmd] and automation I've been maintaining in [a separate Github repo][cpu-autopull]. To date the bot has opened [67 pull requests][pr-stats].

The automation in `cpu/psl-autopull` is based on a fork of [nbio/autopull][nbio-autopull] that @weppos has been using to keep [publicsuffix-go][psl-go] updated with this repo. It's implemented as a BASH script that is meant to be run by a CI provider (in our case, [Travis CI][autopull-travis]). Its configuration is spread between [the script][autopull-script] and Travis CI environment variables.

Unfortunately as of late the Travis CI platform has been letting us down. I've moved several other repos to Github Actions because the Travis CI open source build queue time was getting far too long. More importantly the "cron" build that is scheduled to run the autopull script once a day has not been running reliably. This is causing updates to arrive late (when I remember to kick the builds by hand...).

Many open source projects are replacing Travis with Github Actions and I think it makes sense here too.

[newgtlds-cmd]: https://github.com/publicsuffix/list/blob/7dfbaf93f1444ebb239403ebb897e17b508efe59/tools/newgtlds.go
[cpu-autopull]: https://github.com/cpu/psl-autopull
[pr-stats]: https://github.com/publicsuffix/list/pulls?q=is%3Apr+author%3Atld-update-bot+
[nbio-autopull]: https://github.com/nbio/autopull
[psl-go]: https://github.com/weppos/publicsuffix-go
[autopull-travis]: https://travis-ci.com/github/cpu/psl-autopull
[autopull-script]: https://github.com/cpu/psl-autopull/blob/master/autopull

## Github Actions

Replacing Travis with Github Actions has several advantages:

* The configuration is entirely in this repo, not spread across multiple repos & services.
* It removes a dependency on an external service & their billing/product changes. We already depend on Github.
* Better automation. The [create-pull-request][create-pr-action] action is clever enough to not open duplicates. The old Travis autopull automation would create duplicate PRs if they weren't merged fast enough.
* The Github actions cron schedule is much more flexible. This PR configures the workflow to run once an hour vs once a day with Travis.

[create-pr-action]: https://github.com/peter-evans/create-pull-request

## Bot Account

This branch configures [create-pull-request][create-pr-action] to open PRs from a fork using a Github [personal access token (PAT)][create-PAT] stored as a repository secret env var called `"BOT_PAT"`. 

**This will need to be set up by a repository administrator by following the [docs for creating an encrypted secret for a repository][create-secret].** 

The PAT must correspond to the Github account that owns the fork of the repository configured in the `push-to-fork` setting of the workflow.

I have configured the `push-to-fork` value to `tld-update-bot/list` on the assumption we'll be adding @tld-update-bot's PAT to the `BOT_PAT` secret var. **If the maintainers would prefer to create their own Github account please make sure to update this piece of config to point to the new account's fork.** Otherwise we should exchange the @tld-update-bot secret PAT value out of band using whatever is least painful for you (GPG, encrypting to SSH keys, quantum key exchange, smoke signals).

To be 100% explicit: in either case adding the PAT to the repo config as an encrypted env var does **not** grant any access to the repository to the bot. The bot opens PRs from a fork to avoid needing push access to this repository. No extra access should be configured.

I'll also note that it _is_ possible to not have any bot account, PAT, or fork involved by using the Github Action's default token. This is how I implemented [similar logic](https://github.com/zmap/zlint/pull/513) for another open source repo. It has one important caveat: without a dedicated bot github account the PRs opened by the workflow won't in turn run further workflows (e.g. unit tests, etc). I discuss some of this in detail [in this issue](https://github.com/zmap/zlint/issues/515). For this repository because the data format is more complex & not a Go datastructure I think it's important to maintain the separate bot account + full CI workflow potential for the automation PRs.

[create-PAT]: https://docs.github.com/en/free-pro-team@latest/github/authenticating-to-github/creating-a-personal-access-token
[create-secret]: https://docs.github.com/en/free-pro-team@latest/actions/reference/encrypted-secrets#creating-encrypted-secrets-for-a-repository

## Future Work

It would be nice to move the repository CI from Travis to Github Actions as well. I suspect it will be a better experience overall. I would be willing to take this work on but thought it was best to put one toe in with this automation before revamping anything else.

### DIFF
--- a/.github/workflows/tld-update.yml
+++ b/.github/workflows/tld-update.yml
@@ -18,8 +18,9 @@ jobs:
         with:
           go-version: ^1.15
 
-      - name: Set current date as env variable
-        run: echo "NOW=$(date +'%Y-%m-%dT%H:%M:%S %Z')" >> $GITHUB_ENV
+      - name: Set current date
+        id: get-date
+        run: echo "::set-output name=now::$(date +'%Y-%m-%dT%H:%M:%S %Z')"
 
       - name: Run patchnewgtlds
         run: tools/patchnewgtlds
@@ -30,9 +31,9 @@ jobs:
         with:
           token: "${{ secrets.BOT_PAT }}"
           push-to-fork: tld-update-bot/list
-          commit-message: "util: gTLD data autopull updates for ${{ env.NOW }}"
-          title: "util: gTLD autopull updates for ${{ env.NOW }}"
-          body: "Public suffix list gTLD data updates from `tools/patchnewgtlds` for ${{ env.NOW }}."
+          commit-message: "util: gTLD data autopull updates for ${{ steps.get-date.outputs.now }}"
+          title: "util: gTLD autopull updates for ${{ steps.get-date.outputs.now }}"
+          body: "Public suffix list gTLD data updates from `tools/patchnewgtlds` for ${{ steps.get-date.outputs.now }}."
           committer: "GitHub <noreply@github.com>"
           author: "GitHub <noreply@github.com>"
           labels: tld-update
--- a/.github/workflows/tld-update.yml
+++ b/.github/workflows/tld-update.yml
@@ -1,0 +1,45 @@
+name: tld-update
+on:
+  schedule:
+    # Run every hour, at the 15 minute mark. E.g.
+    # 2020-11-29 00:15:00 UTC, 2020-11-29 01:15:00 UTC, 2020-11-29 02:15:00 UTC
+    - cron:  '15 * * * *'
+jobs:
+  psl-gtld-update:
+    name: Check for TLD data updates
+    runs-on: ubuntu-latest
+    steps:
+
+      - name: Check out code
+        uses: actions/checkout@v2
+
+      - name: Set up Go
+        uses: actions/setup-go@v2
+        with:
+          go-version: ^1.15
+
+      - name: Set current date as env variable
+        run: echo "NOW=$(date +'%Y-%m-%dT%H:%M:%S %Z')" >> $GITHUB_ENV
+
+      - name: Run patchnewgtlds
+        run: tools/patchnewgtlds
+
+      - name: Create pull-request
+        id: cpr
+        uses: peter-evans/create-pull-request@v3
+        with:
+          token: "${{ secrets.BOT_PAT }}"
+          push-to-fork: tld-update-bot/list
+          commit-message: "util: gTLD data autopull updates for ${{ env.NOW }}"
+          title: "util: gTLD autopull updates for ${{ env.NOW }}"
+          body: "Public suffix list gTLD data updates from `tools/patchnewgtlds` for ${{ env.NOW }}."
+          committer: "GitHub <noreply@github.com>"
+          author: "GitHub <noreply@github.com>"
+          labels: tld-update
+          branch: psl-gtld-update
+          delete-branch: true
+
+      - name: Check outputs
+        run: |
+          echo "Pull Request Number - ${{ steps.cpr.outputs.pull-request-number }}"
+          echo "Pull Request URL - ${{ steps.cpr.outputs.pull-request-url }}"


cpu:
@dnsguru @weppos @sleevi Would appreciate your thoughts on this PR when you have a chance to review. I'll continue to kick the old automation until we've all had a chance to discuss what the best path forward is based on the description in this PR.

Thanks!

cpu:
Also worth mentioning that I tested it successfully: https://github.com/cpu/list/pull/1 

I had to cheat slightly and comment out the `push-to-fork` setting and add @tld-update-bot as a contributor to my fork. Github won't let you fork a fork so otherwise the bot wasn't able to push `push-to-fork` to open an automated PR on my fork of the list.

.github/workflows/tld-update.yml
```diff
@@ -0,0 +1,45 @@
...
+      - name: Check out code
+        uses: actions/checkout@v2
+
+      - name: Set up Go
+        uses: actions/setup-go@v2
+        with:
+          go-version: ^1.15
+
+      - name: Set current date as env variable
+        run: echo "NOW=$(date +'%Y-%m-%dT%H:%M:%S %Z')" >> $GITHUB_ENV
```
sleevi, line 22: This is mostly a nit, but I'm curious the choice for $GITHUB_ENV versus using a step output. The main argument against I can think of is that it requires an explicit ID for the step, but were there other considerations?

That is
```
- name: Set the current date as env variable
  id: get-date
  run: echo "::set-output name=now::$(date +'%Y-%m-%dT%H:%M:%S %Z')"
```

and then later using `${{ steps.get-date.now }}`

Not a blocker by any means, and I'm not sure if either approach handles newline injection that well, but that's what made me think of it.

.github/workflows/tld-update.yml
```diff
@@ -0,0 +1,45 @@
...
+          token: "${{ secrets.BOT_PAT }}"
+          push-to-fork: tld-update-bot/list
+          commit-message: "util: gTLD data autopull updates for ${{ env.NOW }}"
+          title: "util: gTLD autopull updates for ${{ env.NOW }}"
+          body: "Public suffix list gTLD data updates from `tools/patchnewgtlds` for ${{ env.NOW }}."
+          committer: "GitHub <noreply@github.com>"
+          author: "GitHub <noreply@github.com>"
+          labels: tld-update
+          branch: psl-gtld-update
+          delete-branch: true
```
sleevi, line 40: In the context of the discussion at https://github.com/zmap/zlint/issues/515#issuecomment-743776240 , does this become easier if we change the default?

And we'd also be affected by the lack of testing runs, right? Or does the fact that we're still on Travis and have not yet migrated to GH Actions make this largely moot until we do?

sleevi (approved):
Thanks for doing this @cpu! Just double checking my understanding on a few bits, but I think we're good to merge (and I too welcome our GitHub Action overlords)

.github/workflows/tld-update.yml
cpu: > This is mostly a nit, but I'm curious the choice for $GITHUB_ENV versus using a step output. The main argument against I can think of is that it requires an explicit ID for the step, but were there other considerations?

I fiddled with a few options for this when I first implemented it in ZLint and I remember having trouble getting an alternative approach to work but I can't remember the details. Possible it was operator error :-)

I'll try testing a step output based approach to see if I can get it working. It does seem cleaner 👍

.github/workflows/tld-update.yml
cpu: > In the context of the discussion at zmap/zlint#515 (comment) , does this become easier if we change the default?

The situation in this repo is slightly different because I chose to keep the separate bot account + to have the PRs opened from a fork. The branch the bot creates will be on the fork repo, whereas in ZLint there is no bot presently and the branch is created on the main repo.

> And we'd also be affected by the lack of testing runs, right? Or does the fact that we're still on Travis and have not yet migrated to GH Actions make this largely moot until we do?

Similar answer here: using the separate bot account makes this repo's implementation a little bit more complex but will avoid the problem related to testing runs in the future if CI is moved from Travis. 

You're correct that it's largely moot until the Travis bits have been migrated - no matter how the PR is created those should continue to work the same.

dnsguru:
@cpu -

Many thanks for this and the manual interventions over the past couple of
weeks...

I just have a metaQ - will this still mean we're manually reviewing the PRs
and squash/merging ?

-Jothan

cpu:
> I just have a metaQ - will this still mean we're manually reviewing the PRs and squash/merging ?

@dnsguru That's correct - this side of the PR workflow will remain unchanged.

.github/workflows/tld-update.yml
sleevi: D'oh, resolving this, I *completely* missed the discussion when you opened the PR.

dnsguru:
Does this need merging, or other next steps?

cpu:
> Does this need merging, or other next steps?

@dnsguru I was going to take a pass at @sleevi's [nit about the date](https://github.com/publicsuffix/list/pull/1166#discussion_r542577189). I have some holiday time this week and should be able to push a commit shortly.

Afterwards I believe it is ready to go. We will still need to coordinate on adding the @tld-update-bot's PAT to the repo configuration. @dnsguru Are you the best person to ask to do that part?

Thanks!

.github/workflows/tld-update.yml
cpu: @sleevi Fixed up in ff6c3ba, thanks for the suggestion :+1: Tested successfully with [another mock PR](https://github.com/cpu/list/pull/3).

> and then later using ${{ steps.get-date.now }}

One small note: using this syntax literally resulted in an empty string (eg [this PR](https://github.com/cpu/list/pull/2)). I had to reference the `now` output ID within a more specific keyspace: `${{ steps.get-date.outputs.now }}`.

cpu:
> I was going to take a pass at @sleevi's nit about the date. I have some holiday time this week and should be able to push a commit shortly.

All set: ff6c3ba.

dnsguru:
@sleevi happy new year.  How is this looking?  Would like to merge this and we need to ensure it is automated as well.

sleevi:
OK, I've added the `BOT_PAT` secret, and am now going to try merging and hoping it works :)

cpu:
Porting my comment from the [first bot PR](https://github.com/publicsuffix/list/pull/1183)

Upside: we know the PAT is correct and the action is executing correctly.

Downside: it's revealed one bug. The previous automation had special case logic to discard these "header date update only" diffs. The new automation does not and so opened a PR that's a no-op :-/

@sleevi @dnsguru How would you like to proceed? I'm slammed this week with other commitments and likely won't have a chance to fix this bug for a few days. Do you want to revert 9d89164 until then so we don't get spammed with silly PRs?

Other options: if you leave #1183 open the bot will just keep updating it with new timestamps but never create a new PR.

sleevi:
@cpu I went ahead and just renamed the file in https://github.com/publicsuffix/list/commit/e69dbb941a7e17799a873110619925ba56b10237 so that it won't run :)

cpu:
@sleevi Sounds good to me. I'll ping folks when I have a fix worked out.

If anyone wants to poke at it before I get a chance I went back and figured out how the old version handled this and it's pretty hacky. The [part of the script](https://github.com/cpu/psl-autopull/blob/7ffe171735407ca471df0bdad07d30f392342359/autopull#L40) that checks for diffs is using `git diff --numstat` and filtering on the output that matches the "header only" update with the expectation those are always a 1 line diff (vs content diffs that would be 2+).

dnsguru:
Definitely want it to not make a pr unless entries change (not just file
header is different from previous)...

Header on ICANN json file changes daily as I believe they generate it
daily, changes or not.

cpu:
Haven't forgotten about picking this work up again and fixing this last bug. Hoping to get a chance to take a look this coming weekend :crossed_fingers:

cpu:
> Hoping to get a chance to take a look this coming weekend 

Take a peek at #1204 - It ended up being a little bit more involved than I was hoping but I'm happy with the outcome in that branch.

PR #1175, opened 2020-12-23
add missing IDN ccTLDs:  for Bahrain and Laos ( .xn--mgbcpq6gpa1a and .xn--q7ce6a )

martimlobao:
* [ ] Description of Organization
* [X] Reason for PSL Inclusion
* [ ] DNS verification via dig
* [ ] Run Syntax Checker (make test)

* [ ] Each domain listed in the PRIVATE section has and shall maintain at least two years remaining on registration.

Description of Organization
====

Organization Website: 

Reason for PSL Inclusion
====
* .wed is a valid TLD since 2013: https://www.icann.org/resources/agreement/wed-2013-10-01-en
* .البحرين [.xn--mgbcpq6gpa1a] is a valid TLD for "albahrain": https://www.iana.org/reports/2020/xn--mgbcpq6gpa1a-report-20200110.html
* .ລາວ [.xn--q7ce6a] is a valid TLD for "Lao": https://www.iana.org/reports/2020/xn--q7ce6a-report-20200110.html
* .za is a valid TLD and the only one in this list of ICANN domains that is not included without any subdomain (even other TLDs which are typically only issued with a secondary domain like .uk are included as a standalone TLD)

DNS Verification via dig
=======

make test
=========
I'm unable to run the tests locally, I keep getting an error with `libicu` that I haven't been able to fix. If anyone wants to double-check that the tests don't fail, I'd appreciate it.

Fixes #1174

Issue #1174, opened 2020-12-23
Handling EBERO TLDs (ex: .wed)


martimlobao:
The list seems to be missing 3 valid TLDs, which are included on IANA.org: https://data.iana.org/TLD/tlds-alpha-by-domain.txt

Namely:
* .wed
* .xn--q7ce6a (and equivalent .ລາວ)
* .xn--mgbcpq6gpa1a (and equivalent .البحرين)

dnsguru:
Being discussed in PR #1175

dnsguru:
We should put to rest the .WED TLD discussion, as it was terminated and should not be present.
See: [ICANN gTLD JSON file, note "ContractTerminated": "True" for wed entry](https://www.icann.org/resources/registries/gtlds/v2/gtlds.json)

With respect to the IDNccTLDs for Bahrain and Laos, there's no official notice from the IANA or ccNSO when IDN ccTLDs process, so we rely on the community or the repsective NIC to submit these, so thank you for submitting these.

dnsguru:
Thank you for splitting the PR for this into separate issues for the IDN ccTLDs for Bahrain and Laos (#1175) / .WED (#1176) , and for your submission and interest in helping keep the PSL fresh and updated.

We were able to merge in #1175, but #1176 is a little trickier due to it being a new precedent.  This Issue may sit open for a while.  .WED domains do exist in the wild, but the TLD is in EBERO state, which is ambiguous as to who to discuss the state of inclusion / exclusion from the PSL.

My bias towards keeping the issues/PR flowing into a closed state had me quickly thinking that .WED being overlooked by the automation and excluded from the PSL by automation was appropriate.

The typical scenario of TLDs which have been marked as cancelled in the ICANN .json file are those which did not "open" or "launch" and thus contain no names other than the nic.[tld] or registry-reserved names.  In the case of those TLDs which flow into EBERO, that is not a valid assumption, as there may be customers with valid registrations within those strings.

My instinct is to proclaim that individual site owners with legacy registrations within .WED are able to open PR to set their domains' options within the PRIVATE section.  IF a legacy registrant has a viable entry in the .WED zone that delegates DNS in the .WED zome to their nameservers, they would be about to validate their authoritative administration of that subdomain. 

[...]

PR #1176, opened 2020-12-24
Add .wed

martimlobao:

Description of Organization
====

Organization Website: 

Reason for PSL Inclusion
====

Even though .WED is listed [by ICANN](https://www.icann.org/resources/registries/gtlds/v2/gtlds.json) as having a terminated contract, there are reasons it might make sense to still include it at present:

* There are still live websites that use the .wed TLD, such as [get.wed](http://get.wed/), [atgron.wed](https://www.atgron.wed/), and even [may27.wed](https://may27.wed/). This is not the case for any other site on any other terminated TLD listed on [Google Search](https://www.google.com/search?q=site%3A.wed), which makes .wed unique in that regard.
* It is still listed by ICANN as an active TLD in this other [list of TLDs](https://data.iana.org/TLD/tlds-alpha-by-domain.txt), as well as in the [Root Zone Database](https://www.iana.org/domains/root/db) where is managed by "Emergency Back-End Registry Operator Program - ICANN." Every other terminated TLD is missing from the first list and has status "not assigned" in the second list.
* Despite being listed as `"contractTerminated" : true` [by ICANN](https://www.icann.org/resources/registries/gtlds/v2/gtlds.json), the record for .wed contains a null `removalDate`. Every other TLD which is listed as `contractTerminated` either has no `delegationDate` or has a non-null `removalDate` in the past.

As long as .wed continues to be managed by ICANN's [EBERO](https://www.icann.org/resources/pages/ebero-2013-04-02-en), we might want to maintain it in the list of active TLDs. If the list of new gTLDs must be auto-generated, then the logic for adding TLDs should be adjusted to account for the last point above.

DNS Verification via dig
=======

make test
=========

See discussions in #1174 and #1175.

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -10227,6 +10227,9 @@ weber
 // website : 2014-04-03 DotWebsite Inc.
 website
 
+// wed : 2013-10-01 Atgron, Inc.
+wed
+
 // wedding : 2014-04-24 Minds + Machines Group Limited
 wedding
 


dnsguru:
I propose we close and wontfix this PR.  I've renamed the #1174 and then suggest we revisit this PR once we can get some feedback from ICANN about what to expect with EBERO strings.

Will leave this here for others to comment on but will close as wontfix in a week if no feedback.

@sleevi @weppos feedback?

martimlobao:
@dnsguru Sounds like this is a wontfix then, right? Unless @sleevi or @weppos have any objections?

dnsguru:
For now, probably.  Would need to get the zone file operator to add in .wed
TXT record or alternatively have specific entries with the DNS validation
for the specifically affected names and have those added in the private
section.  First instance of a request of its kind  -  this is the only TLD
that progressed into EBERO operations.

I think close/wontfix is the fast path for now, and then we can reference
this dialog later.

My dissenting here is OCD... That the TLD is in the root zone / IANA list,
so it should be included as a base entry, so I don't like close/wontfix
from a curators' point of view.

Jothan Frakes
+1.206-355-0230 tel
+1.206-201-6881 fax

dnsguru:
My OCD was triggered after my last comment and I located .wed in https://data.iana.org/TLD/tlds-alpha-by-domain.txt and I think it merits a re-look at the automation so I opened an issue on this. 

The TLD is atypical - it is the only EBERO name - so it falls outside of the norms of the processing done.  The json file from ICANN says it is cancelled like .nationwide or .onyourside but unlike those it remains root-listed and present in the IANA TLDs alpha by domain file.

So I opened #1325 to review the autopull.  Let's leave this PR open as a reminder, for now.

dnsguru:
At this point, I'll close this out because there is now 'drift' and a conflict with the base PSL file which means a rebase is required.  

We can revisit WED in another PR adding WED to the private section but I want to consult with ICANN first on this.  Should we do this, we want to be sure to reference this PR when doing so to ensure there is a good transparent record on the matter.

aaomidi:
Reading the BRs, this has created an interesting situation for these domains obtaining certs for themselves. 

If they’re being served on the root zones, I think PSL should consider having them on here. 

Weird situation.

PR #1078, opened 2020-07-29
add pages.internal.impertrix.com

impertrixsolutions:
* [ ] Description of Organization
* [ ] Reason for PSL Inclusion
* [ ] DNS verification via dig
* [ ] Run Syntax Checker (make test)

Description of Organization
====

Organization Website: https://impertrix.com

we are a hosting provider/CSP

Reason for PSL Inclusion
====

we want to use this with gitlab pages to provide to internal engineers high efficiency when prototyping and easy deployment

DNS Verification via dig
=======

dig +short TXT _psl.pages.internal.impertrix.com
"https://github.com/publicsuffix/list/pull/1078"
make test
=========

make looks good, pass 5/5

PR #1195, opened 2021-01-27
psl.ninja individual domain holder request.dat

ddworken:

Description of Organization
====

Organization Website: https://daviddworken.com

Hi! I'm an individual security researcher interested in having access to a domain that is on the PSL. 

Reason for PSL Inclusion
====

A number of times in the past I have wished that I had a test domain on the PSL that I could use for testing code that interacts with site based restrictions as used in browsers. As one example of this, I wish to have a domain on the PSL to allow me to create better POCs for potential issues around browser's site-keyed double caches. 

I have registered `psl.ninja` for the next 3 years and will keep it renewed and under my control indefinitely. 

DNS Verification via dig
=======

```
dig +short TXT _psl.psl.ninja      
"https://github.com/publicsuffix/list/pull/1195"
```

make test
=========

Tests pass :+1:

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -11180,6 +11180,10 @@ builtwithdark.com
 // Submitted by Richard Li <secalert@datawire.io>
 edgestack.me
 
+// David Dworken : https://daviddworken.com
+// Submitted by David Dworken <david@daviddworken.com>
+psl.ninja
+
 // Debian : https://www.debian.org/
 // Submitted by Peter Palfrader / Debian Sysadmin Team <dsa-publicsuffixlist@debian.org>
 debian.net


dnsguru:
Hi @ddworken - I'd like one too, but don't qualify for the same reason I am about to describe.

Due to the file size and download frequency, we decline adding things with explicitly small demographic user base, and have been doing so for quite some time.  Were this to be advanced and merged, it is in contrast to past requests that have been declined due to the same rationale where there were tens or hundreds of users.  

See #1078 - there are more but the point is, modesty on file size for maximum benefit.

I'd let @sleevi weigh in on this, perhaps there is more to this request than I am aware or I might be missing something.

dnsguru:
I field the majority of the PSL PR's during the past few years, and I am
more involved within the ICANN-o-Sphere than with the browser / standards /
CA stuff - I just responded with a polite decline on this, but now see that
there may be some connection with the efforts of what-wg and security
related efforts.  I'm not involved in that group activity and as-such am
not as familiar with the roster of people involved.

Hard to track all this, but I think I'd do a better job of it if I was more
in the loop on what-wg and some of the ca forum stuff....

How do I get into that circle so that I can recognize when atypical
requests like this that would typically get declined which might merit
overrides?

-J

ddworken:
No worries, that answer totally makes sense and I half expected that to be the answer anyways. :) I just thought that this might be an easier option for me than compiling Chrome etc myself to manually add my own domains to the PSL.

PR #1226, opened 2021-03-09
Switch the TLD update workflow to daily

sleevi:
Rather than checking for IANA updates every hour at 15 minutes past the hour, run every day, at 15:00 UTC

### DIFF
--- a/.github/workflows/tld-update.yml
+++ b/.github/workflows/tld-update.yml
@@ -2,9 +2,8 @@
 name: tld-update
 on:
   schedule:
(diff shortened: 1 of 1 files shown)


sleevi:
@cpu FYI

cpu:
> Rather than checking for IANA updates every hour at 15 minutes past the hour, run every day, at 15:00 UTC

@sleevi SGTM, thanks for making the change 👍

dnsguru:
Thanks @sleevi this is a good change

PR #1231, opened 2021-03-16
Add jprq.io

azimjohn:


dnsguru:
If this gets re-opened - locate the change in the #PRIVATE section please.  This one would be rejected

Issue #1245, opened 2021-03-24
New interaction between IOS 14.5 PCM and Facebook Pixel causing increase in PSL inclusion requests


dnsguru:
(Day 60+ Update, Added June 4, 2021)
----
**TL;DR:** _**Here to fix an issue w Facebook Pixel?**_  [Click here for their guidance](https://www.facebook.com/business/help/126789292407737?id=1205376682832142).

Facebook has updated their support info to route people to [this page](https://www.facebook.com/business/help/126789292407737?id=1205376682832142) with more info for their [Facebook Pixel customers and issue with PCM since IOS 14.5](https://www.facebook.com/business/help/126789292407737?id=1205376682832142),  This includes a form to aid in determining if it is appropriate to make a request to the PSL.  

We have added to our PR form here some additional acceptance elements, and want to caution requestors that rollbacks have delays in downstream PSL integrators that are beyond the control of the project volunteers here to do anything about.  Additionally, new requests for additions are prioritized ahead of rollback requests.

Most of the time it is not appropriate to request a PSL entry to fix the Pixel issue and people have been nerfing their primary URL in the process of making minimal-effort cut-paste type requests, so this updated guidance and intake form will hopefully cut down the time wasted for all involved.

==============

(Original issue, Opened March 24, 2021)
----

Similar to other third party systems (CA's, Cloudflare, etc) using PSL as a ETLD+1 quality sieve, Facebook / Apple interaction between iOS 14 and the Facebook Business Pixel tracker for hosted names has introduced a new interdependancy.

We are seeing an increased number of PSL requests due to a new way Apple iOS 14 changes stuff with Facebook Pixels.

Upon some research, I found the following in the FB for business Pixel documentation.

From "How Apple’s iOS 14 Release May Affect Your Ads and Reporting"
https://www.facebook.com/business/help/331612538028890?id=428636648170202 

>If you plan to deliver ads optimized for conversion events that occur on your business’s website:
> You may need to verify your website’s domain to help avoid any future disruption of your website campaigns. Domain verification must be done for the effective top level domain plus one (eTLD+1). For example, for www.books.jasper.co.uk, books.jasper.co.uk and jasper.co.uk the eTLD+1 domain is jasper.co.uk.
[...]

PR #1260, opened 2021-04-01
adding https://www.riseinteractive.com to psl

steven-chen-rise:

Description of Organization
====

Organization Website: https://www.riseinteractive.com 

Director of Web Development at Rise Interactive.
Rise Interactive is a digital media agency.

Reason for PSL Inclusion
====

Rise is interested in the public suffix list due an Apple iOS 14 update.
Facebook has provided documentation in regards to Enabling More Domain Verification Use Cases in Aggregated Event Measurement.
It mentions it will be supporting the Public Suffix List for domain verification and event configuration.
Rise is adding its domains to the public suffix list to ensure compatability and testing in regards to ad campaigns due to the update.

DNS Verification via dig
=======

make test
=========

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -13144,6 +13144,11 @@ hzc.io
 wellbeingzone.eu
 wellbeingzone.co.uk
 
+// Rise Interactive, Inc. : https://www.riseinteractive.com
+// Submitted by Steven Chen <steven.chen@riseinteractive.com>
+riseinteractive.com
+staging.riseinteractive.com
+
 // Rochester Institute of Technology : http://www.rit.edu/
 // Submitted by Jennifer Herting <jchits@rit.edu>
 git-pages.rit.edu


steven-chen-rise:
Creating the PR to get the url available for the DNS verification for dig.
Will be reaching out to our internal IT team to get the DNS updates and the registration expiration updated.
I will comment back on the PR when those changes are completed.

dnsguru:
Closing as wontfix - see #1245

dnsguru:
https://github.com/privacycg/private-click-measurement/issues/78

bedfordsean:
Hi @steven-chen-rise ,

The guidance from Facebook (clarified last week in the help centre article) is that this process is intended for large platform partners (like Shopify) to allow businesses that they host on behalf of to successfully validate their domain as if it were an eTLD+1.

The Public Suffix List is not useful, nor intended to be used as a means to gain additional subdomain events reporting. If you just want to use subdomains as part of Facebook's aggregate event measurement, you already get 8 events under your eTLD+1 of riseinteractive.com.

Can you confirm why you need subdomains under this eTLD+1 to be individually treated as eTLD+1s with cookie separation, which would be the behaviour post-addition to the PSL?

steven-chen-rise:
@bedfordsean 

Apologies for the late response. I reached out to our team member who requested this and here is the response I received.

We are an agency that represents a number of clients that are in similar situations where they only control a subdomain of the parent domain they are attached to...and those clients would like to be able to rank their own events rather than succumb to the parent domain's decisions and we were looking to test if we would be able to instruct our clients on what to do in order to get registered on the suffix list to enable them to rank their own events.

So the Rise Interactive implementation is a POC for one of our clients. If we need to clarify specifically with regards to the client in question, I'd be happy to reach out via email to connect.

bedfordsean:
Thanks @steven-chen-rise 

There are two effects I want to call out based on how public suffix listing works:
1. Once you're on the list, there is a limitation of same domain cookies and local storage, this means that "riseinteractive.com" can no longer have its own cookies, only subdomains can. I'm not sure that I see any subdomains of riseinteractive.com that show reasonable use cases of different clients being hosted on your domain (or none that a Google search shows up anyway!)
2. The browsers will enforce the behaviour described in #1 based on their own update cadence of the public suffix list. Some browsers don't update their lists more regularly than every few months. This means that if you're on the list and a browser updates their copy of the list, and you later decide to not be on the list, there may not be an easy way to back out the behaviour; it's not as simple as submitting another request here to get taken off of the list.

Given that you say this is a proof of concept for a (singular?) client, you may want to think more about these two considerations. I don't believe we'd want to approve public suffix listing for a single use case that isn't a long term direction you want to go in

PR #1262, opened 2021-04-03
Add nel.family

BCNelson:

Description of Organization
====

Several private families that shares a TLD

Reason for PSL Inclusion
====

This domain is used be many people related to the Nelson Family. It is used for several commercial purposes as well as the FQDN for several home networks.

DNS Verification via dig
=======
```
dig +short TXT _psl.nel.family
"https://github.com/publicsuffix/list/pull/1262"
```

make test
=========

https://gist.github.com/BCNelson/4cbeb064496e357e282f7e79b29f8a45

Example Domains
======
```
paul.nel.family
craig.nel.family
h.nel.family
```

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -12636,6 +12636,10 @@ to.work
 // Submitted by Tocknicsu <admin@nctu.me>
 nctu.me
 
+// Nelson Family: https://nel.family/
+// Submitted by Bradley Nelson <bradley@nel.family>
+nel.family
+
 // Netlify : https://www.netlify.com
 // Submitted by Jessica Parsons <jessica@netlify.com>
 netlify.app


sleevi:
Could you share more details about how this namespace is used? This seems like the relative risk from not being on the PSL is quite small (i.e. may affect <100 Internet users), while the cost of distributing this to billions of users may be quite large.

BCNelson:
It is used by a family but the family is a large one. For example we host a family directory for ~650 active monthly users.

jesse-savary:
I'm not understanding the benefit to having your family's network listed in the PSL. `nel.family` seems private/pretty locked down, what's the benefit to listing it in the PSL?

BCNelson:
It would benefit the security of nearly 1000 people who use it. Are there guidelines that I missed for inclusion in the PSL? Is it only for domains that are used excursively for commercial uses? How much money do I need to be making or visitors do I need to have before I can secure the site?

jesse-savary:
It's right there in the name: **Public** Suffix List. Is there any public information available about `nel.family`?

I'm not a contributor to the project; just genuinely curious as to where the line is and what the benefit to your domain is.

dnsguru:
PSL use is voluntary.  There is really no relationship with all the various downstream projects, programs, integrators etc that is entirely voluntary, and some of the things that keep their rafts tied to this island that allow for all the interop (and keeps them from all starting their own and balkanizing things) is that the list doesn't work around security or other limits that they might have and the list is modest about the file size and scale of names included on the list.

It is a delicate balance that we maintain as volunteers.   I myself would be delighted to add a surname of my domain for family use, but I respect that the file circulates to millions of places and gets integrated in major software browsers and systems.  

If nel.family had 10s of thousands of users this would not cause the question.  Thank you for understanding the rationale behind our declining to add this.

PR #1265, opened 2021-04-05
Add noop.app and developer.app to PSL

nathanielks:

Description of Organization
====

Organization Website: https://noop.app/ 

Noop (a play on words for no-op, or no operations) is a Platform as a Service enabling developers to deploy their applications without the need to manage operations themselves. 

Reason for PSL Inclusion
====

Noop makes use of Let's Encrypt to issue TLS certificates for our customers.

The domains have been renewed for the next 2 years.

DNS Verification via dig
=======

```
dig +short TXT _psl.developer.app
"https://github.com/publicsuffix/list/pull/1265"
```

```
dig +short TXT _psl.noop.app
"https://github.com/publicsuffix/list/pull/1265"
```

make test
=========

```
# TOTAL: 3
# PASS:  3
# SKIP:  0
# XFAIL: 0
# FAIL:  0
# XPASS: 0
# ERROR: 0

# TOTAL: 5
# PASS:  5
# SKIP:  0
# XFAIL: 0
# FAIL:  0
# XPASS: 0
# ERROR: 0
```

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -12653,6 +12653,11 @@ nh-serv.co.uk
 // Submitted by Jeff Wheelhouse <support@nearlyfreespeech.net>
 nfshost.com
 
+// Noop : https://noop.app
+// Submitted by Nathaniel Schweinberg <noop@rearc.io>
+*.developer.app
+noop.app
+
 // Northflank Ltd. : https://northflank.com/
 // Submitted by Marco Suter <marco@northflank.com>
 *.northflank.app


nathanielks:
Something I could use some help with from the maintainers here is determining whether our current domain definitions fit our desired use case. For instance, users will be provided two urls to serve their app: `<app name>.noop.app` and `<app name>.<username>.developer.app`. If I understand the Guidelines correctly, `noop.app` would be ok in this pull request, but I am unsure if only specifying `developer.app` is enough, or if we should actually define it as `*.developer.app`. I appreciate your feedback!

nathanielks:
Noop.app was renewed yesterday for the next two years and Developer.app was just renewed for 2 years as well. https://lookup.icann.org/lookup still shows the unrenewed date for Developer.app, so it might just take a minute for their cache to update.

nathanielks:
This is now ready for review 😄

dnsguru (changes requested):
please re-order the two domains so that they are alphabetically sorted

dnsguru:
> Something I could use some help with from the maintainers here is determining whether our current domain definitions fit our desired use case. For instance, users will be provided two urls to serve their app: `<app name>.noop.app` and `<app name>.<username>.developer.app`. If I understand the Guidelines correctly, `noop.app` would be ok in this pull request, but I am unsure if only specifying `developer.app` is enough, or if we should actually define it as `*.developer.app`. I appreciate your feedback!

The * behavior varies among browsers, so there is not a good answer that is universal due to that variance in what folks do with the PSL in their software.
Is the question about how cookies would break for where setting them for noop.app directly with this change?  Adding an entry would allow for sub1.noop.app and sub2.noop.app to have distinct cookies and no cookies on noop.app..  Is that the desired result?  (answer same on developer.app)

sleevi:
> > If I understand the Guidelines correctly, `noop.app` would be ok in this pull request, but I am unsure if only specifying `developer.app` is enough, or if we should actually define it as `*.developer.app`. I appreciate your feedback!
> 
> The * behavior varies among browsers, so there is not a good answer that is universal due to that variance in what folks do with the PSL in their software.

So, I think there's a clear answer here: putting `developer.app` alone would **not** be sufficient to segment cookies at the `<app name>.<user name>.developer.app` case. While it would segment two different `<user name>`s, every `<app name>` for that user would be sharing the same space.

What @dnsguru is highlighting is that you cannot assume cookies that be set at either the `noop.app` level (the "parent"), nor at either the `<user name>.developer.app` nor the `developer.app` levels (i.e. assume that `*.developer.app` **also** implies `developer.app` is present)

nathanielks:
Many thanks @dnsguru and @sleevi !! I've added the additional record as suggested. This is now ready for final review.

dnsguru:
Gonna have to re-base it Nathan - its failing because the source you edited has changed.

nathanielks:
@dnsguru my apologies, I'm garbage with rebasing so I reset my branch and did a force push. I know that's messy on repos but I don't know if that's bad on forks. Happy to file a different PR if that's a problem.

That said, `make test` said `developer.app` was a "doublette/ambiguity," so I've removed it in favor of just `*.developer.app`. 

I know this has been a process, I am grateful for your patience 🙏 🙇

nathanielks:
(Rebasing included a number of recent commits outside of my branch in the PR which is why I did a reset and force push).

dnsguru (approved):

dnsguru:
- DNS Validation ok
- Tests ok
- Rationale ok
- No conflict with base

**Approved**

nathanielks:
Much obliged, @dnsguru !

PR #1271, opened 2021-04-08
added domain to register into public suffix list

AgenciaTrebolDigital:

* [X] Each domain listed in the PRIVATE section has and shall maintain at least two years remaining on registration.

Description of Organization
====

Organization Website: https://mastrpro.com

Reason for PSL Inclusion
====
I need to register domain into PSL because there is some IOS 14 restriction regarding Facebook event configuration
so I need to add the domain to PSL

DNS Verification via dig
=======
dig +short TXT _psl.mastrpro.com
"https://github.com/publicsuffix/list/pull/1271"

make test
=========
I ran the command but is shows "bash: make: command not found"

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -12571,6 +12571,10 @@ eastus2.azurestaticapps.net
 westeurope.azurestaticapps.net
 westus2.azurestaticapps.net
 
+// Mastrpro : https://mastrpro.com
+// Submitted by trebol <administracion@trebol.digital>
+mastrpro.com
+
 // minion.systems : http://minion.systems
 // Submitted by Robert Böttinger <r@minion.systems>
 csx.cc


dnsguru:
See #1245

dnsguru:
https://github.com/privacycg/private-click-measurement/issues/78

bedfordsean:
Hi @AgenciaTrebolDigital ,

The guidance from Facebook (clarified last week in the help centre article) is that this process is intended for large platform partners (like Shopify) to allow businesses that they host on behalf of to successfully validate their domain as if it were an eTLD+1.

The Public Suffix List is not useful, nor intended to be used as a means to gain additional subdomain events reporting. If you just want to use subdomains as part of Facebook's aggregate event measurement, you already get 8 events under your eTLD+1 of mastrpro.com.

Can you confirm why you need subdomains under this eTLD+1 to be individually treated as eTLD+1s with cookie separation, which would be the behaviour post-addition to the PSL?

PR #1272, opened 2021-04-09
Add *.kctf.cloud to the PSL

sirdarckcat:

Description of Organization
====

Organization Website: https://www.google.com

Google LLC is an American multinational technology company that specializes in Internet-related services and products, which include online advertising technologies, a search engine, cloud computing, software, and hardware. 

This project is to support https://github.com/google/kctf - infrastructure to run CTF competitions.

Reason for PSL Inclusion
====

`*.kctf.cloud` needs to be added for `Cookie Security` and for `Let's Encrypt` issuance limits, as it's meant as a public service self-registering TLD for security testing. It is a free public service, and we already are hitting LetsEncrypt limits (and also, we already have issues with some security configurations that require unique sites).

Registration expiration:
https://domains.google.com/registrar/search/whois/kctf.cloud?searchTerm=kctf.cloud
```
kctf.cloud - Registry Expiry Date: 2031-01-02T08:40:53Z
```

DNS Verification via dig
=======

https://toolbox.googleapps.com/apps/dig/#TXT/_psl.kctf.cloud

```
dig +short TXT _psl.kctf.cloud
"https://github.com/publicsuffix/list/pull/1272"
```

make test
=========

See attached file

[test.log](https://github.com/publicsuffix/list/files/6285260/test.log)

Fixes https://github.com/google/kctf/issues/288

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -11995,6 +11995,8 @@ goip.de
 
 // Google, Inc.
 // Submitted by Eduardo Vela <evn@google.com>
+*.kctf.cloud
+*.ctfcompetition.com
 run.app
 a.run.app
 web.app
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -11996,7 +11996,6 @@ goip.de
 // Google, Inc.
 // Submitted by Eduardo Vela <evn@google.com>
 *.kctf.cloud
-*.ctfcompetition.com
 run.app
 a.run.app
 web.app


sleevi:
Hey @sirdarckcat - Could you check https://github.com/publicsuffix/list/pull/1241 ? I think the same concerns raised on that issue apply here, and so I don't think we should take this.

sleevi:
Also, please see https://github.com/publicsuffix/list/wiki/Third-Party-Diffusion for the Let's Encrypt rate limits.

@weppos @dnsguru How do you feel about closing this WontFix?

dnsguru:
Wontfix is ok by me.

sirdarckcat:
Hi!

I think part of the goal of this as a wildcard TLD, is to allow those like redpwnc.tf to do redpwn.kctf.cloud as a TLD.

Since *.kctf.cloud is a "public" service, it's meant as a way to allow every competition to use it, and avoid having to need one per competition.

sirdarckcat:
another possible solution here is to make this even more generic (say a `.isolated` TLD) that people can register under, however, it seems for know a simple `*.kctf.cloud` might be enough?

what I'm looking to provide is a public-registrable TLD that has a PSL-suffix at +1. So if we add `*.isolated` I would like people to be able to register `whatever.isolated` so that `foo.whatever.isolated` is cross-site with `bar.whatever.isolated`.

the specific goal is to have this for different security challenges (like redpwnc.tf :) but this is likely useful for security just in general.. what would be the best way to approach this? we could try to get a gTLD, but it sounds a lot easier to start with just a normal domain name (happy to switch from kctf.cloud to something more generic like isolated.dev or so - I need to get naming approval first).

dnsguru:
> another possible solution here is to make this even more generic (say a
> .isolated TLD)
>
Interesting idea, but this doesn't address the concerns about adding.

Also, a new TLD, this is not currently an available option as there is not
a window open to apply for these.

There is a process under way within the ICANN community called subsequent
procedures or subpro that are defining parameters for when a company could
put in applications - no dates certain on when a next round would open.


The alt-path is defining via the ietf a special purpose tld such as .onion
did.

This project only accepts ICP-3 names (IANA root), those in the clear
application process from the 2012 round with ICANN, or those vetted by
IAB/IETF

> that people can register under, however, it seems for know a simple
> *.kctf.cloud might be enough?
>
> what I'm looking to provide is a public-registrable TLD that has a
> PSL-suffix at +1. So if we add *.isolated I would like people to be able
> to register whatever.isolated so that foo.whatever.isolated is cross-site
> with bar.whatever.isolated.
>

See above

> the specific goal is to have this for different security challenges (like
> redpwnc.tf :)
>

The PSL file size merits concerns of modesty that have caused a lot of PR
to be declined - also quite a lot of "labs" type scenarios have as well.
The example you are usin was a wontfix scenario that was not added

> but this is likely useful for security just in general.. what would be the
> best way to approach this? we could try to get a gTLD, but it sounds a lot
> easier to start with just a normal domain name (happy to switch from
> kctf.cloud to something more generic like isolated.dev or so - I need to
> get naming approval
>

Maybe get all that worked out and re-apply for a PR once this is a bit more
baked?

sirdarckcat:
Would it make sense to first get the concept apply in principle before I go get naming approval? :-)

It would be a bit silly to go apply for a domain for the purpose of allowing people to register domains whose subdomains are paths just to end up having the PR declined anyway

dnsguru:
Can you clarify... are you referring to 

1. PSL inclusion approval  (already litigated) - or  - 
2. approval of your supervisor or company?

sirdarckcat:
Perhaps a better question is, where is the right channel to discuss this?

I can make an issue explicitly to discuss this, or an email if that's better to involve others not on GitHub.

I can close this PR, as to avoid making this confusing

dnsguru:
> Perhaps a better question is, where is the right channel to discuss this?
Treating as rhetorical

> I can make an issue explicitly to discuss this, or an email if that's better to involve others not on GitHub.
Great - reference this PR in doing so to help fast-track understanding when doing so

> I can close this PR, as to avoid making this confusing
I will do this as it sat too long open since updates.

Please reference this PR if another opens

PR #1283, opened 2021-04-15
Add Selecteev.io

albandum:
**Reason of this Pull Request**
Selecteev.io is a SaaS that allows people to create call for applications for candidates to apply to programs and events. We provide specific subdomains for each call for applications with a custom design.
Every subdomain is in the form of foo.selecteev.io with an example here : https://demo.selecteev.io

We need people to be able to track their marketing efforts for their call for applications, notably on facebook which uses Public Suffix list to validate domains on some of its tools.

**Dig request output**
```
dig +short TXT _psl.selecteev.io
"https://github.com/publicsuffix/list/pull/1283"
```

Extra info
I am the founder of Selecteev and am mandated to make this request.

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -13200,6 +13200,10 @@ spdns.org
 // Submitted by Artem Kondratev <accounts@seidat.com>
 seidat.net
 
+// Selecteev : https://selecteev.io
+// Submitted by Alban Dumouilla <alban@selecteev.io>
+selecteev.io
+
 // Senseering GmbH : https://www.senseering.de
 // Submitted by Felix Mönckemeyer <f.moenckemeyer@senseering.de>
 senseering.net


dnsguru:
#1245

benjaminsavage:
Hi @albandum 

Would you mind telling me who from FB guided you towards opening a request here?

The guidance from Facebook (clarified last week in the help centre article) is that this process is intended for platform providers that provide subdomains for large numbers of small businesses which really ought to be treated as though they were in fact separate domains.

The Public Suffix List is not useful, nor intended to be used as a means to gain additional subdomain events reporting.

I did a Google for "*.selecteev.io" and do see various calls for applications. It's not super clear to me that these all represent fully separate businesses.

To clarify; do you wish for and intend for total cookie separation of each of those subdomains? Are you trying to provide a security boundary between these subdomains so that one cannot influence another? Is your goal for subdomain owners at other companies to be able to individually register through Facebook AEM and be treated as their own eTLD+1s?

Effectively beyond owning "selecteev.io", you'll not have much control for that site itself. For example, I note you have a /login page on that domain. This may not work as it does today if you proceeded with a PSL addition.  This is because "selecteev.io" will no longer be able to set domain level cookies after PSL addition. I'd recommend reading up on how PSL works here first: https://github.com/publicsuffix/list/wiki

albandum:
Hi @benjaminsavage and thanks for your answer !
We provide separate businesses with their call for applications pages. Some businesses have several call for applications, but each one has a different /login page (example here : https://demo.selecteev.io/login) with different cookies and we have the main site selecteev.io where users can manage the billing for their different platforms.

Me showing up here doesn't come from anybody at Facebook but from this SO post : https://stackoverflow.com/a/66344625/1017052

But maybe I am mistaken on what PSL exactly is doing ? Indeed, if this blocks any cookies from the root selecteev.io domain that would not work for us.

sleevi:
> But maybe I am mistaken on what PSL exactly is doing ? Indeed, if this blocks any cookies from the root selecteev.io domain that would not work for us.

Yes. That's what will happen. I'll keep this closed as WontFix.

In general, user generated/controlled content needs to be in an entirely separated domain namespace, at least with how cookies and the PSL works today. This is why there are sites like `github.com` which host/provided content via `github.io`.

PR #1296, opened 2021-04-26
Update public_suffix_list.dat for Grupo A education

Ttoledo28:
I would like to add subdomains of Group A to create aggregate events in facebook campaigns.

Examples of URLs:
https://unifeob.grupoa.education/orbita/subscription/?ies=101
https://pucpr.grupoa.education/orbita/subscription/?ies=111

Description of Organization
====

Organization Website:  https://www.grupoa.com.br/ 

 
My name is Tayro, I'm a marketing coordinator at Grupo A. We are a company in the education segment

Reason for PSL Inclusion
I would like to add subdomains of Group A to create aggregate events in facebook campaigns.
Examples of URLs:
https://unifeob.grupoa.education/orbita/subscription/?ies=101
https://pucpr.grupoa.education/orbita/subscription/?ies=111

DNS Verification via dig
=======

make test
=========

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -13671,4 +13671,8 @@ basicserver.io
 virtualserver.io
 enterprisecloud.nu
 
+// Company : http://www.grupoa.education
+// Submitted by Tayro Toledo <ttoledo@maiscampus.com.br>
+*.grupoa.education
+
 // ===END PRIVATE DOMAINS===
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -10640,6 +10640,10 @@ ltd.ua
 // 611coin : https://611project.org/
 611.to
 
+// A+ Educação : https://www.grupoa.com.br/
+// Submitted by Tayro Toledo <ttoledo@maiscampus.com.br>
+*.grupoa.education
+
 // Aaron Marais' Gitlab pages: https://lab.aaronleem.co.za
 // Submitted by Aaron Marais <its_me@aaronleem.co.za>
 graphox.us
@@ -13670,9 +13674,5 @@ bss.design
 basicserver.io
 virtualserver.io
 enterprisecloud.nu
-
-// Company : http://www.grupoa.education
-// Submitted by Tayro Toledo <ttoledo@maiscampus.com.br>
-*.grupoa.education
 
 // ===END PRIVATE DOMAINS===


dnsguru:
See and read #1245 - this also is not organized in to the correct location
in the file (sort by company name, not append) were it to have proceeded.

Ttoledo28:
@dnsguru done! Anything more?

n8schloss:
Hi @Ttoledo28,

Facebook recently updated our [help center guidance](https://www.facebook.com/business/help/126789292407737?id=1205376682832142) around this and I'd like to specifically call out this section:

> Our current efforts are designed to support preexisting Public Suffix List domain registrations or eTLDs. This support is in line with Apple’s recent Private Click Measurement update. There are other technical implications if a domain is registered as a Public Suffix that a business should consider (for example, the domain that is registered on the Public Suffix List cannot have its own cookies) and as such, we do not recommend that clients register their domains on the Public Suffix List specifically for Facebook event configuration.

When I visit either of the examples you linked to above I see some cookies getting set for *.grupoa.education. Being added to the PSL will break that functionality and given that browsers have varying cadence for updating their copies of the PSL, if you get added and notice breakages then it may be months before you could be removed. On the other hand, I do not see content on www.grupoa.education, searching for '*.grupoa.education' does show quite a few distinct subdomains and it seems like those subdomains are used for distinct entities. So if cookie security is also a desired outcome then this request may make sense. Given that you do have cookies getting set on *.grupoa.education, which will break if you get added to the PSL, before continuing with this request please make sure to read Facebook's updated guidance and make sure that you understand the full implications of being added to the PSL.

Ttoledo28:
Hi @n8schloss.

I read all the information available on facebook DOC and before making a decision, I would like to understand a point:

When I add the domain 'www.grupoa.education' on PSL, that domain will not be able to have its own cookies and my other subdomains '* .grupoa.education' will not be able to share cookies between them (they must work independently), this is correct?

n8schloss:
@Ttoledo28,

> When I add the domain 'www.grupoa.education' on PSL, that domain will not be able to have its own cookies and my other subdomains '* .grupoa.education' will not be able to share cookies between them (they must work independently), this is correct?

That's correct. If the domain get's added to the PSL once browsers update their copy of the list, which can take a while in some cases, then cookies would not work on grupoa.education and cookies on *.grupoa.education could not be shared between the subdomains.

Ttoledo28:
@n8schloss Thank you for your help.

I opened #1305, where we set up an alternative route, with all the necessary specificities and we will use it to send events to the pixel.

PR #1301, opened 2021-04-27
Add coachy.net

dennisspohr:

Description of Organization
====
Coachy is a membership tool. Each of our coaches/clients get their own subdomain (customname.coachy.net)
Organization Website: https://www.coachy.net

I'm Dennis and the owner/developer of Coachy International Ltd, running coachy.net.

Reason for PSL Inclusion
====

We provision all new clients with a *.coachy.net subdomain. Because of this, requests between companies using *.coachy.net domains are considered to be same site, leaving them open to CSRF.

Our domain is not set to expire until 2023.

DNS Verification via dig
=======

```
dig +short TXT _psl.coachy.net
points to this URL
```

make test
=========

Yes I ran the test

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -13675,4 +13675,8 @@ basicserver.io
 virtualserver.io
 enterprisecloud.nu
 
+// Coachy : https://www.coachy.net
+// Submitted by Dennis Spohr <support@coachy.net>
+*.coachy.net
+
 // ===END PRIVATE DOMAINS===


sleevi:
This doesn't actually do what's intended. You would need to drop the wildcard, based on the examples provided. However, that will also affect cookies and potential navigation to `coachy.net`. Are you sure that's what you intended?

dennisspohr:
To be honest I don't completely understand. It's important that each subdomain is handled as an own domain for each client.
We are also having troubles with Facebook Advertising, because their don't know that the subdomains are unique domains for us. This was she original reason why we've attempted this request. Does this clarify our issue? Thank you!

sleevi:
Thanks. Closing this as it's related to #1245 for Facebook.

The wildcard does not cause each subdomain to be treated as its own, it causes each subdomain of each subdomain. Putting `coachy.net` would be the correct entry, but that would then prevent your cookies (including what looks like your signon cookies) from working.

n8schloss:
Hi @dennisspohr,

Facebook recently updated our [help center guidance](https://www.facebook.com/business/help/126789292407737?id=1205376682832142) around this and I'd like to specifically call out this section:

> Our current efforts are designed to support preexisting Public Suffix List domain registrations or eTLDs. This support is in line with Apple’s recent Private Click Measurement update. There are other technical implications if a domain is registered as a Public Suffix that a business should consider (for example, the domain that is registered on the Public Suffix List cannot have its own cookies) and as such, we do not recommend that clients register their domains on the Public Suffix List specifically for Facebook event configuration.

When I visit https://www.coachy.net I see lots of content and get cookies set for the coachy.net domain. Because there's a full site at that domain I'm not sure if being added to the PSL is a solution here. Being added to the PSL will break a lot of the functionality on that page and given that browsers have varying cadence for updating their copies of the PSL, if you get added and notice breakages then it may be months before you could be removed.

PR #1300, opened 2021-04-27
Update public_suffix_list.dat for FAESA

Ttoledo28:
* [X] Description of Organization
* [X] Reason for PSL Inclusion
* [ ] DNS verification via dig
* [X] Run Syntax Checker (make test)

* [ ] Each domain listed in the PRIVATE section has and shall maintain at least two years remaining on registration.

Description of Organization
====FAESA is a Brazilian college with undergraduate and graduate courses

Organization Website: https://www.faesa.br/

 My name is Tayro, I'm a marketing coordinator at Grupo A. We are a company in the education segment. 
FAESA is a partner institution of which we share domains and subdomains:
FAESA domain: https://www.faesa.br/ 
Grupo A subdomain: https://ead.faesa.br/

Reason for PSL Inclusion
====FAESA is a partner college of the company A + Educação, which shares subdomains. We would like to add this domain to the PSL to use Facebook's PIXEL events in campaigns

#1296 
#1245 

DNS Verification via dig
=======

make test
=========
dig +short TXT _psl.faesa.br
"https://github.com/publicsuffix/list/pull/1300

PR #1305, opened 2021-05-01
Add hubpixel.info for A+ Educação

Ttoledo28:

Description of Organization
====

Organization Website: https://www.grupoa.com.br/ 

 My name is Tayro, I'm a marketing coordinator at Grupo A. We are a company in the education segment

Reason for PSL Inclusion
==== 
We created an alternative route for use in the pixel for IOS 14.

We will create unlinked subdomains for our partner institutions in order to be able to set up events

 
#1245 
#1296 
#1300 

DNS Verification via dig
=======

make test
=========

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -10640,6 +10640,10 @@ ltd.ua
 // 611coin : https://611project.org/
 611.to
 
+// A+ Educação : https://www.grupoa.com.br/
+// Submitted by Tayro Toledo <ttoledo@maiscampus.com.br>
+*.hubpixel.info
+
 // Aaron Marais' Gitlab pages: https://lab.aaronleem.co.za
 // Submitted by Aaron Marais <its_me@aaronleem.co.za>
 graphox.us


n8schloss:
Hi @Ttoledo28, unfortunately given that there are no full websites at *.hubpixel.info it does not make sense to add it to the PSL. The PSL is not intended as a way to work around the limits of AEM or PCM and as such adding a website that only hosts the FB pixel to PSL is not a valid use case. What you were doing in #1296 was the best route, since that domain was hosting subdomains for distinct organizations and each subdomain contained a full website that's displayed to users.

Ttoledo28:
Hi @n8schloss 

We decided not to proceed with the inclusion of the subdomain * .grupoa.education because we have addresses that run dependent and independently on that address.

What we are doing is directing the domains of the different institutions to domains * hubpixel.info (we are creating these pages yet). EX: 
https://uniarp.hubpixel.info/orbita/subscription/
https://pucpr.hubpixel.info/orbita/subscription/

In that case, can't we include our address in the PSL?

n8schloss:
@Ttoledo28, can you elaborate how this is different from #1296 then? When will hubpixel.info be used and when will grupoa.education be used?

Ttoledo28:
@n8schloss 
We will transfer our Facebook campaigns to the * .hubpixel.info domains, where we will be able to set up different conversions for each institution separately.

The other domain #1296 will continue to work for registration on the site or Google campaigns (although they are separate domains, we were unable to leave them as PSL, as they have other companies at that address).

n8schloss:
@Ttoledo28, so ads on Facebook will use *.hubpixel.info but from all other places you'll use *.grupoa.education? Given that this domain is just for Facebook ads purposes I'm not sure if it's a valid use case for being added to the PSL.

I appreciate that you're in a tough situation here and I wish we had an easier answer for you. Of the requests you've made I think #1296 is the most likely to be approved, but I understand that there's a real tradeoff you'll need to make. Unfortunately this is the situation as it currently stands, and while we're working on lobbying Apple to implement a solution that will work for for folks in your situation there currently ins't one available.

dnsguru:
Any and all PR that are related to Facebook IOS constraints are not being merged.   This will sit here as a wontfix.

We gain no joy from this situation. The reason we reject requests that are attempting to use the PSL as workaround for limitations or security purposes because browsers and other software / services / integrations that rely ***voluntarily*** on the PSL - and could stop doing so if it is seen to be abused as a means to get around them.

dnsguru:
I will add that there is engaged and open dialog with Facebook on the matter, as volunteer time allows.  We're currently behind due to the the higher than typical number of requests.

Ttoledo28:
Hey guys!

Just so I understand then, to include my domain in the PSL, I would have to stop using the * .grupoa.education domain and move the entire application to * .hubpixel.info. Doing this, could the domain be inserted in the PSL?

My domains work separately and, therefore, it would be interesting to include in the PSL, except that I have a part of the architecture that I cannot remove from the * .grupoa.education domain. If I start using * .hubpixel.info as the main one and leave only the essentials in the old one, would you authorize it?

n8schloss:
Posting to let you know that Facebook has updated our guidance to include a validation flow run by Facebook staff. We encourage anyone requesting addition to the PSL as it relates to Facebook and iOS 14 interactions to make a request directly to Facebook via the form linked from our Help Center Article: https://www.facebook.com/business/help/126789292407737  

In the form, given that you have already raised a pull request here, please include a link to this pull request in the description.

We have agreed to do this to reduce the burden on PSL maintainers, ensure that you are asking to be on the PSL for the right reasons, and to allow us to collate and understand the issues that businesses are facing as part of this shift. 

After Facebook has assessed one of these requests we will make a recommendation or no recommendation decision, the PSL maintainers will make the final determination and we will coordinate with them on next steps accordingly.

dnsguru:
excellent @n8schloss closing this, reference this pr# should a new one be requested

PR #1290, opened 2021-04-20
Add github dev domains

dmilburn:
* [x] Description of Organization
* [x] Reason for PSL Inclusion
* [x] DNS verification via dig
* [x] Run Syntax Checker (make test)

* [x] Each domain listed in the PRIVATE section has and shall maintain at least two years remaining on registration.

Description of Organization
====

Organization Website: github.com (though this is for github.dev and githubpreview.dev)

I'm a software engineer at GitHub working on the new Codespaces product

Reason for PSL Inclusion
====

We want to restrict the ability to create cookies that span subdomains of github.dev and githubpreview.dev 

Both of these domains aren't set to expire until 2023:
```
$ whois github.dev|grep -i expir
Registry Expiry Date: 2023-01-16T16:27:37Z
$ whois githubpreview.dev|grep -i expir
Registry Expiry Date: 2022-12-09T21:03:15Z
```

DNS Verification via dig
=======

`dig +short TXT _psl.github.dev` and `dig +short TXT _psl.githubpreview.dev` both point to this URL

make test
=========

I wasn't able to get the tests to run locally but they do pass in CI. Let me know if this isn't sufficient!

Issue #1319, opened 2021-05-12
Proposal:  Deprioritization of subsequent rollbacks/mod PRs received within 60 days of add PR


dnsguru:
I propose we deprioritize rollback requests that come within 60 days of a PR to place them at the absolute back of the line.

One of the lead contributions to wasted volunteer cycles and requestor frustration generally is where there is a request made to add an entry, and then iterations of updates or rollbacks when the request does not behave as expected.

The rollbacks or updates typically take many cycles (can sometimes be months/years/never) to propogate and flow out to whatever is using the PSL on the timeframe that they flow at - which the PSL volunteers have no control over. (See [Propagation / Expectations](https://github.com/publicsuffix/list/wiki/Guidelines#appropriate-expectations-on-derivative-propagation-use-or-inclusion) for better understanding).

Sure, casual requests work the willing and abuse volunteer cycles, which seem free to the requestors, but request/rollbacks are rude and wasteful as they rob productive cycles from other PR or things the project generally needs that slack time can sometimes allow us to accomplish.  

In order to ensure that the requestor on a PR has really taken a sober and careful review of the impact that their change will have, it is important to understand that there will be a lower priority given to rollbacks against current add requests.  

This is a necessity and consequence of the increase in volunteer cycles lost to casual requests.  The majority of those have been triggered by changes to some interop between tracking vendors and operating systems' privacy handling altering some projects.

dnsguru:
Revised.  

**tl;dr:** Rollbacks will have +60days waiting period, minimum.

With the thundering herd of higher effort requests that IOS 14.5 has granted the volunteers on the PSL, and the iteration overhead of all the workaround guidance in the wild that is wrong and needs interaction with angry affected folk, the resource burden on folks donating their time is out of hand.  In order to help folks not casually request workarounds after lengthy interactions only to subsequently submit rollbacks, there is a need to impose a 60 day wait on rollbacks of PRs.  

Get it right in the first inquiry.  Get help from your Facebook rep or provider if you are not sure.

bedfordsean:
Facebook will be standing up an inbound process very shortly, which should be your first port of call (not GitHub). I'll update here with a link to our documentation on how to go about raising a PSL request where Facebook will validate that request soon

dnsguru:
>Facebook will be standing up an inbound process very shortly.

[...]

PR #1335, opened 2021-05-27
Remove github.dev from PSL

josebalius:
* [x] Reason for PSL Exclusion

Description of Organization
====

Organization Website: github.com

I'm a software engineer at GitHub working on the Codespaces product.

Reason for PSL Exclusion
====

We no longer want `github.dev` to be included in the PSL. The original intent for the addition was predicting several use cases that are no longer valid for our plans with `github.dev`. We have some initiatives that require removal.

DNS Verification via dig
=======

Do I still need to update DNS to this PR even though it is a removal? You can see the current record points to the PR where we added the domain: https://github.com/publicsuffix/list/pull/1290

make test
=========

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -11953,7 +11953,6 @@ gsj.bz
 // GitHub, Inc.
 // Submitted by Patrick Toomey <security@github.com>
 githubusercontent.com
-github.dev
 githubpreview.dev
 github.io
 


dnsguru:
>Do I still need to update DNS to this PR even though it is a removal? You can see the current record points to the PR where we added the domain: #1290

Not needed if prior submitter = update/removal submitter, but is needed where these are not same party.  Because this was submitted by @dmilburn w a 'bless' by @ptoomey3 we'd need the DNS _PSL txt to match this PR or both their thumbs up on this. 

See #1319 which is under proposal to help parties submitting PR to really consider their initial request to not disrupt their domain in the first place.

Just want to note, there is not a predictable (or controllable) timeline for when downstream PSL integrators/users/projects each update.  We really discourage (non-technical term) 'willy-nilly' add-remove as it can be disruptive to the requestor's affected domain(s), but also it is net double tax on disposible volunteer cycles vs taking no action at all to begin with.

Also, please add the WHY to the reason for excllusion.   We appreciate where parties submitting their rationale with some details as to what promoted the need to roll back the requested PR, as it builds documentation of things that people can avoid doing in the first place if at all possible.

josebalius:
Thanks @dnsguru . They are both aware of this change, so I'm sure I can get their approval on this.

I'll update the OP with the rationale. We are also aware of the unpredictable timeline with a change like this and have discussed it at length. Ultimately we think this is the right move, and thank you for raising it.

dnsguru:
No doubt you are right they will provide input or a thumbs up.  Will watch for their feedback.

That expanded rationale will help others avoid whatever was attempted and didn't work for github in this instance so describing what the initial request had hoped to acchieve, how the result differed from that once the PSL entry was present, and what unexpected outcomes that were bad.   Anything you can furnish to help others not make the same futile attempt is good for the community.

dnsguru:
``` We no longer want github.dev to be included in the PSL. The original intent for the addition was predicting several use cases that are no longer valid for our plans with github.dev. We have some initiatives that require removal. ```

That added more words but remains unhelpful to others because it was equally vague

dnsguru:
Not seeking to be punative for wasting my time, but was hoping we could get something better as a means to help others as opposed to "because: our reason", as it has zero real value - the primary objective is in reduction of the onslaught from the IOS/FB thundering herd... if we can point at things that outline requests that can be avoided and why.

dnsguru:
So this doesn't rollback #1290, just github.dev.  githubpreview.dev remains, correct?

josebalius:
`githubpreview.dev` remains correct.

ptoomey3 (approved):
👍 - Also wanted to acknowledge that removing something from the PSL doesn't guarantee is't removal from anything downstream of the list.

dnsguru (approved):
submitter ack'd rollback.

dnsguru:
@ptoomey3 @josebalius @dmilburn THX for ACKS 

Since this project is currently crashing on your gitcouch... and technically #1319 is in discussion, merging this now

PR #330, opened 2016-11-14
Added enonic.io domain to the private list

drerik:
Enonic is using Enonic.io as a base for out Cloud hosting service, and every customer project is delegated its own domain name under enonic.io and customer.enonic.io. To get cookie handling rignt and to circumvent lets encrypts rate limits, we need to add enonic.io and customer.enonic.io to the public suffix list.

PR #1399, opened 2021-08-24
these domains are being deprecated, we wish to clean up after ourselves

diegopasten:
### Info ### 
Years ago we needed these addresses for our customers. We have moved away from these domains and are able to what we need with a new setup and new domain.

We are trying to clean all up so we don't leave anything in a broken state.

### Question ### 

Do I need to remove the TXT entry on our end first? 
```
dig +short TXT _psl.customer.enonic.io. 
"https://github.com/publicsuffix/list/pull/330"
```

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -11676,11 +11676,6 @@ staging.onred.one
 // Submitted by Jacob Bunk Nielsen <jbn@one.com>
 service.one
 
-// Enonic : http://enonic.com/
-// Submitted by Erik Kaareng-Sunde <esu@enonic.com>
-enonic.io
-customer.enonic.io
-
 // EU.org https://eu.org/
 // Submitted by Pierre Beyssac <hostmaster@eu.org>
 eu.org


dnsguru:
@drerik is this a legit / approved change to the PR you'd submitted?

dnsguru:
Appreciate where folks clean up after themselves ; want to get the prior submitter to acknowledge the change

diegopasten:
He quit the company years ago. I am leaving very soon. Not sure if anybody else here will know how to clean it up. Can I prove control by removing the DNS-entry or by some other mean ?

diegopasten:
Actually. I will just remove them now. Then it's up to you/drerik to solve it on this end.

diegopasten:
All right, the entries have been removed from enonic.io and customer.enonic.io

```
dap:~/ $ dig +short TXT _psl.customer.enonic.io.                     [12:15:32]
dap:~/ $ dig +short TXT _psl.enonic.io.                              [12:30:27]
dap:~/ $                                                             [12:30:33]
```

Up to you guys to solve the rest, but I'm not sure if drerik's account is active. Could be a deadlock.

dnsguru:
Ah, that works.    #smh I should have thought of that or having you update the PR# to this one.

The objective with all things is demonstrating control, as it should be difficult for a third party to make modifications that could alter entries that they do not control in order to keep things trusted.

Anyway I will approve this .... Because there was a test (I did but didn't document here in the thread) to lookup the _PSL entry and the condition changed (verifying your documented dig results were what I also experienced before and after) in the manner you asserted and documented, that creates fair connective tissue between your administrative oversight, the changes in this PR and the initial entries.

dnsguru (approved):
Submitter different than initial section creator, BUT the submitter was able to demonstrate administrative control by altering the state of the _PSL entries by removing them from the zone.

APPROVED

dnsguru:
Also see #330

PR #1393, opened 2021-08-06
add diher.solutions and rss.my.id to the list

pereceh:
add a new public suffix for Diher Solutions to use by yourself. everyone can use the domains I added for informational purposes.

* [x] Description of Organization
* [x] Reason for PSL Inclusion
* [x] DNS verification via dig
* [x] Run Syntax Checker (make test)

* [x] Each domain listed in the PRIVATE section has and shall maintain at least two years remaining on registration, and we shall keep the _PSL txt record in place

__Submitter affirms the following:__ 
  * [x] We are listing any third party limits that we seek to work around in our rationale such as those between IOS 14.5+ and Facebook (see [Issue #1245](https://github.com/publicsuffix/list/issues/1245) as a well-documented example)
  * [x] This request was _not_ submitted with the objective of working around other third party limits
  * [x] The [Guidelines](https://github.com/publicsuffix/list/wiki/Guidelines) were carefully _read_ and _understood_, and this request conforms
  * [x] The submission follows the [guidelines](https://github.com/publicsuffix/list/wiki/Format) on formatting

---
__For Private section requests that are submitting entries for domains that match their organization website's primary domain:__

``` 
Seriously, carefully read the downline flow of the PSL and the guidelines.
Your request could very likely alter the cookie and certificate (as well as other) behaviours on your 
core domain name in ways that could be problematic for your business.

Rollback is really not predicatable, as those who use or incorporate the PSL do what they do, and when.
It is not within the PSL volunteers' control to do anything about that.  

The volunteers are busy with new requests, and rollbacks are lowest priority, so if something gets broken 
it will stay that way for an indefinitely long while.
```
(Link: [about propogation/expectations](https://github.com/publicsuffix/list/wiki/Guidelines#appropriate-expectations-on-derivative-propagation-use-or-inclusion))

 * [x] Yes, I understand.  I could break my organization's website cookies etc. and the rollback timing, etc is acceptable.  Proceed.
---

Description of Organization
====

Organization Website:  https://rss.my.id

I am a blog writer. and I'm also the sole manager for the listed domains.

Reason for PSL Inclusion
====

I'm planning to automate the provision of a destination subdomain in place of an address on a web hosting server. my purpose of implementing this is for Cookie Security, as different customers will have control over the subdomain.

The rss.my.id domain is on a 3-year renewal and currently registered until 2024-07-24,

and diher.solutions domain is on a 9-year renewal and currently registered until 2031-03-15.

DNS Verification via dig
=======

```
[...]

PR #1406, opened 2021-09-02
Roll back diher.solutions PR #1393

dnsguru:
foo.bar and *.foo.bar can't both be added, and I let #1393 through, where the linter typically catches and fails them.

It subsequently causes failed imports on some browser inclusions, so this must be backed out.

@pereceh you must re-sumbit and choose only one of each, please reference #1393 in the new PR - leaving the prior _psl txt record is fine when you re-submit 

// Diher Solutions : https://diher.solutions
// Submitted by Didi Hermawan <mail@diher.solutions>
rss.my.id
*.rss.my.id
diher.solutions
*.diher.solutions

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -11258,13 +11258,6 @@ deno-staging.dev
 // Submitted by Peter Thomassen <peter@desec.io>
 dedyn.io
 
-// Diher Solutions : https://diher.solutions
-// Submitted by Didi Hermawan <mail@diher.solutions>
-rss.my.id
-*.rss.my.id
-diher.solutions
-*.diher.solutions
-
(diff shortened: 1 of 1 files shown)


sleevi (approved):
For Chrome, at least, `*.rss.my.id` would also encompass `rss.my.id`, although my understanding is for Firefox, this isn't the case (but Firefox doesn't support listing both).

Setting this Approved, though, and I think you're right to address this through a new submission.

pereceh:
@dnsguru @sleevi i will use *.diher.solutions and *.rss.my.id ,,, then delete diher.solutions and rss.my.id  from public_suffix_list.dat

is this the correct method so that all match ?

dnsguru (commented):
Revised to submitter preferences

Issue #1349, opened 2021-06-09
Add a generic public-service namespace suffix for sandbox testing / lab use


sirdarckcat:
It happens (apparently often) that people request a public suffix to the PSL (see #1272 and #1241) for a use-case of a one-time "lab" type of usecase (eg, trigger behavior that browsers/clients/libraries/servers use the PSL for). This is particularly important for some security applications because "Site" is a first-level concept of the browser security model (so, subdomains don't work).

The proposal would be to have a domain name (name TBD) added to the PSL of the form: `*.psllabs.com` where a custom DNS server is running on `something.com` that returns to SOA requests with the value of the domain being requested. For example `1-2-3-4.psllabs.com` could respond with `1.2.3.4` (this is an example).

The cost of running a service like this would be modest, and a large company could sponsor it (I could seek Google to run this), and the advantage would be that people that just need one-off PSL additions can use this service.

I can experiment with something similar to this (there seems to be expired suffixes in the PSL I could use), but it probably makes sense to agree to do this first.

dnsguru:
Is this kind of like a sandbox type system ?
Wonder if it is worth getting something underarpa via the IETF formally?

sirdarckcat:
it's kindof like a sandbox, yes for testing, development, demos, etc.. the 2 PRs I showed are for "demo vulnerable apps" and how they would behave with different browsers.

I'm not sure what's the process with .arpa - but yea, I guess something like `in-addr.arpa` but that resolves to SOA instead of A

dnsguru:
@sirdarckcat Would you mind typing up something that I could review with the ICANN Office of the CTO so that we can get their recommendation ?  I believe you are largely describing that there be the creation of some form of a general purpose namespace solve for the lab user types who want to poke around with example domains to test cookie horizons, DOM, etc.

I like this idea because it allows for the type of CTF or lab scenarious that help make things safer for users, and the proposed concept preserves the ambition of filesize modesty.

sirdarckcat:
Would the idea be that ARPA would run it? Or they would just provide the TLD.

Anyway, I can write something up, although what format and how detailed does it need to be? Is this a one-pager or a RFC type of thing.

dnsguru:
>arpa

Some undelegated name within .arpa or peraps example.com - I don't want to
constrain or limit the idea early on.

There are other proposals heading to or under discussion within IETF that
this might mix in with.

I vaguely recall a proposal that created some rfc1918-ish private test
[...]

PR #1409, opened 2021-09-07
Add test-domains.org to private domains

rerudiger:

  * [x] We are listing any third party limits that we seek to work around in our rationale such as those between IOS 14.5+ and Facebook (see [Issue #1245](https://github.com/publicsuffix/list/issues/1245) as a well-documented example)
  * [x] This request was _not_ submitted with the objective of working around other third party limits

Description of Organization
====

Organization Website: https://re-creation.co.jp

We're a website and web application development company from Japan.

Reason for PSL Inclusion
====

The domain is mostly used internal, but also shared with some contractors with servers not under our control. Projects under the domain are completely independent, so they should not be able to share cookies.
Domain is in use since 2013 and registered until 2026, with no plans to drop it.

DNS Verification via dig
=======

```
dig +short TXT _psl.test-domains.org
"https://github.com/publicsuffix/list/pull/1409"
```

make test
=========

Tests are all green.

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -13163,6 +13163,10 @@ g.vbrplsbx.io
 // Submitted by David Fischer <team@readthedocs.org>
 readthedocs.io
 
+// Re:creation Corp Inc : https://re-creation.co.jp
+// Submitted by Jonas Rüdiger <rudiger@re-creation.co.jp>
+test-domains.org
+
 // Red Hat, Inc. OpenShift : https://openshift.redhat.com/
 // Submitted by Tim Kramer <tkramer@rhcloud.com>
 rhcloud.com


sleevi:
We don't really accept requests for internal servers, for the reasons mentioned at https://github.com/publicsuffix/list/wiki/Guidelines#validation-and-non-acceptance-factors

rerudiger:
@sleevi I appreciate your time investment into this issue, I'm sorry to have to bother you again.
Please let me explain the nature of these domains.

These servers are NOT for internal use. I must admit, the name is a bit misleading, but these servers are manly staging servers and servers for collaboration. They are often viewed by customers, not only by our staff. Project A should not be able to set cookies for Project B, as they have no connection besides being related to our company.

I carefully read the guidelines, and as far as I can see, the only valid point to deny this request is "projects that are smaller in scale". I apologize that we're a small company, but isn't the Internet made to work that way? Would you be so kind to explain to me how I am supposed to fix this issue without your support.

sleevi:
@rerudiger Thanks for the further context.

As it relates to the sort of “Dev” and “Staging” servers, I hope you can understand why it’s a bit unreasonable to ask every Internet client out there to carry code to test to see if they’re accessing one of your staging servers, which is what this sort of request comes down to.

It’s not that you’re a small company, but rather, the utility of having every user have to download (potentially many dozens of times) the entries here, to know about your staging servers.

Understandably, when a request is closed, it’s entirely reasonable to ask “What am I supposed to do, then?” One clear option is simply register different domains for your different customers, and no modifications to the PSL are needed. It might be tempting to say “but this would cost me $1/customer/year” - but the cost of the addition here could easily represent tens or hundreds of thousands of dollars in bandwidth costs, spread across billions of users of the PSL. That’s what the Guidelines are referencing, by trying to gauge potential impact and cost.

If you’re a hosting site with tens of thousands or millions of users (e.g. GitHub), then the equities clearly make sense. But if you’re looking to host dozens or a few hundred sites, or that maybe a thousand users will ever need to know about this separation, then it’s less clear.

This is why the remarks about “staging” don’t really seem to inspire confidence, but am I missing something?

rerudiger:
@sleevi  Thanks for your reply!

To be honest, I am also skeptical of using a centralized list, instead of some decentralized mechanism, how it *ought to be* in the Internet. I'm fully aware that the current method doesn't scale well, but this is not really my fault, is it? And there is no alternative. My line of thinking was: the HSTS Preload List is also huge, and nobody complains, so somehow this should be all right too.

I'm happy to hear that you do not intend discriminate against smaller players and that your reason for refusal is in the nature of staging/collaboration, which is somewhere in the area between public and private.

We won't/can't change the domains, there about 80 domains under test-domains.org, but I guess we can fix security somewhat by prefixing every cookie with __Host-, at least for the projects we fully control. But Firefox will still behave mostly broken (incorrect URL bar, incorrect history, incorrect passwords…). I wish there was a better solution…

Anyway, I'd appreciate if you could clarify your guidelines. Please state that you will refuse anything that is not 100% public, like internet-facing staging servers and semi-private collaborations or domains that don't have at least a thousand subdomains or so.

dnsguru:
Thanks for understanding.  And good suggestion about guidance.

I will look at updating the guidance about low acceptance likelihood on
lab / dev / internal / staging entries or those that have <1000s users so
this is more clear about respecting filesize/bandwidth modesty objectives.

sleevi:
@rerudiger It looks like @dnsguru is (as always) fantastically on top of the guidelines bit. With respect to the static list being bad, as a maintainer, I'm in [violent agreement with you](https://github.com/sleevi/psl-problems) :)

With respect to your specific use case, it'd be worth understanding how much (or not at all) the discussion in https://github.com/publicsuffix/list/issues/1349 would be applicable to your case here.

PR #832, opened 2019-06-01
Add nom.tr

ilhansubasi:
There are 51 .nom.tr domains https://nic.tr/index.php?USRACTN=STATISTICS

PR #1410, opened 2021-09-07
Remove .nc.tr

ilhanyumer:
As sleevi has pointed out we should follow the official policies https://github.com/publicsuffix/list/pull/832#issuecomment-507376281

The document https://www.nic.tr/forms/eng/policies.pdf shows that there is no reserve for .nc.tr and thus .nc.tr should be removed.

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -6261,10 +6261,6 @@ tel.tr
 tsk.tr
 tv.tr
 web.tr
-// Used by Northern Cyprus
-nc.tr
-// Used by government agencies of Northern Cyprus
-gov.nc.tr
 
 // tt : http://www.nic.tt/
 tt


dnsguru:
@ilhanyumer some ccTLD registries have entries like this deliberately.  This happens where there may have previously been namespace like this which were previously operated, and the namespace entry is left in the PSL intentionally in order to not disrupt the existing legacy uses.  Which appears to be the point wisely made in #832.

To proceed, we would need a _psl txt record added into the nic.tr zone or the nc.tr zone, and I would be willing to revisit #832 with a representative of the .tr registry should they engage here.  As this project is volunteer-based, we're not resourced to do outbound calls, but I make it a point to make presentations as frequently as possible about the PSL to various ccTLD gatherings in order to encourage they are engaged on keeping their information up to date.

Your contributions are appreciated, and I wonder if you could interact with the nic.tr to review this section and validate who they are with us to ensure that the PSL reflects an intentional representation of their zone.

PR #1415, opened 2021-09-10
Update rss.my.id and diher.solutions

pereceh:
As per reference [#1406](https://github.com/publicsuffix/list/pull/1406) , i want to remove wildcard characters for rss.my.id and diher.solutions to make them look better. I also used the previous _psl txt record [#1393](https://github.com/publicsuffix/list/pull/1393) .

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -11260,8 +11260,8 @@ dedyn.io
 
 // Diher Solutions : https://diher.solutions
 // Submitted by Didi Hermawan <mail@diher.solutions>
-*.rss.my.id
-*.diher.solutions
+rss.my.id
+diher.solutions
 
 // DNS Africa Ltd https://dns.business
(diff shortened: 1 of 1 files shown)


dnsguru (approved):
Bugfix; Previously vetted
APPROVED

dnsguru:
@pereceh I rushed this one through - I see that the timing of when I modified / merged the last one with the change and when your message clarifying which to keep of the two conflicting (*. and not) entries per name in the thread happened right at around the same time and we got our wires crossed.

pereceh:
> @pereceh I rushed this one through - I see that the timing of when I modified / merged the last one with the change and when your message clarifying which to keep of the two conflicting (*. and not) entries per name in the thread happened right at around the same time and we got our wires crossed.

wow looks like a mistake i modified the branch... i want it without the wildcard (*). so the master branch should be:
rss.my.id
diher.solutions

please help apply this patch to master branch @dnsguru .

PR #1107, opened 2020-09-22
Adds translate.goog and tr-test.goog

sicter:
* [x] Description of Organization
* [x] Reason for PSL Inclusion
* [x] DNS verification via dig
* [x] Run Syntax Checker (make test)

Description of Organization
====

Organization Website: https://translate.google.com/intl/en/about/

Engineer at Google on the Translate team

Reason for PSL Inclusion
====

.translate.goog will be serving 3rd party code separated by publisher. This is very similar to (but a different use case than) appspot.com. Currently there is no much public usage, but since it takes a while to get into the PSL, we would like to have it available before wider use.

DNS Verification via dig
=======

```
dig +short TXT _psl.translate.goog
"https://github.com/publicsuffix/list/pull/1107"
```
```
dig +short TXT _psl.tr-test.goog
"https://github.com/publicsuffix/list/pull/1107"
```

make test
=========

All tests passed.
```
============================================================================
Testsuite summary for libpsl 0.21.1
============================================================================
# TOTAL: 3
# PASS:  3
# SKIP:  0
# XFAIL: 0
# FAIL:  0
# XPASS: 0
# ERROR: 0
============================================================================

============================================================================
Testsuite summary for libpsl 0.21.1
============================================================================
# TOTAL: 5
# PASS:  5
# SKIP:  0
# XFAIL: 0
# FAIL:  0
# XPASS: 0
# ERROR: 0
============================================================================
```

PR #1417, opened 2021-09-10
Add sandbox.usercontent.goog

terjanq:

* [X] Each domain listed in the PRIVATE section has and shall maintain at least two years remaining on registration, and we shall keep the _PSL txt record in place

  * [X] We are listing any third party limits that we seek to work around in our rationale such as those between IOS 14.5+ and Facebook (see [Issue #1245](https://github.com/publicsuffix/list/issues/1245) as a well-documented example)
  * [X] This request was _not_ submitted with the objective of working around other third party limits
  * [X] The submission follows the [guidelines](https://github.com/publicsuffix/list/wiki/Format) on formatting

---

``` 
Seriously, carefully read the downline flow of the PSL and the guidelines.
Your request could very likely alter the cookie and certificate (as well as other) behaviours on your 
core domain name in ways that could be problematic for your business.

Rollback is really not predicatable, as those who use or incorporate the PSL do what they do, and when.
It is not within the PSL volunteers' control to do anything about that.  

The volunteers are busy with new requests, and rollbacks are lowest priority, so if something gets broken 
it will stay that way for an indefinitely long while.
```

 * [X] Yes, I understand.  I could break my organization's website cookies etc. and the rollback timing, etc is acceptable.  Proceed.
---

Description of Organization
====

Organization Website: https://www.google.com

Information Security Engineer at Google

Reason for PSL Inclusion
====

.sandbox.usercontent.goog will be serving 3rd party code separated by publisher. This is very similar to (but a different use case than) .translate.goog (https://github.com/publicsuffix/list/pull/1107). Currently there is no public usage, but since it takes a while to get into the PSL, we would like to have it available before wider use.

*Note: sandbox.usercontent.goog is not an internal/test-only service. In this case "sandbox" refers to code that is isolated akin to "CSP: sandbox".*

The domain is registered for 2 years and will maintain more than 1 year term thereafter.

```
Name: usercontent.goog
Registry Domain ID: 46D7EB740-GOOG
Domain Status:
clientDeleteProhibited
clientTransferProhibited
clientUpdateProhibited
Nameservers:
ns1.googledomains.com
ns2.googledomains.com
ns3.googledomains.com
ns4.googledomains.comDates
Registry Expiration: 2024-06-25 18:47:03 UTC
Updated: 2021-09-13 17:19:33 UTC
Created: 2021-06-25 18:47:03 UTC
```

DNS Verification via dig
=======
```
dig +short TXT _psl.usercontent.goog
"https://github.com/publicsuffix/list/pull/1417"
                                                             
dig +short TXT _psl.sandbox.usercontent.goog
"https://github.com/publicsuffix/list/pull/1417"
```

make test
=========

```
# TOTAL: 3
# PASS:  3
# SKIP:  0
# XFAIL: 0
# FAIL:  0
# XPASS: 0
# ERROR: 0
Making check in tests
  CC       test-is-public-all.o
  CC       test-is-public.o
  CC       test-is-cookie-domain-acceptable.o
  CC       test-is-public-builtin.o
  CC       test-registrable-domain.o
  CCLD     test-is-cookie-domain-acceptable
  CCLD     test-is-public-builtin
  CCLD     test-is-public
  CCLD     test-is-public-all
  CCLD     test-registrable-domain
PASS: test-is-public-builtin
PASS: test-is-public
PASS: test-registrable-domain
PASS: test-is-public-all
# TOTAL: 5
# PASS:  5
# SKIP:  0
# XFAIL: 0
# FAIL:  0
# XPASS: 0
# ERROR: 0
Making check in msvc
```

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -12083,6 +12083,8 @@ withyoutube.com
 *.gateway.dev
 cloud.goog
 translate.goog
+usercontent.goog
+sandbox.usercontent.goog
 cloudfunctions.net
 blogspot.ae
 blogspot.al
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -12083,7 +12083,7 @@ withyoutube.com
 *.gateway.dev
 cloud.goog
 translate.goog
-sandbox.usercontent.goog
+*.usercontent.goog
 cloudfunctions.net
 blogspot.ae
 blogspot.al
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -12083,7 +12083,6 @@ withyoutube.com
 *.gateway.dev
 cloud.goog
 translate.goog
-usercontent.goog
 sandbox.usercontent.goog
 cloudfunctions.net
 blogspot.ae


sleevi:
@terjanq This appears to be adding “staging” domains to the PSL, which we’ve fairly consistently rejected. Just because Google calls it’s internal services “sandbox” does not seem like it’s a practical difference.

However, perhaps more importantly, it’s unclear while Google needs to keep adding entries. Why aren’t the existing entries sufficient?

sleevi (changes requested):
Marking “Request Changes” to, at a minimum, remove `sandbox`, but potentially to reject the whole thing.

terjanq:
> @terjanq This appears to be adding “staging” domains to the PSL, which we’ve fairly consistently rejected. Just because Google calls it’s internal services “sandbox” does not seem like it’s a practical difference.

Can you please clarify what you mean by "staging" domains. This is a new product that will serve users' data on `xxx.sandbox.usercontent.goog` and it's required that `xyz.sandbox.usercontent.goog` doesn't affect `zyx.sandbox.usercontent.goog` in any way.  

It's interesting how you come up with the conclusion 'Google calls it’s internal services “sandbox”' but it has nothing to do with Google's internal services. `xxx.sandbox.usercontent.goog` is already publicly accessible, but in order to fully launch the product we need to ensure the mentioned requirement, and which is what PSL is intended for. 

> However, perhaps more importantly, it’s unclear while Google needs to keep adding entries. Why aren’t the existing entries sufficient?

Because it's a new, different product, and can't use solution of already existing ones. 

> Marking “Request Changes” to, at a minimum, remove sandbox, but potentially to reject the whole thing.

Can you please explain why? The priority of the change is to put `sandbox.usercontent.goog` on the PSL. `usercontent.goog` is only put there because as far as I am concerned, `sandbox.usercontent.goog` cannot be a public suffix without `usercontent.goog` being one. If that helps, we can add an entry for `*.usercontent.goog`, instead.

dnsguru:
If I understand what the request is for, this is an anticipated offering and a proactive request to ensure that this would propagate out to the browsers etc before the launch so it is good to go at that point. 

Ryan's concerns about sandbox type requests being declined is something that we have a heighened sensitivity to, as the optics of declining other requets for lab/sandbox/test/dev environment type entries is something that we do to preserve file/bandwidth modesty, and the requestors express high level of scorn over those outcomes.  So the optics are not good on approving one for Google while declining those of other projects, as it demonstrates what could be interpreted as inconsistency and bias - especially to those who have been declined.

That said, there are a number of requests that have been allowed through where there is a production and development environment combined in a request, and this looks more like those types of requests.

I believe that the challenge and difference is that the use of the 3ld/2ld in the same name is more of an issue.  I would be inclined to consider approving this proceeding if 

1. - the sandbox were not a subdomain of usercontent.goog but instead was sandbox-usercontent.goog, and 
2. - more furnished information on the anticipated number of subname users that this pr would enable.

We would need to show consistency in the handling of 2, to where this would likely need to be close to or >1000

terjanq:
Hi @dnsguru, now it is clearer what your concerns are. So as I mentioned, `sanbdox` in the domain part does not refer to domains requested for internal purposes. It literally refers to "sandboxing" users' data. 

As we probably both know, there is no available feature that would make sites `xyz.example.com` and `zyc.example.com` to be treated as `cross-site` domains except for the PSL. Browsers use the relation to decide which protection mode to choose and being `same-site` means that:
- in browsers supporting multi-threading, data will likely end up in the same process opening a way for speculative leaks (e.g. SPECTRE)
- subdomains can manipulate the Cookie Jar of their parent site
- with features such as `document.domain` can directly read data from each other. 

Our product is all about ensuring that a malicious actor can't leak sensitive data of our users. Every user will have a unique origin under `unique.sandbox.usercontent.goog` so we expect the number of subnames to be much higher than 1000 since we have bilions of users. 

You're asking us to re-design the project which seems to be based on one's discrection. Please point me to the specific rule from guidelines that the requested 3ld domain (or the PR) violates. 

____ 

 Answering all the questions from: https://github.com/publicsuffix/list/wiki/Guidelines#validation-and-non-acceptance-factors
 
 > Does the PR have the template filled out completely and clearly in a way that answers the common questions

I believe that is answered now. 

> Why is the PR being requested? Please be detailed, but not so detailed as to lose the reviewer(s) in the weeds.

Same.

> What are example domains' use scenarios? Describe the desired outcome. Enumerate through example subdomains or the name iteself and how these are desired to work as a result of the PR being merged. This is particularly important with respect to wildcards (*) and bangs (!) and ensuring the result will align with the request.

Example has been provided. 

> How many domains are being requested within the PR?
2 (1 if only `*.usercontent.goog` or `sandbox.usercontent.goog` without `usercontent.goog`

> Are the domains being requested for internal purposes (e.g. organization Foo wants their dev/staging servers, which only they use/access) or is this for general utility

No. 

> Basically, does every user shipping the PSL benefit from the added size/processing?

Yes. 

> Does every domain in the PRIVATE section have at least two years remaining in the registration term?

For some reason I though it is registered for 2 years, but I will get it fixed as soon as possible. 

> We do not accept entries for use as DNS wildcards, such that e.g. 1-2-3-4.foo.tld resolves as IP address 1.2.3.4. This basically projects the security properties of the IP address space onto the domain name space, and we don't feel that is safe. IP addresses can be dynamically allocated to multiple mutually-untrusting parties; domain names generally are not.

They point to the same servername. 

> We do not accept entries whose sole purpose is to circumvent Let's Encrypt rate limits. They have a form you can use.

N/A

> We do not accept entries that have the objective of getting around limitations that have been put in place by a vendor to protect internet users. The PSL is not a 'workaround', and Pull Requests that appear to be doing this should expect to be declined. Be thorough and candid with the rationale furnished with the request.

N/A. We aim to increase the protection of our users, not to decrease it. 

> File Size and scale of impact of a request is a consideration. Request modesty is important, and projects that are smaller in scale or are temporary or seasonal in nature will likely be declined.

This is a big project, and as mentioned will potentially have millions of users. 

____

I believe that the process should be transparent and well defined. It feels to me that the reasoning behind pushing us to re-design our solutions does not comply with the written guidelines in any way. As far as I am concerned, we are meeting all the written requirements (except 2 years which will be fixed). That said, I request full transparency about what harm including `sandbox.usercontent.goog` in the PSL poses to the outside world and/or the project. 

As the available compromise we can add `*.usercontent.goog` to the list. As of now, `sandbox.usercontent.goog` is the only use-case of the newly created domain, but in the future we might want to add new subproducts, not necessairly having the need of being put on the PSL. That's why I am more inclined towards having put the straight use-case of `sandbox.usercontent.goog` that indeed needs it.

sleevi:
> Because it's a new, different product, and can't use solutions of already existing ones.

This is the most concerning part of the response, and I’m particularly sensitive to the optics here re: Google (my employer) and Google’s own contributions to the growth of the PSL, and the concerns with unfettered growth.

The reason here is that if the goal is for such SPECTRE resistance, then it seems like any of Google’s _existing_ domains and entries on the PSL can account for this, and there’s no need to add new entries at all. The above answer provided doesn’t actually explain why that’s the case: it just tautologically says Google can’t.

@dnsguru ’s advice about creating a new 2LD wasn’t where I was trying to go. Rather, that `sandbox` as a term is specifically overloaded at Google to be both a name for `pre-production`/`staging`/`internal service` (which we reject) and, as you seem to be suggesting here, `isolated`/`independent`. However, if you’re going for that latter meaning, then it remains unclear why you need two distinct entries, since one entry alone provides it.

Concretely:
1. Google has a lot of domains already on the PSL. Each new request is a new imposition on billions of users, and so each new request by Google will be subjected to increased scrunity.
2. Google can already achieve the technical goals through a myriad number of existing domains. As best I can tell, the main argument at present is related to product goals.
3. Hopefully, by highlighting points 1 and 2, Google can better determine a path forward that can achieve its product goals, while minimizing or eliminating the need to continue to add new entries to the PSL.

> That said, I request full transparency about what harm including `sandbox.usercontent.Goog` in the PSL poses to the outside world and/or to the project.

A demand (given the overall previous message) phrased as a request, of volunteers, is a quick way to demoralize volunteers from wanting to deal with a particular request.

Did you review the information previously provided, such as in  https://github.com/publicsuffix/list/pull/1409#issuecomment-915739303 ? I’m assuming not, but it’s worth highlighting there was already an attempt to explain. The reason is simple: the PSL cannot grow unbounded to simply reflect every domain or subdomain someone may want. This is not sustainable, not tenable with existing implementations. The maintainers try, with the extent of their available time, to reduce the rate of growth and ensure it’s necessary, as the bits and bytes of the list often need to make it to users devices and programs dozens to hundreds of times - an impact to users in terms of cost, and to the world in terms of energy.

It remains unclear that, given the many existing domains that Google has on the list, that any further change is technically necessary. And that’s what this reply is about trying to evaluate and understand.

I think it would be useful if you can concretely explain what technical reason, relevant to the billions of clients that will need to be informed of the distinction and be capable of processing the list thousands to hundreds of thousands of times in the course of normal browsing, why `usercontent.goog` and `translate.goog` need to be different.

dnsguru:
Three inputs on this from me, and some comments.

1: I stepped in on this discussion as a neutral non-Google-employee party
so folks don't think this is the "GSL".  If there are dynamics within
Google that may be opaquely impacting this request, I am ignorant of them.
That may either be a bug or a feature here.

2: Empowered by that ignorance, I am looking at this through the lense of
the request as I would any other request.  I have seen (and have approved)
a number of requests that have projects submitting a combination of
production and test for whatever it is they are asking for.  These, at
least to me, are quite different than a purely lab/experiment/dabble zone,
which are the type of bloat we seek to minimize.  We also, on top of this,
do factor in the estimated scale of the anticipated affected reach and
benefit of the change, and the permanence of it.

3: One of the biggest time/energy/peace of mind drains on the small
volunteer team here is dealing with the entitled, but setting that aside, I
generally don't at all appreciate where the maintainers here are challenged
to defend a rationale for declining something.  Throwing back at @sleevi to
defend a response we use with other declines indicated to me that there is
an experiential awareness and perhaps empathy of the third party
butthurtedness that those who are declined experience, and when considered
in tandem with my point 1, really requires consistency in how our outcomes
and the dialog are documented.  Making an exception for Google requests
where we decline others for the same type of thing demonstrates poor
trusteeship of the resource we pour a lot of volunteer time into.

I dont understand what usercontent.goog is for.  If that is
lab/test/dibbledabble/sandbox stuff that 10 or 15 people will use, this is
a request that will not proceed.

I thought I saw a request for something different from a pure labble-dabble
type, and instead a request for two distinct subdomain eTLD - and suspected
that as worded the might not work as expected or could be phrased better.

Our chief concern is consistency and anti-bloat.

dnsguru:
Using the dialog from this PR and #1409, I have updated the non-acceptance section of the wiki.  Thank you for the help in evolving the wording on this, as the volunteer team on the PSL is quite small.

dnsguru:
I appreciate seeing any innovation with domain names, what strikes me here is how to reconcile where this would be a public thing, because as I read exhibit A of the [.goog registry operator code of conduct exemption request](https://itp.cdn.icann.org/en/files/registry-agreements/goog/goog-ccerrp-18mar14-en.pdf), page 4 states quite clearly in the second paragraph, that, "Members of the public will not be able to register domain names in this new gTLD",  which would mean to me that this is ONLY usable by Google directly to remain in compliance with the Registry Agreement with ICANN.   

If this code of conduce exemption has since been voided or amended to remove that stipulation, I was not able to redily locate where that is documented.

I realize that applies to second level domains, but it is vague and subjective as to if it applies to third or lower.  Meaning, if this is ONLY for use at Google for Google, then this PR would completely be for a private namespace thing, and the request entirely falls inside the same category of the #1409 wontfix rationale.

If this IS indended for third parties, it is an interesting case of using an exemption TLD like this. It might be able to move forward but it may need to reconcile with the manner in which Google has indicated they will operate it to ICANN, which might expire the code of conduct exemption to ensure it does not void the exemption for the TLD.

sirdarckcat:
Hi

Never a bad time to jump in a PSL debate on a weekend :-)

Google has a ton of domains in the PSL. The reason is because we have a ton of products, and every new product that requires hosting user data ends up requesting one.

https://security.googleblog.com/2012/08/content-hosting-for-modern-web.html
https://sites.google.com/site/bughunteruniversity/nonvuln/xss-in-sandbox-domain

This sucks. Wouldn't it be nice if we could unify all future product requests under one suffix?

One idea for such a suffix is `usercontent.goog`. So, translate could use `translate.usercontent.goog` and the successor of FileComp (the internal name for sandboxing of user content at Google) could use `sandbox.usercontent.goog`

Now, we can do that, have a single suffix (`*. usercontent.goog`) to have a Sauron moment and reduce the future requests, under one single suffix.

Or, we can just request a new one every time.

I, for one, like LOTR, so would prefer the wildcard approach, but if Jan should instead ask product teams to request their own subdomains every time, then that is likely the opposite of what you want.

In case it helps, some questions that went unanswered:
1. **How many people will use it?** Well, it will be part of Google's core infrastructure for download serving, so, assuming we can ship it (which this is blocking) and then we deprecate the old version (FileComp), it would be at least GMail and Drive users (that's googleusercontent.com - seems to be the 21 most popular site in the whole universe!!!! https://moz.com/top500 ;-)
2. **How come this will become a public TLD?** It won't. It will allow Google to serve user content. You know, like downloading attachments.

Anyway, I don't have a horse on this race, but I figured that since I've had similar endless debates on the PSL before, I would be able to provide background in a common language, as it seems a lot of the issues on this PR are cause by vocabulary and misunderstandings.

terjanq:
Hey all, 

For startes, I didn't mean to sound demanding, I'm sorry if someone felt attacked. I just wanted to understand what are the concerns of adding new entry to the PSL because I read the guidelines a few times and I couldn't find how adding `sandbox.usercontent.goog` is different from adding `translate.goog` (#1417) or `cloud.goog` (#449) which are already on the list, and were not challenged about theirs product purposes.  

I agree with most of have been said but I still believe that there is misunderstaning and unnecessary assumptions about `sandbox` in the name. Let me answer a few question first then I'll try to draw the bigger picture of our plans. 

___ 
> then it remains unclear why you need two distinct entries, since one entry alone provides it.

`sandbox.usercontent.goog` and `usercontent.goog` are not distinct entries, or at least that's how I understand the PSL works. I would be perfectly fine with only adding `sandbox.usercontent.goog` to the list if not for the fact that `usercontent.goog` is no on the public list and therefore could cause algorithm breakage. But I see that there are examples of not having parent part on the PSL, e.g. `amazonaws.com` so I might be wrong. 

> This is not sustainable, not tenable with existing implementations. The maintainers try, with the extent of their available time, to reduce the rate of growth and ensure it’s necessary, as the bits and bytes of the list often need to make it to users devices and programs dozens to hundreds of times - an impact to users in terms of cost, and to the world in terms of energy.

If the environmental concern is real, there are entries on the PSL with expired domains, products that do not comply with the guidelines, or entries that potentially could be compressed to something like `*.amazonaws.com` (57 entries)

> This is the most concerning part of the response, and I’m particularly sensitive to the optics here re: Google (my employer) and Google’s own contributions to the growth of the PSL, and the concerns with unfettered growth.

Excluding `blogspot` that took 74 entries, Google has 17 entries on the PSL. I think that given the number of products we own the number is very small, and we really try to keep the number low, in fact. 

> Using the dialog from this PR and #1409, I have updated the non-acceptance section of the wiki. Thank you for the help in evolving the wording on this, as the volunteer team on the PSL is quite small.

@dnsguru thank you for updating the guidelines. I also believe that given the impact the PSL has on the Web, both parties should be responsible about the process, because it's our common goal to make the Web more secure, no? 
___ 


Going to the more throughout explaination about technical reasoning. 

Google currently uses `googleusercontent.com` as a shared domain to isolate user provided data from our main product domains, e.g. google.com. Most of Google products use it, but given that it's not on the PSL the user data isn't very safe there. It's prone to speculative leaks and cookie manipulation of parent domain. For example, if malicious user content on `xxxx.googleusercontent.com` finds a way to execute XSS, it can add new cookies on `drive.googleusercontent.com` and hence authenticate malicious user and display attacker's data. 

Our ideal plan is to move all of the products from `googleusercontent.com` to `usercontent.goog` and have it more protected. For that, we need `*.usercontent.goog` to be on the PSL. But, because we have only one product that will pilot the solution I only added `sandbox.usercontent.goog` to not go through all the technical details that didn't seem important based on the acceptance of other PRs, and hard to justify the wildcard based on the recent rejections. 

Just by searching `site:googleusercontent.com` you can see that there are about 1’440’000 results in Google search [query](https://www.google.com/search?q=site:googleusercontent.com). 

That said, I am more than happy to replace both `sandbox.usercontent.goog` and `usercontent.goog` entries with `*.usercontent.goog` if not having `usercontent.goog` there already is not a problem from the algorithm perspective. 

Going abck to sandbox "controvery", tt literally is connected to sandbox solutions alerady existing on the web, for example `CSP: sandbox` or `<iframe sandbox=`, which is about sandboxing potentially malicious content.

terjanq:
We updated the domain for 3 years, I also clarified in the description of the issue that `sandbox` does not refer to internal services. 

Does it make more sense now? Which of these two proposed entries `*.usercontent.goog` and `sandbox.usercontent.goog` look more reasonable to you? 

With `*.usercontent.goog` it's very likely that there won't be many new requests created by Google since we would share the universal domain that can be easily customised for different products, e.g. `drive.usercontent.goog`, or as @sirdarckcat mentioned `translate.usercontent.goog` (which I don't believe we can roll-back now).

terjanq:
@sleevi @dnsguru is there anything else for me to do here? I replaced the record with `*.usercontent.goog` to reduce the 'bloat' that you mentioned in your responses.

sleevi (approved):
Sorry, this slipped off the radar with some mail filters.

Confirmed that `_psl.usercontent.goog` exists contains the right record.

Please ensure this record remains to continually authorize its inclusion in the PSL.

PR #1421, opened 2021-09-16
Add digitaloceanspaces.com & regional subdomains

robbat2:
Add all regions of DigitalOcean Spaces to the PSL, to deliniate the
names which can be registered by users. This mirrors the inclusion of
Amazon S3.

Please note the contact email address goes to a superset of the domain
registration group, suitable for reaching all critical points of contact
regarding DigitalOcean's PSL entries.

Signed-off-by: Robin H. Johnson <rjohnson@digitalocean.com>

* [X] Description of Organization
* [X] Reason for PSL Inclusion
* [X] DNS verification via dig
* [X] Run Syntax Checker (make test)

* [X] Each domain listed in the PRIVATE section has and shall maintain at least two years remaining on registration, and we shall keep the _PSL txt record in place

__Submitter affirms the following:__ 
  * [ ] We are listing any third party limits that we seek to work around in our rationale such as those between IOS 14.5+ and Facebook (see [Issue #1245](https://github.com/publicsuffix/list/issues/1245) as a well-documented example)
  * [X] This request was _not_ submitted with the objective of working around other third party limits
  * [X] The [Guidelines](https://github.com/publicsuffix/list/wiki/Guidelines) were carefully _read_ and _understood_, and this request conforms
  * [X] The submission follows the [guidelines](https://github.com/publicsuffix/list/wiki/Format) on formatting

---
__For Private section requests that are submitting entries for domains that match their organization website's primary domain:__

``` 
Seriously, carefully read the downline flow of the PSL and the guidelines.
Your request could very likely alter the cookie and certificate (as well as other) behaviours on your 
core domain name in ways that could be problematic for your business.

Rollback is really not predicatable, as those who use or incorporate the PSL do what they do, and when.
It is not within the PSL volunteers' control to do anything about that.  

The volunteers are busy with new requests, and rollbacks are lowest priority, so if something gets broken 
it will stay that way for an indefinitely long while.
```
(Link: [about propogation/expectations](https://github.com/publicsuffix/list/wiki/Guidelines#appropriate-expectations-on-derivative-propagation-use-or-inclusion))

 * [X] Yes, I understand.  I could break my organization's website cookies etc. and the rollback timing, etc is acceptable.  Proceed.
---

Description of Organization
====

Organization Website: https://www.digitalocean.com/

I am Robin H. Johnson. I work for DigitalOcean. DigitalOcean offers many cloud services including virtualized servers, managed databases, managed kubernetes clusters, storage services, and an application platform.

Reason for PSL Inclusion
[...]

PR #1404, opened 2021-09-01
Added Amazon EMR Notebooks domain suffixes

paragpc:
* [X] Description of Organization
* [X] Reason for PSL Inclusion
* [x] DNS verification via dig
* [X] Run Syntax Checker (make test)

* [x] Each domain listed in the PRIVATE section has and shall maintain at least two years remaining on registration, and we shall keep the _PSL txt record in place

__Submitter affirms the following:__ 
  * [ ] We are listing any third party limits that we seek to work around in our rationale such as those between IOS 14.5+ and Facebook (see [Issue #1245](https://github.com/publicsuffix/list/issues/1245) as a well-documented example)
  * [X] This request was _not_ submitted with the objective of working around other third party limits
  * [X] The [Guidelines](https://github.com/publicsuffix/list/wiki/Guidelines) were carefully _read_ and _understood_, and this request conforms
  * [X] The submission follows the [guidelines](https://github.com/publicsuffix/list/wiki/Format) on formatting

---
__For Private section requests that are submitting entries for domains that match their organization website's primary domain:__

``` 
Seriously, carefully read the downline flow of the PSL and the guidelines.
Your request could very likely alter the cookie and certificate (as well as other) behaviors on your 
core domain name in ways that could be problematic for your business.

Rollback is really not predictable, as those who use or incorporate the PSL do what they do, and when.
It is not within the PSL volunteers' control to do anything about that.  

The volunteers are busy with new requests, and rollbacks are lowest priority, so if something gets broken 
it will stay that way for an indefinitely long while.
```
(Link: [about propogation/expectations](https://github.com/publicsuffix/list/wiki/Guidelines#appropriate-expectations-on-derivative-propagation-use-or-inclusion))

 * [x] Yes, I understand.  I could break my organization's website cookies etc. and the rollback timing, etc is acceptable.  Proceed.
---

Description of Organization
====
Amazon EMR is the industry-leading cloud big data platform for processing vast amounts of data using open source tools such as Apache Spark, Apache Hive, Apache HBase, Apache Flink, Apache Hudi, and Presto.

Organization Website: https://aws.amazon.com/emr/

I'm a Sr. Software Development Engineer working for Amazon Web Services in Amazon EMR team.

Reason for PSL Inclusion
====
[...]

PR #1403, opened 2021-08-31
Add AWS Cloud9 domains

aprex:
Add domains for AWS Cloud9 service. AWS Cloud9 is IDE in browser, which allows customers to write and run code inside of browser.
Cloud9 provides Application Preview and Webviews functionalities, which implementation relies on domains added with this PR.

* [x] Description of Organization
* [x] Reason for PSL Inclusion
* [x] DNS verification via dig
* [x] Run Syntax Checker (make test)

* [x] Each domain listed in the PRIVATE section has and shall maintain at least two years remaining on registration, and we shall keep the _PSL txt record in place

__Submitter affirms the following:__ 
  * [x] We are listing any third party limits that we seek to work around in our rationale such as those between IOS 14.5+ and Facebook (see [Issue #1245](https://github.com/publicsuffix/list/issues/1245) as a well-documented example)
  * [x] This request was _not_ submitted with the objective of working around other third party limits
  * [x] The [Guidelines](https://github.com/publicsuffix/list/wiki/Guidelines) were carefully _read_ and _understood_, and this request conforms
  * [x] The submission follows the [guidelines](https://github.com/publicsuffix/list/wiki/Format) on formatting

---
__For Private section requests that are submitting entries for domains that match their organization website's primary domain:__

``` 
Seriously, carefully read the downline flow of the PSL and the guidelines.
Your request could very likely alter the cookie and certificate (as well as other) behaviours on your 
core domain name in ways that could be problematic for your business.

Rollback is really not predicatable, as those who use or incorporate the PSL do what they do, and when.
It is not within the PSL volunteers' control to do anything about that.  

The volunteers are busy with new requests, and rollbacks are lowest priority, so if something gets broken 
it will stay that way for an indefinitely long while.
```
(Link: [about propogation/expectations](https://github.com/publicsuffix/list/wiki/Guidelines#appropriate-expectations-on-derivative-propagation-use-or-inclusion))

 * [x] Yes, I understand.  I could break my organization's website cookies etc. and the rollback timing, etc is acceptable.  Proceed.
---

Description of Organization
====

Organization Website: https://aws.amazon.com/cloud9/
AWS Cloud9 is a service within Amazon Web services. It's a cloud-based integrated development environment (IDE) that lets you write, run, and debug your code with just a browser. It includes a code editor, debugger, and terminal.

I'm a software engineer working for Amazon Web Services, Cloud9 team https://www.linkedin.com/in/viktor-shcherba.

Reason for PSL Inclusion
====

[...]

PR #1422, opened 2021-09-17
Update AWS Elastic Beanstalk domain suffixes.

erichays:

* [X] Each domain listed in the PRIVATE section has and shall maintain at least two years remaining on registration, and we shall keep the _PSL txt record in place

  * [X] We are listing any third party limits that we seek to work around in our rationale such as those between IOS 14.5+ and Facebook (see [Issue #1245](https://github.com/publicsuffix/list/issues/1245) as a well-documented example)
  * [X] This request was _not_ submitted with the objective of working around other third party limits
  * [X] The submission follows the [guidelines](https://github.com/publicsuffix/list/wiki/Format) on formatting

---

``` 
Seriously, carefully read the downline flow of the PSL and the guidelines.
Your request could very likely alter the cookie and certificate (as well as other) behaviours on your 
core domain name in ways that could be problematic for your business.

Rollback is really not predicatable, as those who use or incorporate the PSL do what they do, and when.
It is not within the PSL volunteers' control to do anything about that.  

The volunteers are busy with new requests, and rollbacks are lowest priority, so if something gets broken 
it will stay that way for an indefinitely long while.
```

 * [X] Yes, I understand.  I could break my organization's website cookies etc. and the rollback timing, etc is acceptable.  Proceed.
---

Description of Organization
====
AWS Elastic Beanstalk is an easy-to-use service for deploying and scaling web applications and services.

Organization Website: https://aws.amazon.com/elasticbeanstalk/

I am a Sr. Security Engineer on the AWS Elastic Beanstalk team.

Reason for PSL Inclusion
====

AWS Elastic Beanstalk domains have been on the public suffix list since 2013. This pull request adds new regional suffixes to the list.

Domain Name: elasticbeanstalk.com
Updated Date: 2019-08-26T19:19:56+0000
Creation Date: 2011-01-04T23:11:58+0000
Registrar Registration Expiration Date: 2024-01-04T08:00:00+0000

Related Commits and Pull Requests:
```
https://github.com/publicsuffix/list/commit/86848d24d25d408f4b68fd9b053279d41c06d2f0
https://github.com/publicsuffix/list/commit/bb0be1b211e3620360d3ed91bcf13fda8d1378de
https://github.com/publicsuffix/list/pull/484
https://github.com/publicsuffix/list/pull/579
https://github.com/publicsuffix/list/pull/597
https://github.com/publicsuffix/list/pull/628
```

DNS Verification via dig
=======

```
$ dig +short TXT _psl.elasticbeanstalk.com
"https://github.com/publicsuffix/list/pull/1422"
```

make test
=========

make test results: https://pastebin.com/nxH2eCpV

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -10710,6 +10710,8 @@ us-east-1.amazonaws.com
 cn-north-1.eb.amazonaws.com.cn
 cn-northwest-1.eb.amazonaws.com.cn
 elasticbeanstalk.com
+af-south-1.elasticbeanstalk.com
+ap-east-1.elasticbeanstalk.com
 ap-northeast-1.elasticbeanstalk.com
 ap-northeast-2.elasticbeanstalk.com
 ap-northeast-3.elasticbeanstalk.com
@@ -10718,13 +10720,17 @@ ap-southeast-1.elasticbeanstalk.com
 ap-southeast-2.elasticbeanstalk.com
 ca-central-1.elasticbeanstalk.com
 eu-central-1.elasticbeanstalk.com
+eu-north-1.elasticbeanstalk.com
+eu-south-1.elasticbeanstalk.com
 eu-west-1.elasticbeanstalk.com
 eu-west-2.elasticbeanstalk.com
 eu-west-3.elasticbeanstalk.com
+me-south-1.elasticbeanstalk.com
 sa-east-1.elasticbeanstalk.com
 us-east-1.elasticbeanstalk.com
 us-east-2.elasticbeanstalk.com
 us-gov-west-1.elasticbeanstalk.com
+us-gov-east-1.elasticbeanstalk.com
 us-west-1.elasticbeanstalk.com
 us-west-2.elasticbeanstalk.com
 


sleevi (changes requested):
@erichays In line with #1421 and #1417 , can you provide specific details why a wildcard listing for EBS wouldn’t be more appropriate? As part of trying to ensure a respectful growth rate, we’re asking organizations with a significant number of existing listings to review, and potentially reuse, their existing listings before adding more. If there are technical reasons why not now, it’s useful to have them listed explicitly in order to resolve them for new domains.

erichays:
As mentioned in https://github.com/publicsuffix/list/pull/484 when the wildcard was removed, we do allow some of our customers to use third-level hostnames in elasticbeanstalk.com in addition to all of the region-specific suffixes. A wildcard would negatively impact those customers.

sleevi:
Can you share examples? Is there any hope of moving these customers?

If not, can you consider adding new names on a subdomain specific for wildcards?

To the extent the PSL remains a thing, which is admittedly unfortunate, it’s utility is matched or limited by how domain operators operate their domain namespace. While I personally would love for the PSL to not need to exist, and domain operators to not have to consider these tradeoffs, in the absence of that, it seems worth rethinking the proposed hierarchy to limit the growth rate/need to modify the PSL. This gives you the flexibility to not need to update the PSL for new regions, and gives users the benefit of not requiring knowledge of ELB baked into software downloaded by billions of users - effectively, anything using cookies, but also frequently, things which scan for domains (e.g. to auto link in chats) or manage spam. As a community resource, hopefully there’s a balance to be had here?

erichays:
> Can you share examples? Is there any hope of moving these customers?

Any example I provide would be listing a customer-specific Elastic Beanstalk environment. This usage pattern is an artifact of the age of our service. When we launched Elastic Beanstalk in 2011, we did not use regional suffixes. However, after several years, we added the regional suffixes in order to allow reuse of hostnames across multiple regions. At that time, customers who were using the legacy elasticbeanstalk.com were allowlisted to continue using that suffix. All new customers beyond that point use the regional suffixes.

> If not, can you consider adding new names on a subdomain specific for wildcards?

Unfortunately, at this point in our service's lifecycle, it is unlikely that we would migrate from elasticbeanstalk.com to some other domain. As I mentioned in the edit of my previous comment, all new AWS services should follow a regional pattern where wildcards are sufficient. Only older services, like S3 or Elastic Beanstalk should need to add regional suffixes going forward.

sleevi:
There’s a possibility we may need to reject this request, and that’s why I’d like to try find workable alternatives here.

For example, because you have this mix of legacy and regions on the second level domain, I’m trying to ask about setting aside a third-level domain (`Foo`), such that all future regions are of the form `*.foo.elasticbeanstalk.com`, and that’s how the PSL entry looks.

Explicitly: I’m trying to signal that the current pattern is not really sustainable, and so we need to find a strategy here that will work going forward.

erichays:
> Explicitly: I’m trying to signal that the current pattern is not really sustainable, and so we need to find a strategy here that will work going forward.

I acknowledge this. We can start looking into a solution going forward, but the regions listed in this PR were all publicly launched and are in use by customers, some as far back as 2018. Any new solution will take some time to implement.

sleevi:
Entirely fair! And that’s certainly a factor we’re looking at: is this the tail end of the old batch, or the continuation of a pattern with no end in sight. Equally, given how many cloud services look to Amazon and model their offerings based on existing Amazon approaches, it seems like a great opportunity to identify solutions that could or could not work for Amazon, because it means it should still scale even for the smaller providers who are not yet near the number of entries.

Without trying to ask you to preannounce your regions in this, it’s hopefully a reasonable question to ask how long do you think you’d need to identify viability/non-viability of options going forward? I’m asking to try and get a sense of how long the status quo (which presumably may include more regions not yet launched, but in progress) is maintained, versus how likely we can get to something scalable for the future.

erichays:
At Amazon, the pattern has shifted to things that are wildcardable for new services. You can look at my draft PR for AWS App Runner, launched in May 2021, to see evidence of this: https://github.com/publicsuffix/list/pull/1424

Looking at the list, it appears that only S3 and Elastic Beanstalk have legacy patterns that do not match a wildcardable standard. CloudFront, Elastic Cloud Compute (EC2), Elastic Load Balancing (ELB), and AWS Global Accelerator all have patterns that do not appear to require non-wildcard regions.

I've already started a conversation about what options we may have moving forward (migrate customers off of elasticbeanstalk.com so that it can be wildcarded, add a 3rd level subdomain so that it can be wildcarded, stop using the PSL), but my guess is that either of the first two options will take more time than it takes for Amazon to launch one or more new regions...and I obviously don't like the third option, as the PSL provides significant protections to our customers.

sleevi:
Thanks! The fact that the conversation has started, the implications understood, and that new services have less risk is great. As it relates to browsers and cookies and renderable (vs downloadable) content, it’s probably also worth exploring if there’s opportunities to consolidate domains, rather than service-specific domains (some of the discussion in #1417 ), but that’s primary for thinking of the future.

dnsguru:
@erichays is there a reason that there was not a response to the 

- [ ]  "We are listing any third party limits that we seek to work around in our rationale such as those between IOS 14.5+ and Facebook" ?

erichays:
> @erichays is there a reason that there was not a response to the
> 
>    "We are listing any third party limits that we seek to work around in our rationale such as those between IOS 14.5+ and Facebook" ?

@dnsguru I understood this as an either/or with the next question:
Question 1: We're trying to work around these listed limits
Question 2: We aren't trying to work around any limits

Since we are not attempting to work around any limits, I selected the second option. Should I check the first option, anway, even though we aren't attempting to work around limits?

dnsguru:
> > @erichays is there a reason that there was not a response to the
> > "We are listing any third party limits that we seek to work around in our rationale such as those between IOS 14.5+ and Facebook" ?
> 
> @dnsguru I understood this as an either/or with the next question:

These are individual questions. It is an "AND".  ( 😸 )

> Question 1: We're trying to work around these listed limits
> Question 2: We aren't trying to work around any limits

A little different interpretation than I intended.  I may need to revise that if that was your takeaway. 

The objective of Q1 is to have the PR submitting party describe any and all intended limits that they are attempting to deliberately work around, so that we (and future eyes - the comment thread of the PR is essentially the transparency mechanism for all to review) clearly understand the objectives of the PR.  The objective of Q2 is to affirm that Q1 covered it and there are no others the submitting party intends for working around imits that were left out.

> Since we are not attempting to work around any limits, I selected the second option. Should I check the first option, anway, even though we aren't attempting to work around limits?

It sounds like yes.  You've listed "all" of them, because the full set is zero, do by overtly stating this is not intending to work around third party limits, et viola.

erichays:
@dnsguru Thanks for the clarifiaction! I have updated the answer to check the box for that. It may be worth adding a section to the bottom part of the form that explicitly asks for a list of limits. In that case, I would have put None and confidently checked that box.

sleevi:
Unfortunately both https://github.com/publicsuffix/list/pull/1404 and https://github.com/publicsuffix/list/pull/1403 are instances of new AWS services that don't scale. I think we'll need to hold on validating those. I'm going to hold off processing this further until we decide on whether to go forward with https://github.com/publicsuffix/list/pull/1460 to better account for this.

PR #1459, opened 2021-10-14
Add `user.localcert.dev`

lann:

  * [x] We are listing *any* third-party limits that we seek to work around in our rationale such as those between IOS 14.5+ and Facebook (see [Issue #1245](https://github.com/publicsuffix/list/issues/1245) as a well-documented example)
  * [x] This request was _not_ submitted with the objective of working around other third-party limits

---

``` 
Seriously, carefully read the downline flow of the PSL and the 
guidelines. Your request could very likely alter the cookie and 
certificate (as well as other) behaviours on your core domain name in 
ways that could be problematic for your business.

Rollback is really not predicatable, as those who use or incorporate 
the PSL do what they do, and when. It is not within the PSL volunteers' 
control to do anything about that.  

The volunteers are busy with new requests, and rollbacks are lowest 
priority, so if something gets broken by your PR, it will potentially 
stay that way for an indefinitely long while.
```

 * [x] *Yes, I understand*.  I could break my organization's website cookies etc. and the rollback timing, etc is acceptable.  *Proceed*.
---

Description of Organization
====

Organization Website: https://localcert.dev

I am an individual providing a public service for software developers to simplify development and testing of TLS services. Localcert is a CLI tool and hosted service that provides developers with an experience similar to [ngrok](https://ngrok.com/), but for local networks and without proxying.

Reason for PSL Inclusion
====

Localcert works in tandem with an ACME service provider (i.e. Let's Encrypt, though other RFC8555-compliant providers should work) to provide a developer with a unique user domain and certificate that looks like `*.r5pemzugjz27crch5n3v4d7wh3r62xxwytbolbek5x7py.localcert.dev` that can be used to resolve local network addresses with domain names of the form `ip10-11-12-13.r5pemzugjz27crch5n3v4d7wh3r62xxwytbolbek5x7py.user.localcert.dev`. Inclusion in the PSL would prevent cookies from crossing between user domains and allow each user to have a separate Let's Encrypt rate limit bucket.

_This dynamic IP resolution may appear to violate the first [Non-Acceptance Factor](https://github.com/publicsuffix/list/wiki/Guidelines#validation-and-non-acceptance-factors)_, but I think the design of Localcert makes it an acceptable exception:

> We do not accept entries for use as DNS wildcards, such that e.g. 1-2-3-4.foo.tld resolves as IP address 1.2.3.4. This basically projects the security properties of the IP address space onto the domain name space, and we don't feel that is safe. IP addresses can be dynamically allocated to multiple mutually-untrusting parties; domain names generally are not.

* The Localcert DNS server only resolves [reserved IPv4 addresses](https://en.wikipedia.org/wiki/Reserved_IP_addresses#IPv4): private networks (e.g. 192.168.0.0/16), loopback addresses (127.0.0.0/8), and link-local addresses (169.254.0.0/16).
* Because these addresses aren't publicly routed, users cannot obtain a ACME certificate with the [HTTP or TLS challenges](https://letsencrypt.org/docs/challenge-types/). Instead, the Localcert server assists in DNS-01 challenge validation.
* Localcert generates the user domain (`<user_hash>.user.localcert.dev`) by hashing the user's ACME Account URL (e.g. ` https://acme-v02.api.letsencrypt.org/acme/acct/12345` for Let's Encrypt). This Account URL is authenticated as part of the DNS-01 challenge validation by proxying a replay-resistant [signed ACME request](https://datatracker.ietf.org/doc/html/rfc8555#section-6.5) from the client to the ACME server.

This security model means that only the ACME account holder can obtain a certificate for their specific user domain. I believe this addresses the rationale behind the non-acceptance factor. If there are still concerns I very much want to discuss - I don't want to run a public service that makes the internet worse :slightly_smiling_face:.

DNS Verification via dig
=======

```
dig +short TXT _psl.user.localcert.dev
"https://github.com/publicsuffix/list/pull/1459"
```

make test
=========

```
$ make test
...
# TOTAL: 5
# PASS:  5
# SKIP:  0
# XFAIL: 0
# FAIL:  0
# XPASS: 0
# ERROR: 0
```

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -12526,6 +12526,10 @@ ip.linodeusercontent.com
 // Submitted by Victor Velchev <admin@liquidnetlimited.com>
 we.bs
 
+// Localcert : https://localcert.dev
+// Submitted by Lann Martin <security@localcert.dev>
+*.user.localcert.dev
+
 // localzone.xyz
 // Submitted by Kenny Niehage <hello@yahe.sh>
 localzone.xyz


public_suffix_list.dat
```diff
@@ -12516,6 +12516,10 @@ ip.linodeusercontent.com
 // Submitted by Victor Velchev <admin@liquidnetlimited.com>
 we.bs
 
+// Localcert : https://localcert.dev
+// Submitted by Lann Martin <security@localcert.dev>
+user.localcert.dev
```
sleevi, line 12521: Based on your description of the names, shouldn't this be

`*.user.localcert.dev`?

sleevi (changes requested):

public_suffix_list.dat
lann: Maybe? My intent is for `A.X.localcert.dev` and `A.Y.localcert.dev` to be in different "security domains". My understanding is that `*.user.localcert.dev` would put `A.X.localcert.dev` and `B.X.localcert.dev` in different domains as well, which I don't think is necessary in this case.

lann: After further review and testing I believe Let's Encrypt _will_ issue a cert for domains in the private section, so I don't think that is a concern here.

I would still slightly prefer the rule as-is; I believe it fits the purposes of the PSL and would be slightly more useful to users. But again, if there is still concern about it I'll make the change.

lann:
If its useful, the (staging) web application is now up: https://staging.localcert.dev/

dnsguru:
@lann sleevi suggested a change - once that is put in place we can review this to nudge it forward from limbo towards either merge/include or wontfix/close

lann:
Done.

dnsguru (approved):
- DNS Verification Done
- Tests Pass
- No Conflict With Base
- Guidlines Conformant

Approved

PR #1515, opened 2022-02-08
Add `example` and `test` TLDs

jeffkaufman:
The `example` and `test` TLDs are intended for documentation and testing respectively (see RFCs 2606 and 6761).  Even though subdomains cannot actually be registered under these TLDs, if they were real TLDs they would be public suffixes.  For example, documentation might give `news.example` as an example of a site, and automated testing might use `news.test` as a test site. Including these domains on the public suffix matches how they are used.

This PR was prompted by debugging why a browser was not applying cross-site protection in an automated test which used `site1.test` and `site2.test` as examples of separate sites.

In putting this PR together I noticed that https://web.dev/samesite-cookies-explained/ gives `evil.example` and `your-blog.example` as examples of separate sites, even though `example` isn't on the public suffix list.

--------------

* [n/a] DNS verification via dig

  * [x] We are listing *any* third-party limits that we seek to work around in our rationale such as those between IOS 14.5+ and Facebook (see [Issue #1245](https://github.com/publicsuffix/list/issues/1245) as a well-documented example)
  * [x] This request was _not_ submitted with the objective of working around other third-party limits

---

``` 
Seriously, carefully read the downline flow of the PSL and the 
guidelines. Your request could very likely alter the cookie and 
certificate (as well as other) behaviours on your core domain name in 
ways that could be problematic for your business.

Rollback is really not predicatable, as those who use or incorporate 
the PSL do what they do, and when. It is not within the PSL volunteers' 
control to do anything about that.  

The volunteers are busy with new requests, and rollbacks are lowest 
priority, so if something gets broken by your PR, it will potentially 
stay that way for an indefinitely long while.
```

 * [x] *Yes, I understand*.  I could break my organization's website cookies etc. and the rollback timing, etc is acceptable.  *Proceed*.
---

Description of Organization
====

Organization Website: https://www.jefftk.com

I ran into this in my day job as a software engineer at Google, but I'm not speaking for Google in filing this.

Reason for PSL Inclusion
====

The `example` and `test` TLDs are intended for documentation and testing
respectively (see RFCs 2606 and 6761).  Even though subdomains cannot actually
be registered under these TLDs, if they were real TLDs they would be public
suffixes.  For example, documentation might give `news.example` as an example
of a site, and automated testing might use `news.test` as a test site.
Including these domains on the public suffix matches how they are used.

DNS Verification via dig
=======

Not applicable: this isn't something we do for ICANN domains, as far as I can tell?

make test
=========

All tests passed.

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -998,6 +998,9 @@ net.et
 // eu : https://en.wikipedia.org/wiki/.eu
 eu
 
+// example : https://en.wikipedia.org/wiki/.example
+example
+
 // fi : https://en.wikipedia.org/wiki/.fi
 fi
 // aland.fi : https://en.wikipedia.org/wiki/.ax
@@ -6157,6 +6160,9 @@ td
 // tel: https://en.wikipedia.org/wiki/.tel
 // http://www.telnic.org/
 tel
+
+// test : https://en.wikipedia.org/wiki/.test
+test
 
 // tf : https://en.wikipedia.org/wiki/.tf
 tf


jeffkaufman:
Thanks!  I tried to find earlier discussion but I guess I didn't look hard enough... Looking now, #2 links to https://bugzilla.mozilla.org/show_bug.cgi?id=1161102 for earlier discussion.

dnsguru:
Rather than closing this now, will leave as a `wontfix` and let discussion or updates happen if they will.

Talk us out of this being a wontfix, basically.

jeffkaufman:
Reading more about it, I think the current case is ok.  Step #2 of the [algorithm](https://publicsuffix.org/list/) covers these through having a fallback of `*`, so there should be no practical effect of including vs excluding these domains.

dnsguru:
The converse of that argument applies as well, though.

dnsguru:
@jeffkaufman I appreciate that you took the time to submit this and you make good salient points.

It triggered some good conversations and after socializing this a bit offline within the dev community, adding these would alter expected behaviors and test cases on a number of libraries because of the special-ness of these particular strings and how they can/should be used. 

The PSL has such a massive footprint and libraries that incorporate it, which requires some care in changing things.  

Many projects, libraries, browsers, etc incorporate the PSL and then perform some delta on it to add/remove as they see fit, and the suggestion I received from developers was to leave the behavior of test and example in that realm, where it can be electively managed in that integrator/incorporator system, at their discretion.

We tend to be cautious and thoughtful about impacts like this that have effect on downline users/integrations on changes that have a cascade impact, and this change definitely would.  

Because a solution does exist for individual integrations, this does not hobble choice, but it seems like a bad path to make that choice for everyone using the PSL.

I am going to close this, but wanted to provide the additional narrative out of respect for your time and care in submitting this, and I hope this above will help you and others reading this to understand that it was not a situation that did not receive consideration before closing.

PR #1519, opened 2022-02-15
Remove `couk.me` and `ukco.me` from private section

fearworksmedia:
We have decided to not renew the domains couk.me and ukco.me, and instead we are focusing on our .uk domains. I wanted to clean these two .me domains up so they aren't left on the list.

These two domains are currently in the expiration grace period, and we would appreciate their removal from the list.

Many thanks.


dnsguru:
rolls back portion of #963 

Thank you for making the effort to have us remove it, most folks in similar situations don't have the class to submit removals and just leave the entries in for the expired names, so that part is appreciated.

This will be shown as a rationale support for the requirements of >2y registrations on new PR

dnsguru:
- updater = submitter
- no conflict with base
- tests pass

Approved

Issue #1533, opened 2022-03-10
Remove `no.com` from PSL due to new owner/different use than subdomain registry


cicku:
Hi,

Just a note that no.com should be removed from the list. It is now a crypto wallet website.

gbxyz:
CentralNic no longer runs it as a registry zone, but I cannot say what the current owner/operator is doing with it.

We've held off on submitting change requests for domains that have been off-boarded from our registry system, based on the view that only the current domain owner can properly request changes that affect their domains.

cicku:
Well I wonder if there is a better way to remove it because originally this domain (along with few others *.com) was not added by the domain owner but the registry.

dnsguru:
Will review - want to take care here.  The domain may have been repurposed,
but if the new use did some continuity plan for affected subdomain
registrants to mitigate cookie or other behaviors that a change to there
being a PSL entry present (or not) might disrupt, the PSL volunteers do not
want to trigger that by our own PR to address this issue.

IF there is no _psl TXT record in-zone for this entry, this can be taken
into consideration in a community submission of a pull request to remove
no.com, especially where there is a submitter or commentary from a prior
submitting party.

dnsguru:
This can be re-opened by the domain owner of no.com as a Pull Request to indicate removal, tied to the associated DNS _PSL proof

PR #1534, opened 2022-03-12
Remove no.com from PSL due to new owner/different use than subdomain registry

cicku:
#1533

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -10973,7 +10973,6 @@ za.com
 ar.com
 hu.com
 kr.com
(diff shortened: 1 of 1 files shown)


dnsguru:
We typically would want to see this come from the domain name owner, validated by them placing a _PSL TXT record matching this PR number URL in the zone file for no.com 

The precedent of a third party submission affecting a change in a zone that they cannot demonstrate administrative control over is a slippery slope, so this may sit here a while until some form of intervention, expiry automation or addition into zone by the owner/admin of that name.

PR #1557, opened 2022-04-21
Fix comments delete space and deprecation of io/ioutil

gongqi-zhen:
Description
=========
I would like to contact you regarding modifications to the PSL creation script.

* fix comments
* delete space end of line
* deprecation of "io/ioutil"

This is not a feature that must be modified.
Please use this information for future revisions.

Thanks again for your community.

make test
=========

```
% pwd
/Users/makoto/github/list/tools
git remote -v
origin	https://github.com/gongqi-zhen/list.git (fetch)
origin	https://github.com/gongqi-zhen/list.git (push)
upstream	https://github.com/publicsuffix/list.git (fetch)
upstream	https://github.com/publicsuffix/list.git (push)
% go version
go version go1.18.1 darwin/amd64
% go test
PASS
ok  	github.com/publicsuffix/list/tools	0.483s%
```
workflow-passed

### DIFF
--- a/tools/newgtlds.go
+++ b/tools/newgtlds.go
@@ -9,7 +9,6 @@ import (
 	"flag"
 	"fmt"
 	"io"
-	"io/ioutil"
 	"net/http"
 	"os"
 	"strings"
@@ -268,7 +267,7 @@ func (d datFile) String() string {
 // data is found within the dat file. An error is returned if the file can't be read
 // or if the gTLD data span can't be found or is invalid.
 func readDatFile(datFilePath string) (*datFile, error) {
-	pslDatBytes, err := ioutil.ReadFile(datFilePath)
+	pslDatBytes, err := os.ReadFile(datFilePath)
 	if err != nil {
 		return nil, err
 	}
@@ -339,7 +338,7 @@ func getData(url string) ([]byte, error) {
 			url, http.StatusOK, resp.StatusCode)
 	}
 
-	respBody, err := ioutil.ReadAll(resp.Body)
+	respBody, err := io.ReadAll(resp.Body)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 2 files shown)


cfredric:
https://pkg.go.dev/io/ioutil does indicate that ioutil's functionality is provided by os or io. Specifically:
* ioutil.ReadFile just calls os.ReadFile (https://pkg.go.dev/io/ioutil#ReadFile)
* ioutil.ReadAll just calls io.ReadAll (https://pkg.go.dev/io/ioutil#ReadAll)
* ioutil.WriteFile just calls os.WriteFile (https://pkg.go.dev/io/ioutil#WriteFile)

This PR looks reasonable to me, provided that whoever runs the script has Go 1.16 or later.

cpu:
> This PR looks reasonable to me, provided that whoever runs the script has Go 1.16 or later.

@cfredric That's a good point, we would need to update the Github action workflow that runs this tool before merging:
https://github.com/publicsuffix/list/blob/663236d380ce0f80fa5880b0536988ebc0af6d54/.github/workflows/tld-update.yml#L19

It's presently using Go 1.15 - not for any particular reason, its just what was the latest when I wrote the workflow and I haven't done a good job of keeping it up to date.

gongqi-zhen:
Thanks for the confirmation.

The workflow was already working with go1.18.1 with the current caret ranges settings.
So it seems to me that as long as go1.x, there is no problem without modification here.

https://github.com/publicsuffix/list/blob/663236d380ce0f80fa5880b0536988ebc0af6d54/.github/workflows/tld-update.yml#L19

From the following figure you can see that go1.18.1 is used and the unit test passes.
Running_on_go1 18 1

If it is administratively more appropriate, please correct it.

I really appreciate your time and effort.

dnsguru:
>@dnsguru Would you be willing to merge this PR? It gets my stamp of approval and should be a no-op in terms of the automated updates.

Your vouch is mighty here, as this has largely been your area. 
so, 

## Yes

on the condition you monitor that it behaves as expected :)

cpu:
Thanks :-)

gongqi-zhen:
I have confirmed that the workflow has been completed successfully and I will delete the branch.
Thank you for accepting.

#895_workflow_completed

cpu:
The updated automation produced its first pull-request (https://github.com/publicsuffix/list/pull/1567) since this update and the content looks good to me 👍

PR #1448, opened 2021-10-07
add ip.linodeusercontent.com

stvnjacobs:
* [x] Description of Organization
* [x] Reason for PSL Inclusion
* [x] DNS verification via dig
* [x] Run Syntax Checker (make test)
* [x] Each domain listed in the PRIVATE section has and shall maintain at least two years remaining on registration, and we shall keep the _PSL txt record in place

__Submitter affirms the following:__ 

  * [x] We are listing *any* third-party limits that we seek to work around in our rationale such as those between IOS 14.5+ and Facebook (see [Issue #1245](https://github.com/publicsuffix/list/issues/1245) as a well-documented example)
  * [x] This request was _not_ submitted with the objective of working around other third-party limits
  * [x] The [Guidelines](https://github.com/publicsuffix/list/wiki/Guidelines) were carefully _read_ and _understood_, and this request conforms
  * [x] The submission follows the [guidelines](https://github.com/publicsuffix/list/wiki/Format) on formatting
---

Description of Organization
====

Organization Website: https://www.linode.com

I am an Engineer at Linode.

We are a cloud hosting provider, offering Linux virtual machines (Linodes), a hosted load balancing solution (NodeBalancers), S3 compatible object storage, and other hosted services.

Reason for PSL Inclusion
====

We are requesting the addition of `ip.linodeusercontent.com`. We are in the process of transitioning our users from `*.members.linode.com` and `*.nodebalancer.linode.com` to this new domain, away from a subdomain of our business domain at `linode.com`. These domain records are provided as a convenience to our customers for each virtual machine and load balancer.

Both `members.linode.com` and `nodebalancer.linode.com` domains are already on this list, and we will request their removal once they are no longer in use.

I have read the guidelines, and did want to note that we plan to have the format of `1-2-3-4.ip.linodeusercontent.com` for a virtual machine with an IP address of `1.2.3.4`. I understand the guideline prohibiting this as a type of wildcard domain, but I would like to assure the reviewer that we will only be serving these records for public IP addresses within ranges under our ownership, and that each IP is reserved for a customer's virtual machine or load balancer.

I believe the intent of this entry in the guidelines is to prohibit hosting a DNS service for IP addresses which are not under our control, and therefore may be shared among multiple untrusted parties -- this is not the case here. Please correct me if I have misinterpreted this guideline.

DNS Verification via dig
=======

```
dig +short TXT _psl.ip.linodeusercontent.com
"https://github.com/publicsuffix/list/pull/1448"
```

Domain Registration
=========

[...]

PR #1589, opened 2022-07-11
Add `cloudscale.ch` domains

gaudenz:

* [X] Each domain listed in the PRIVATE section has and shall maintain at least two years remaining on registration, and we shall keep the \_PSL txt record in place in the respective zone(s) in the affected section

  * [X] We are listing *any* third-party limits that we seek to work around in our rationale such as those between IOS 14.5+ and Facebook (see [Issue #1245](https://github.com/publicsuffix/list/issues/1245) as a well-documented example)

  * [X] This request was _not_ submitted with the objective of working around other third-party limits

  * [X] The submission follows the [guidelines](https://github.com/publicsuffix/list/wiki/Format) on formatting and sorting
  * [X] *Yes, I understand*.  I could break my organization's website cookies etc. and the rollback timing, etc is acceptable.  

Description of Organization
====
cloudscale.ch AG is a Swiss based cloud hosting (IaaS) provider. We offer cloud VMs, an S3-compatible Object Storage and other related services to customers.

Organization Website: https://www.cloudscale.ch/

I'm a System Engineering team lead at cloudscale.ch. 

Reason for PSL Inclusion
====

cust.cloudscale.ch
----

DNS records in this domain are provided as a convenience to our customers for each virtual machine.

I have read the guidelines, and did want to note that we plan to have the format of 1-2-3-4.cust.cloudscale.ch for a virtual machine with an IP address of 1.2.3.4. I understand the guideline prohibiting this as a type of wildcard domain, but I would like to assure the reviewer that we will only be serving these records for public IP addresses within ranges under our ownership, and that each IP is reserved for a customer's virtual machine.

I believe the intent of this entry in the guidelines is to prohibit hosting a DNS service for IP addresses which are not under our control, and therefore may be shared among multiple untrusted parties -- this is not the case here. Please correct me if I have misinterpreted this guideline.

objects.lpg.cloudscale.ch / objects.rma.cloudscale.ch
----

These are the primary domain names for our S3-compatible object storage. We support virtual-host style bucket addressing. This means customers can access their bucket under the name `<bucket_name>.objects.(lpg|rma).cloudscale.ch`. This is a common addressing scheme for S3-compatible object storage systems. Other such services are already listed in the PSL, including `s3.amazonaws.com`, `s3.teckids.org` and Scaleway's object storage (multiple domains). 

DNS Verification via dig
=======

```
dig +short TXT _psl.cust.cloudscale.ch
"https://github.com/publicsuffix/list/pull/1589"
```

```
dig +short TXT _psl.objects.lpg.cloudscale.ch
"https://github.com/publicsuffix/list/pull/1589"
```
```
dig +short TXT _psl.objects.rma.cloudscale.ch
"https://github.com/publicsuffix/list/pull/1589"
```

Results of Syntax Checker (`make test`)
=========

All tests passed

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -11075,6 +11075,12 @@ pages.dev
 trycloudflare.com
 workers.dev
 
+// cloudscale.ch AG : https://www.cloudscale.ch/
+// Submitted by Gaudenz Steinlin <support@cloudscale.ch>
+cust.cloudscale.ch
+objects.lpg.cloudscale.ch
+objects.rma.cloudscale.ch
+
 // Clovyr : https://clovyr.io
 // Submitted by Patrick Nielsen <patrick@clovyr.io>
 wnext.app


gaudenz:
> As you wisely note, the cust.cloudsale.ch portion of the request is in violation of the guidelines. 

I tried to convince you in the remainder of my rationale that this is not in violation of the guidelines. The important point is that we do not use this as a wildcard mapping of IP addresses to DNS names. This is only used for IP addresses under our control which are assigned to customer VMs. The main reasoning in the guidelines about this issue is "IP addresses can be dynamically allocated to multiple mutually-untrusting parties; domain names generally are not.". This does not apply to our case as only addresses from our allocation will be used.

The entry `ip.linodeusercontent.com` serves the same purpose for Linode and was accepted in #1448. 

The rationale for including these in the PSL is that each name below `cust.cloudscale.ch` belongs to a different customer. They don't have anything in common and having them under the same `cust.cloudscale.ch` suffix should not indicate any relation or trust between them. The VMs the entries below `cust.cloudscale.ch` point to are run by independent parties. These parties should not be able to set cookies for each other or cross other security boundaries. The PSL entry should also signal to blacklist providers that these are independent entries and should not be blacklisted together.

> If you remove that portion of the requested PM, the other objects entries can be reviewed and considered, assuming that you can provide a rationale for the request that states the purpose. Each PR should stand on its own for its rationale, and the PR currently refers to others but does not contain an actual rationale that outlines the purpose.

Sorry for not providing enough rationale for these entries. As said they refer to an S3-compatible object storage with virtual-host style bucket addressing. The bucket name is the part of the DNS name below the proposed PSL entries. These bucket names are chosen by the customer and independent of each other. They belong to independent parties. These object storage buckets can be used to host static web content under subdomains of the proposed PSL entries. The entries serves the purpose of enforcing security boundaries between these sub-domains. It should also signal to third-parties that subdomains of these entries are unrelated to each other and should for example not be blacklisted together.

dnsguru:
In the rationale for the IP-based cust entry, it is IP based and is in violation of [https://github.com/publicsuffix/list/wiki/Guidelines#validation-and-non-acceptance-factors](https://github.com/publicsuffix/list/wiki/Guidelines#validation-and-non-acceptance-factors) Bullet 1

In the case of the other PR for Linode where an exception was made, a similar assertion was provided that claimed that only owned IP addresses would be used.  A similar exception will be made on this request, but a new Label was created called "MAY DESERVE SECURITY REVIEW" to flag for future audit of IP space against the assertion of the entries only being used for IP addresses held by the company.    A similar flag was added to the Linode PR which had the same rationale and assertion.

href:
I'm hesitant to comment on this, since I know you guys are volunteers and I've been there. But I figured since it's close to a year since the last update, I might try my luck.

Anything we can do to get this merged?

simon-friedberger:
Sorry, I am a little late to this party. Could you please explain why `1-2-3-4.cust.cloudscale.ch` is necessary?
1. Since the IP is available a connection can be established without the DNS lookup.
2. Wouldn't it also be possible to pick a different name? What exactly makes the IP address relevant here?

simon-friedberger:
@dnsguru It seem expiration information is unavailable for .ch domains. How do we handle this?

simon-friedberger:
> We have to set some kind of reverse PTR for the IP addresses our customers get when they launch a VM, similar to how other providers do this (see Linode's "Reason for PSL Inclusion": [#1448 (comment)](https://github.com/publicsuffix/list/pull/1448#issue-1020134108)).

Can you elaborate, please. The Linode PR also doesn't give a good explanation.

> Additionally, this allows our customer to easily get a Let's Encrypt certificate for a specific IP (Let's Encrypt does not issue certificates for IPs).

This would work just as well if the customers got randomly generated hostnames or $clientname_$counter. Which would make it less likely that somebody accidentally ends up with a valid certificate for somebody else's domain when IPs get shuffled around.

href:
> We have to set some kind of reverse PTR for the IP addresses our customers get when they launch a VM, similar to how other providers do this (see Linode's "Reason for PSL Inclusion": [#1448 (comment)](https://github.com/publicsuffix/list/pull/1448#issue-1020134108)).

From what we can tell it's best practice. See https://en.wikipedia.org/wiki/Forward-confirmed_reverse_DNS

> Forward-confirmed reverse DNS (FCrDNS), also known as full-circle reverse DNS, double-reverse DNS, or iprev, is a networking parameter configuration in which a given IP address has both forward (name-to-address) and reverse (address-to-name) Domain Name System (DNS) entries that match each other. This is the standard configuration expected by the Internet standards supporting many DNS-reliant protocols. David Barr published an opinion in [RFC 1912](https://tools.ietf.org/html/rfc1912) (Informational) recommending it as best practice for DNS administrators, but there are no formal requirements for it codified within the DNS standard itself.[[1]](https://en.wikipedia.org/wiki/Forward-confirmed_reverse_DNS#cite_note-1)

> > Additionally, this allows our customer to easily get a Let's Encrypt certificate for a specific IP (Let's Encrypt does not issue certificates for IPs).
> 
> This would work just as well if the customers got randomly generated hostnames or $clientname_$counter. Which would make it less likely that somebody accidentally ends up with a valid certificate for somebody else's domain when IPs get shuffled around.

`$clientname_$counter` might be a bad idea as it would be trivial to find out the client name and how many instances he runs, which would violate privacy. Random hostnames would be an option, but it's not as straight-forward to implement, and it would not be predictable for the customer who would first have to do a lookup of his reverse PTR. All doable, but it seems to not provide us with any tangible benefits.

Our approach of setting a PTR is also obvious to the end-user and leaves no open questions. Otherwise, I might wonder: is the random name generated each time? Might it include an undesired combination of words that then show up on my dashboard?

We also do not necessarily want to "leak" how often we re-assign IPs to changing server instances, which is something a competitor could find out if we cycle our random domains.

As for Let's Encrypt Cert re-use: You would not typically use this domain for anything long-lasting. Usually, you might want to serve some test-site for an hour or so, and then throw it away. The chances of someone else hijacking that same IP (which is not guaranteed), and being able to show the same content as before and trick a user are slim.

Our customers are largely engineers themselves, so the implications are likely to be well understood here.

simon-friedberger (approved):

dnsguru:
> this allows our customer to easily get a Let's Encrypt certificate for a specific IP (Let's Encrypt does not issue certificates for IPs).

LE not allowing this was a strong component for the rule and the restriction here.

You do mention there is precedent in similar entries.  

I notice that @simon-friedberger has reviewed this but wanted me to review also. Rather than leave this pr pending I will note that we added a flag for commentary to come from the security commumity on this PR.

We have not received such commentary or any objections to this request proceeding.

To me, this may be reasonable rationale for an exception, noting that others requesting similar in the future citing the exception should not expect a similar exception, and should  expect a similar more protracted review and/or likely wontfix outcomes.

dnsguru (approved):
CA do not issue certificates for IP addresses typically, and there is a rule about this in the guidelines.

* exception to ip address rule received extended comment period with a flag for review from the security community
÷ no objections were received during this flag period
* flag will remain
* the entry will be removed upon comment from the security community of it creating security issues

Approving, with important notes.

in the absence of commentary objecting to the exception where the security community had been granted opportunity to do so.

*Given that the PR will be reverted if there is later indication this exception created unexpected issues, it is not recommended to establish commercial dependance on this exception remaining in place.*

simon-friedberger:
@weppos also approved in private correspondence.

PR #1759, opened 2023-05-12
Remove .zip and .mov

mholt:
Many websites and apps automatically linkify text matching the PSL, causing surprising results and creating a false sense of security. These TLDs are unsafe to be on the PSL due to their deceptive nature. PoCs are emerging and organizations are blocking these TLDs.

- https://twitter.com/SwiftOnSecurity/status/1657111224931459072
- https://twitter.com/_2c2c2/status/1657113931935260672
- https://financialstatement.zip/
- https://twitter.com/mholt6/status/1657133439546695680

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -8569,9 +8569,6 @@ moto
 // motorcycles : 2014-01-09 XYZ.COM LLC
 motorcycles
 
-// mov : 2014-01-30 Charleston Road Registry Inc.
-mov
-
 // movie : 2015-02-05 Binky Moon, LLC
 movie
 
@@ -10083,9 +10080,6 @@ zara
 
 // zero : 2014-12-18 Amazon Registry Services, Inc.
 zero
-
-// zip : 2014-05-08 Charleston Road Registry Inc.
-zip
 
 // zone : 2013-11-14 Binky Moon, LLC
 zone


CydeWeys:
TLD operator here -- these are real TLDs, existing on the root DNS nameservers, that should be on the PSL. Removing them from the PSL would create grave vulnerabilities for users of these websites.

felixfontein:
Removing existing TLDs from the PSL for this reason would just be wrong. This list is used for **many** different reasons, and just because these entries are bad for one very specific use-case, they are still needed for (almost) all others.

dnsguru:
These are legit TLDs in the ICP3 root.   This will not proceed

Really, the expressed concerns are more of a glaring example of a
disconnect between the developer and security community and domain name
governance, where they would benefit from more engagement within ICANN.

weppos:
Hi Matt,

As much as I understand where you are coming from, I agree with @dnsguru. As project maintainers, it is beyond our scope to become the internet policeman and we should stick to the core purpose of this library.

Over the time, consumers have adopted this list for many different purposes. It would be impossible (and inappropriate) to make ad-hoc choices that could please this or that consumer use case. If any tool is using the list to perform auto-linking, the most appropriate approach would be that the tool implements an allow list (or block list) approach to filter out extensions they consider inappropriate.

CydeWeys:
It's also worth pointing out that the public suffix list is poorly suited for linkification purposes anyway; it's the wrong tool for the job.  It's much larger than necessary for this purpose, making linkification inefficient.  What you want is the list of TLDs, [which can be obtained here](https://data.iana.org/TLD/tlds-alpha-by-domain.txt).  But even for that you might want to rank the TLDs in descending order by # of domains or some usage metric and then pick a cut-off point.

[This list](https://ntldstats.com/tld) gives a good idea of which of the new gTLDs are actually being used and which aren't.  Just to pick one from that list at random, [.open](https://ntldstats.com/tld/open) is an unlaunched brand TLD with a single domain on it (the ICANN-mandated NIC domain).  The upside of linkifying .open domain names is thus zero, whereas the downside of linkifying them is you'll incorrectly create links anytime someone accidentally uses the common English word "open" without including a space after a period.  So clearly this shouldn't be linkified (unless and until it's actually launched), even though it appears on the PSL and on the list of delegated TLDs.

dnsguru (commented):
These are sourced from ICANN's json, and the PSL is just a catalog of what's in that + root + .onion as defined by IETF.

This PR is being labelled as wontfix and will be closed

pmnlla:
I don't mean to become an expert on the topic but I argue that the TLDs should be removed from the list. As it's a primary source for many linkification processes, I don't see what's the issue with removing these two specific TLDs as they pose many risks to certain users.

Since decades ago, programs haven't embedded .zip links as links to local files on the filesystem, just like they won't do now. But people will assume that these are links to local files, click on them, and some **minor** issues may occur there (emphasis on minor). I think we can all agree that being able to buy domains like financialstatement.zip and invoice-form.zip is an issue, and they could potentially be used for malicious intent. **This does not invalidate the argument that they are invalid TLDs, and removing them purely to this reason will spark interest in the removal of other TLDs as they may be potential risk factors.**

The biggest argument that I've seen in this thread is that "it's still a valid TLD". And it's a valid argument that I don't want to dismiss. However, are there any other reasons for why this list still needs to keep these two faulty-by-design TLDs?

Aside from its existence, and a lack of regulation, are there any other reasons to keep the TLD on the list? I'm asking for the sake of curiosity.

felixfontein:
> However, are there any other reasons for why this list still needs to keep these two faulty-by-design TLDs?

Because this list contains **all** TLDs. And this list is mainly NOT used for linkification. Removing the TLDs because two TLDs on this list are not good for one very specific use-case of this list (out of many, many others where these TLDs must be part of the list!) is a really bad idea.

ThiefMaster:
AFAIK the PSL is used by browsers to determine for example what's considered an origin root e.g. for cookies. For example. it indicates by having `com` in it that `example.com` and `foo.example.com` and `bar.example.com` typically have the same owner so `foo.example.com` can set a cookie for `example.com`.

OTOH, because `co.uk` is on the list, `foo.co.uk` and `bar.co.uk` are considered completely separate (in particular those two do NOT tend to have the same owner and in fact one of them could very well be malicious and would *love* to steal cookies from the other one, or set cookies for the whole `.co.uk` domain to do some nasty user tracking ;))

There's a nice explanation here: https://publicsuffix.org/learn/

mholt:
To clarify, this PR was more of an awareness campaign, which seems to have succeeded. gg, everyone :100:

Honorable mention goes to Google for undoing years of anti-phishing and anti-deception work overnight, and for making computers more confusing and risky to use. :medal_sports:

pmnlla:
And now you'll never know whether the PR in PR campaign stands for Pull Request or Public Relations.

teh-maxh:
> This PR was more of an awareness campaign, which seems to have succeeded.

Who exactly are you trying to make aware of this?

Issue #1806, opened 2023-07-20
Update Autopull from ICANN JSON to address contract renewal noise


dnsguru:
I have a (hopefully) fast and easy update to the autopull @cpu

Please See PR #1805 and look at the code change being automated.

The very first of 10 year contract renewals between registry operators and ICANN are being reflected in the json as a reset contract date, and this is causing autopull noise.  Can the autopull have some logic that ignores a contract date change if there is no change in the sponsoring entity name?  

We are now at one decade from some of the first contracts being signed between ICANN and the registry operators that applied for and successfully received delegations to operate registries that they applied for in the 2012 round of new gTLDs.  So we'll have a lot of noise come through as all of these will have date changes as they renew contracts.

The 99.8% majority, but not all, of the new TLDs from the ICANN 2012 new gTLD round (give or take .web and/or some others that are taking longer to sort out _stuff_) have been delegated.

We'd want to keep the autopull going as it has been helpful for quickly catching changes in sponsoring organization per TLD or where there are applicants sunsetting their aspirations, but it is believed that at whatever time in the future that the next next round opens and delivers TLDs, that the current automation process would be able to handle it.

As a background, we used the contracting date to front-load the addition of TLDs before they were being added to the root by the IANA.  The contracting date would precede whatever pre-delegation testing period that would occur before the TLD would be added to the root zone.

To address the time-lag for browsers' (and others') incorporation of the PSL as a static resource within code creating a situation where a gTLD would get added to the root but work on some browsers and not others based upon the timing of the last snapshot taken, there was a choice to use the date which the registry operator and ICANN had entered into a contract as a 'trajectory signal' - that the TLD would be being added to the root within a reasonably short timeframe.

We would still want that to be happening, but just want to make this little tweak on the date changes to mute a flood of autopulls that would have no practical functional value.

cpu:
Just wanting to note that I saw the tag here and will take a look when time permits. Thanks!

dnsguru:
Thanks Daniel - hopefully this is trivial to ignore the `DateOfContractSignature` string if it is the only change in `list/tools/newgtlds.go` per entry.

I will leave the #1805 in an open state so that testing is possible, but I suspect we'll get daily nudges

dnsguru:
ALTERNATIVE: 
[...]

PR #1809, opened 2023-07-26
Update AutoPull - Temporary fix for #1806

dnsguru:
This updates newtlds.go to replace the <contract date> with "Contracted" in the header per 2012 round nTLD as a tempfix to #1806 

Short version of #1806 is that now we're 10 years ++ hence from the 2012 round TLD launches, the contracts are hitting their 10 year horizon and are being renewed.   ICANN is updating the contract date field in the gtlds.json file, and this would result in a large quantity of auto-pull to merge with the new dates as this happens.

There are discussions with ICANN's technical team to see about them leaving the contract dates intact in contractdate vs re-using the contractdate field to store the contract renewal date.

It is anticipated that this pull request being merged will cause one large pull request that alters the entire nTLD section's contracted dates in the comment per nTLD, but will then reduce the subsequent noise level when ICANN renews contracts with the various Registries.

### DIFF
--- a/tools/newgtlds.go
+++ b/tools/newgtlds.go
@@ -118,7 +118,7 @@ func (e *pslEntry) normalize() {
 	e.ALabel = strings.TrimSpace(e.ALabel)
 	e.ULabel = strings.TrimSpace(e.ULabel)
 	e.RegistryOperator = strings.TrimSpace(e.RegistryOperator)
-	e.DateOfContractSignature = strings.TrimSpace(e.DateOfContractSignature)
+	e.DateOfContractSignature = "Contracted"
 
 	// If there is no explicit uLabel use the gTLD as the uLabel.
 	if e.ULabel == "" {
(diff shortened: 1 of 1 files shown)


cpu:
(If it's easier I can open a PR implementing my suggested feedback in a couple hours)

cpu:
There's actually more work to do... It looks like the Go tool unit tests aren't being run as part of pull request status checks anymore, just the cronjob that runs the code in `master`. We can tell that's the case because your change would have broken the unit tests and the branch status is reporting all green :-/

The Go tests used to be run explicitly via Travis for all pull requests:
https://github.com/publicsuffix/list/blob/f95b832e23abb49dd5d54a3272f710247bcb13d9/.travis.yml#L6

The new GitHub actions replacement for Travis that was added in https://github.com/publicsuffix/list/pull/1750 only runs the `Makefile` tests:
https://github.com/publicsuffix/list/blob/f95b832e23abb49dd5d54a3272f710247bcb13d9/.github/workflows/test.yml#L18-L19

cpu:
PTAL at https://github.com/publicsuffix/list/pull/1812 I think it will achieve the same goals as this branch in a tidier way.

weppos:
Closing in favor of https://github.com/publicsuffix/list/pull/1812

dnsguru:
I have just been notified by ICANN's Technical Services that they will preserve the contract date of original signing.

So now the question will be, "do we re-introduce the date?"

Rather than reintroducing it, I think instead introducing some human-readable datestamp in the .dat header would solve the problem that the contract date was helping address - troubleshooting time-lag where folks were not updating their file as TLDs were being added to the zone file.

dnsguru:
It was really only for the troubleshooting specifics of lag time when TLDs
would get added to the root but not function in some of the browsers or
other applications or systems.

I'd say it is ok to not re-introduce

PR #1815, opened 2023-08-01
Update tld-update.yml to automatically add labels when autopull catches deltas and generates PR

dnsguru:
This is non-urgent.

When the autopull runs, this auto-adds labels to indicate this is automation and impacts the upper 'ICANN' section

Did not see a way to also automate the project as well in the automation handler, but this will save volunteers steps in labelling when reviewing.

### DIFF
--- a/.github/workflows/tld-update.yml
+++ b/.github/workflows/tld-update.yml
@@ -39,6 +39,9 @@ jobs:
           committer: "GitHub <noreply@github.com>"
           author: "GitHub <noreply@github.com>"
           branch: psl-gtld-update
+          labels: |
+            ✅ autopull
+            🚩ICANN (IANA/ICP-3) Section
           delete-branch: true
 
       - name: Check outputs


dnsguru:
```
name: Create or Update Project Card for new item
        if: ${{ steps.cpr.outputs.pull-request-number }}
        uses: peter-evans/create-or-update-project-card@v2
        with:
          project-number: 2
          column-number: 8657926
          issue-number: ${{ steps.cpr.outputs.pull-request-number }}
```

does that look right?  I confirmed that 
* "List Add/Mod/Del" = project-number 2 and 
* "To-Do" is column-number 8657926

dnsguru:
@cpu can you give it a 'sanity peek'?

.github/workflows/tld-update.yml
```diff
@@ -39,7 +39,18 @@ jobs:
           committer: "GitHub <noreply@github.com>"
           author: "GitHub <noreply@github.com>"
           branch: psl-gtld-update
+          labels: |
```
weppos, line 42: Just as an FYI, in the past I had bad experience with labels with icons. And by "bad" I mean the unicode char did cause random failures in integrations.

I've personally moved away from using any icon/emoji at all in labels.

weppos (approved):

.github/workflows/tld-update.yml
```diff
@@ -39,7 +39,18 @@ jobs:
...
+            ✅ autopull
+            🚩ICANN (IANA/ICP-3) Section
           delete-branch: true
+          
+      - name: Create or Update to List Add/Mod/Del To-Do Card
+        if: ${{ steps.cpr.outputs.pull-request-number }}
+        uses: peter-evans/create-or-update-project-card@v2
+        with:
+          project-number: 2
+          column-number: 8657926
```
cpu, line 52: Should this be a string-typed `column-name` input instead? I don't see `column-number` listed in [the documented action inputs](https://github.com/peter-evans/create-or-update-project-card#action-inputs)

dnsguru:
Updated with 
```
column-name: To-Do
```
Should be good to go @cpu @weppos

cpu (commented):
LGTM

weppos (approved):

dnsguru:
OK - Gonna merge this, and then @CPU's updated automation that adds the IANA DB link to see if the automation pull behaves as expected

weppos:
Screenshot 2023-08-07 at 11 10 35

@dnsguru the action failed to run. The reason is because that repo has projects disabled. Even if we turn that on, I'm afraid this won't work (we can still test).

I just noticed now we use an org-level project, and I'm afraid that will require a bit of changes. The token generated by GH during an action run is a transient repo-level token. It doesn't have the permission to touch something outside the repo. If we want to continue with an org-level project we need to generate an individual token and add it to the action secrets.

I turned projects on at the repo level to see if by chance we get lucky and it works. But I don't have too much confidence it will.

dnsguru:
Yeah, I think we just need to settle for labels.

dnsguru:
There used to be a lovely konban-esque chart for org level projects that vanished.

weppos:
> There used to be a lovely konban-esque chart for org level projects that vanished.

It's still there, but legacy (called _classic_) projects are receiving less attention. There's a new type of projects which is the current focus and has different views, including Kanban boards.

dnsguru:
Is there a way to transfer the past 6 years' legacy project attributes to
the new ones?

weppos:
It seems to be possible, with certain limitations.
https://docs.github.com/en/issues/planning-and-tracking-with-projects/creating-projects/migrating-from-projects-classic

dnsguru:
Thank you for looking -  I'd vote no on that...

I am not sure that the "juice/squeeze" ratio of that specific migration
technique merits doing it.

it looks like it would move the project itself to another project as part
of the process... and with my fast-read, it appeared to have a 1200 item
limit, which would cause us to lose a lot of institutional data, and I
would likely need to recreate labels and other stuff....

PR #1812, opened 2023-07-26
tools: skip contract date rendering, small CI fixups.

cpu:
## Description

This is an alternative to https://github.com/publicsuffix/list/pull/1809

More cleanup is probably in order (in particular, making a proper Go module for this tool) but for now let's do the simple stuff.

### tools: don't render date of contract signature.
ICANN is updating the contract date field in the `gtlds.json` file for many contracts in the near future. Since including the contract date in the `.dat` file as comments isn't very helpful to anyone, and leaving it in during this renewal period would result in a large quantity of auto-pull requests for no-op changes, we choose to omit the date as a quick fix.

This commit updates the `newtlds.go` tool to no longer process/render the `DateOfContractSignature`. Unit tests are updated to reflect this change.

### ci: run Go unit tests as part of test.yml.
Previously the Go unit tests for the `newgtlds` tool were run as part of the Travis workflow for each branch push. This is helpful to catch breaking changes to the Go tooling _before_ they are merged and used in the `tld-update.yml` workflow.

This commit adds the required steps to `test.yml` so that these unit tests are run by GitHub actions similar to the way they were for Travis.

### ci: switch to latest stable go.

Does what it says on the tin.

### ci: remove stale comment in tld-update.yml

The comment about the automation being on-hold was leftover from a period ~two years ago. It isn't accurate, the automation _has_ been running, and it properly handles the header timestamp issue that caused it to be temporarily disabled.

## Meta discussion

Speaking frankly: I find it hard to prioritize supporting this automation. I'm not a maintainer of this repo, I don't have as much stake in the PSL as when I worked at Let's Encrypt, I've been writing less Go lately, and I have other hobby projects to spend my free time on :-) 

I'd be happy for someone else to take over maintenance and future development. Alternatively, if Mozilla or another organization has funding I'm available for part-time contract work. 

However, since I know many of the folks involved here are volunteers and the PSL is load bearing for the broader internet I will continue to help where I can if there aren't better alternatives. Mostly I feel it's important to lay my cards on the table to help set expectations.

PR #1817, opened 2023-08-03
tools: include IANA TLD URL in new gtld updates.

cpu:
This branch updates the `tools/newgtlds.go` tooling to include a comment line with the gTLD's IANA registry URL for each gTLD.

E.g. for the tld `aaa`, includes a comment line: `// https://www.iana.org/domains/root/db/aaa.html`.

Follow-up to #1812

### DIFF
--- a/tools/newgtlds.go
+++ b/tools/newgtlds.go
@@ -10,6 +10,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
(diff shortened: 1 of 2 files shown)


weppos (approved):

dnsguru (approved):
LGTM

PR #1827, opened 2023-08-09
Automation - Roll back broken 'project' assignment in autopull 

dnsguru:
Stripping out the project/card definition stuff - appears to use a different context of "project" than intended

This rolls back the cause of the https://github.com/publicsuffix/list/pull/1815#issuecomment-1667496019 and fixes the daily ICANN automation


weppos (approved):
👍 

(sorry for the delay, I was on holiday)

Issue #1439, opened 2021-09-28
How to help vs "help" maintain the list - _psl txt record DNS validation crucial


dnsguru:
Hello, gratefully, to those that review the list and submit PR as updates to listings that they are not the admin/registry/manager of.  These help the volunteers keep up the freshness of the PSL, and it is absolutely and always appreciated.

That said, there is a validation step used for ensuring that the NIC or domain owner is in the loop and supports the change, to ensure these are intentional and match their desired namespace operation.  We do this by requiring that a _psl txt record be added on affected zones or domains that matches the PR # - and this helps as connection between the PR and the DNS record show administration proof.

When submitting a PR, understand that we as the PSL maintainers are volunteers and are not resourced to chase up the NIC or adminstrators of domains; while there are some of us who participate in ICANN or who have the opportunity to make presentations inside the TLD registry community and regional ccTLD events to help with awareness of the PSL, and we describe what it is and isn't and all of the benefits to maintaining entries for a ccTLD or gTLD, it is ultimately they the registries who must be involved in at very least approving / showing validation in the DNS.

When submitting a request for a TLD or domain name that you are a third party to, please include within your efforts the outreach to the party that will be affected by the PR and a request to review the change that you've proposed and for them to add the _psl txt leaf into the affected zones so that the change can be approved and merged.

Otherwise, the effort of the submission often sits without update or change, until we ultimately may close it due to absense of response from the administrator or ability to verify it.  These circumastances are bad from our perspective, as we are far too familiar with the loss of some volunteer time to 'disposible effort' that comes from things like #1245 with subsequent rollbacks or where the requestor was 'experimenting'.

So please do the additional outreach and contact the respective domain owner or admins.  And assume this has been tried previously.  No, we do not have a contact list to share, we use the IANA.org db listed contacts for ccTLDs just like anyone.

Thank you to community members and submitters who are taking their time to help offload the burden from us, and hopefully this issue identifies how to ensure that effort is most effective.

Examples: #1437 #1436 #1435 #1432 #1213 #1021 #1032 #1182

sleevi:
@dnsguru Beyond the validation side, this swath of PRs seems to be net-negative (in the bloaty sense). Do you have any concerns with me closing these all as WontFix?

dnsguru:
Can we leave them open for a few weeks?  i will be speaking at centr.org
tech day and they are good reference

sleevi:
Not sure I follow?

[...]

PR #1868, opened 2023-10-07
Update public_suffix_list.dat to include `mil.uk` / `mod.uk`

mbfrias:
Added mod.uk and mil.uk second level domains in use by the Ministry of Defense

https://www.nominet.uk/uk-domains/additional-domains/

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -6710,7 +6710,7 @@ org.zw
 
 // newGTLDs
 
-// List of new gTLDs imported from https://www.icann.org/resources/registries/gtlds/v2/gtlds.json on 2023-12-06T15:14:09Z
+// List of new gTLDs imported from https://www.icann.org/resources/registries/gtlds/v2/gtlds.json on 2023-12-12T15:13:54Z
 // This list is auto-generated, don't edit it manually.
 // aaa : American Automobile Association, Inc.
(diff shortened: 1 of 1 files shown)


dnsguru:
Thank you for taking the tine to raise this PR.  It is an outstanding example of why this project is reviewed by humans.

As mentioned in issue #1439, where third parties signal changes to entries, especially those delegated TLDs direct from IANA / ICANN namespace, and even more particularly gov/mil, there is validation required.

We would certainly want to see DNS proof to ensure the additions are intentional and deliberate and would involve the appropriate administrative parties, and/or proof of nexus with the affected namespace.

These two submitted are not new namespace.  Also, because they have existed for a long period of time, it may be that these are deliberately and intentionally absent in the PSL.

Due to the manner in which presense or absence in the PSL might affect behaviours or how a namespave might be treated by apps or systems that might voluntarily opt to do so, things like the PR are important to confirm are intentional, known changes.

I will be face to face with Nominet Technical staff in two weeks and will review this with them.

dnsguru:
Nominet was spoken to and they pointed to the whois/rdap records for each of these namespaces as indicating that these are each a public suffix, but did not get confirmation that the cookie affectation or other results of PSL listing were desired by the respective namespace operators.

PR #1896, opened 2023-11-16
Remove trailing whitespace

wangweij:
Remove the single trailing whitespace in the whole text.


simon-friedberger (commented):
LGTM but also fixed in https://github.com/publicsuffix/list/pull/1894

PR #1901, opened 2023-11-20
Introduce Go Modules to tooling

aaomidi:
Solves #1897 - also spoken of here: https://github.com/publicsuffix/list/issues/1325#issuecomment-855462155.

I initially added this with go workspaces, but turns out those aren't actually meant to be pushed (even though, in this specific scenario, they made the commands a little bit cleaner).

I've made sure that none of the existing workflows fail with this change.

I've moved the go tools outside of `tools`. This wasn't absolutely necessary but I didn't want the bash scripts and the go module to live in the same directory -- I think it would've added additional long term cost to keep them in the same directory.

I've updated all the github actions workflows to keep working with the new layout. 

`make test`:

```
~~ redacted
PASS: libpsl_icu_load_dafsa_fuzzer
PASS: libpsl_icu_fuzzer
PASS: libpsl_icu_load_fuzzer
============================================================================
Testsuite summary for libpsl 0.21.2
============================================================================
# TOTAL: 3
# PASS:  3
# SKIP:  0
# XFAIL: 0
# FAIL:  0
# XPASS: 0
# ERROR: 0
============================================================================

~~ redacted

PASS: test-is-public-builtin
PASS: test-is-public
PASS: test-registrable-domain
PASS: test-is-cookie-domain-acceptable
PASS: test-is-public-all
============================================================================
Testsuite summary for libpsl 0.21.2
============================================================================
# TOTAL: 5
# PASS:  5
# SKIP:  0
# XFAIL: 0
# FAIL:  0
# XPASS: 0
# ERROR: 0
============================================================================

```

```
$ ./tools/patchnewgtlds 
++ command -v go
+ '[' -x /usr/local/go/bin/go ']'
++ realpath ./tools/patchnewgtlds
+ SCRIPT=/workspaces/list/tools/patchnewgtlds
++ dirname /workspaces/list/tools/patchnewgtlds
+ BASEDIR=/workspaces/list/tools
+ go run -C /workspaces/list/tools/../newgtlds . -overwrite -psl-dat-file=/workspaces/list/tools/../public_suffix_list.dat
```

PR #1906, opened 2023-11-27
Change the order of flags in github actions workflows.

aaomidi:
Mistake introduced in #1901

Fixes https://github.com/publicsuffix/list/actions/runs/7006982080/job/19060050437

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -18,7 +18,7 @@ jobs:
           go-version: 'stable'
 
       - name: Run Go unit tests
-        run: go test -C ./tools .
+        run: go test -C ./tools -v .
 
       - name: Install dependencies
         run: sudo apt install -y autopoint
--- a/.github/workflows/tld-update.yml
+++ b/.github/workflows/tld-update.yml
@@ -26,7 +26,7 @@ jobs:
           go-version: 'stable'
 
       - name: Run unit tests
-        run: go test -v -C ./tools .
+        run: go test -C ./tools -v .
 
       - name: Set current date
         id: get-date


dnsguru:
Why is there a rush?

aaomidi:
I accidentally introduced a bug in #1901 to the github worklow: https://github.com/publicsuffix/list/blob/2d1427a99ee0c3eef8b22d06c4d7099200ed41ac/.github/workflows/tld-update.yml#L29

This caused the last tld-update run to fail: https://github.com/publicsuffix/list/actions/runs/7006982080/job/19060050437

This PR fixes that so we don't have this failure on the next run.

.github/workflows/test.yml
```diff
@@ -18,7 +18,7 @@ jobs:
           go-version: 'stable'
 
       - name: Run Go unit tests
-        run: go test -C ./tools .
+        run: go test -C ./tools -v .
```
weppos, line 21: Out of curiosity, any reason to not have:

```suggestion
        run: cd ./tools && go test -v ./...
```

cpu: Instead of chaining a `cd`, what about using the [working-directory](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsworking-directory) job config element?

aaomidi: I don't have a strong opinion on how this is done. Up to maintainers to decide this. My main opinion is this is currently breaking the pipeline, so either we rollback #1901, or we accept this as is and revisit it if the need arises?

I do have a preference for using the tools that Go has built in to the toolchain.

PR #1929, opened 2024-02-02
Add examples of limitations to PR template

### LINKED ISSUE #1889: Add 3rd party limit examples to template
The template currently states:

`__Submitter affirms the following:__ 
\
  * [ ] We are listing *any* third-party limits that we seek to work around in our rationale such as those between IOS 14.5+ and Facebook (see [Issue #1245](https://github.com/publicsuffix/list/issues/1245) as a well-documented example)`

I have not seen any such claims in the issues I have reviewed so far. If we actually want this information. I think it would be a good idea to add an additional list with some examples as follows:

  * [ ] We are listing *any* third-party limits that we seek to work around in our rationale such as those between IOS 14.5+ and Facebook (see [Issue #1245](https://github.com/publicsuffix/list/issues/1245) as a well-documented example)`
    - [Cloudflare](https://developers.cloudflare.com/learning-paths/get-started/add-domain-to-cf/add-site/) 
    - [Letsencrypt](https://letsencrypt.org/docs/rate-limits/)
    -  ...add others, remove irrelevant...

simon-friedberger:
Fixes #1889 

Note: These are the only external links that I am aware of.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -45,6 +45,9 @@ to submit entries to the PSL as a means to work around those limits or
 restrictions.
 -->
   * [ ] We are listing *any* third-party limits that we seek to work around in our rationale such as those between IOS 14.5+ and Facebook (see [Issue #1245](https://github.com/publicsuffix/list/issues/1245) as a well-documented example)
+    - [Cloudflare](https://developers.cloudflare.com/learning-paths/get-started/add-domain-to-cf/add-site/)
+    - [Letsencrypt](https://letsencrypt.org/docs/rate-limits/)
+    - <UPDATE THIS LIST WITH YOUR LIMITATIONS!>
 
 <!--
 The purpose of the question above is to expose limit workarounds.


dnsguru:
this should actually get more enumerated vs leaving that open in the Pull Request - it looks like [TBD] and incomplete

I artuculated a couple in the FB/IOS issue

suggest we close this and get some more in the issue and reopen once theres 4 or more to add here.

the key issue is pushing back on well-funded companies and salaried people punting their responsibilities onto volunteer-run resources.  it is a slap in the face to volunteers and sends a signal they have zero or less value.

dnsguru:
Clarity is likely not the issue - across 15+ years what I have found is that it needs to be somewhat structured or it will not get an answer...

simon-friedberger:
I could extend the list with:
* Facebook Pixel
* Apple IOS PCM
if you think that's an improvement.

Generally, I think this is better than the current situation so I would like to merge it even if it's not the ideal end-state.

PR #1933, opened 2024-02-09
Add Github workflow to check _psl DNS entries on PRs

simon-friedberger:
Let's have more automatic checks!

### DIFF
--- a/tools/pr_checker/check_dns.py
+++ b/tools/pr_checker/check_dns.py
@@ -1,0 +1,148 @@
+import click
+import dns.message
+import dns.name
+import dns.query
+import dns.resolver
+import re
+
+
+def read_rules(psl_filename):
+    """Read rules from a file into a set.
+
+    >>> sorted(read_rules('test_psl_01.dat'))
+    ['!main.bar.example.org', '*.bar.example.org', '*.bd', 'a.example.com', 'at', 'b.example.com.', 'example.com', 'example.org', 'foo.example.org']
+    """
+    rules = set()
+
+    with open(psl_filename) as f:
+        for line in f:
+            line = line.strip()
+
+            if line == "":
+                continue
+
+            if line.startswith("//"):
+                continue
+
+            rules.add(line)
+    return rules
+
+
+def rule2fqdn(rule):
+    """Return the domain name for a rule.
+
+    Removes wildcards and exception qualifiers.
+
+    >>> rule2fqdn("alpha.beta.example.com")
+    'alpha.beta.example.com'
+    >>> rule2fqdn("*.hokkaido.jp")
+    'hokkaido.jp'
+    >>> rule2fqdn("!pref.hokkaido.jp")
+    'pref.hokkaido.jp'
+    """
+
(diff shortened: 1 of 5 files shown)


weppos:
Hi @simon-friedberger, can you provide a bit more context about this PR. Is there a conversation somewhere about it? What's the purpose of this script, and what is the expectation?

I also wonder, I see it was developed in Python. Is it because that's the language you are most familiar with, or is there another reason? I am asking as we had Python code in the past but it was not really the most preferred language and we moved most of the tooling to Go. I am a bit concerned about adding more Python again, for the long term maintainability.

dnsguru:
@simone I asked simon to pause this...  There is a separate dialog with
google about development of automation which I believe was go-based - I
have shared the diagrams and specs I took time to develop in the hopes of
avoiding parallel evolution costing valuable volunteer cycles

simon-friedberger:
I am not against using Go for this if that is the project preference. Python was just the natural choice for me.

dnsguru:
> I am not against using Go for this if that is the project preference. Python was just the natural choice for me.

Same thought, as go is less familiar to me, but want to minimize complexity with >1 approaches at this point so we don't add interop burdens

simon-friedberger:
The submission by Sam [here](https://github.com/publicsuffix/list/pull/1953) is also in Python and @dnsguru has expressed support. Can I go ahead with this?

dnsguru:
After discussion, this looks like a good idea - accompanied by test suite elements to know about impacts.  The automation work is really appreciated

tools/pr_checker/check_dns.py
```diff
@@ -0,0 +1,150 @@
...
+import dns.query
+import dns.resolver
+import re
+
+
+def read_rules(psl_filename):
+    """Read rules from a file into a set.
+
+    >>> sorted(read_rules('test_psl.dat'))
+    ['!main.bar.example.org', '*.bar.example.org', 'a.example.com', 'b.example.com.', 'example.com', 'example.org', 'foo.example.org']
```
dnsguru, line 13: do we need a flat TLD and flat TLD w wildcard rule or not here?

ex: add ```'.com', '*.com'```

dnsguru:
I suggested a TLD and TLD w a sld wildcard in the examples for the test suite 

I was also thinking of IDN support.  Upon some contemplation, I believe those scenarios would funciton identically to their LDH equivelents in the test suite.  Just documenting here that it was considered.  Adding them here would be placebo; cosmetic-only, require additional dependencies/libraries/modules, add complexity and not have any difference in outcome, so would not deliver any value or benefit in exchange for the overhead.

simon-friedberger:
As soon as I get to it we should merge this with parsing and checks for individual rules in the linter here https://github.com/publicsuffix/list/blob/master/linter/pslint.py that should cover IDN issues.

simon-friedberger:
@weppos Please have a look!

tools/pr_checker/check_dns.py
simon-friedberger: Done!

PR #1968, opened 2024-04-19
Clean up list to fix rule sorting within orgs

simon-friedberger:
This makes the sorting of rules within organization's sections match what the script in #1953 wants.

I checked that the actual contents are still the same using:
```
import click


def read_rules(filename):
    rules = set()
    with open(filename) as f:
        for line in f:
            if line.startswith("//"):
                continue
            line = line.strip()
            rules.add(line)
    return rules


@click.command()
@click.argument("file1")
@click.argument("file2")
def main(file1, file2):
    rules1 = read_rules(file1)
    rules2 = read_rules(file2)
    print(f"Are they equal? {rules1 == rules2}")


if __name__ == "__main__":
    main()

```


mozfreddyb (approved):
reviewed the code (but did not rerun it)

weppos (approved):

PR #1995, opened 2024-06-13
Reattach of.by to the Belarus ccTLD block

danderson:
It's in the same situation as `com.by`: not officially reserved according to available documentation, but third-level domain registrations are available as if it were. Logically it belongs in the same block as `com.by`.

Found while developing #1987, because it resulted in surprising output. simon-friedberger recommended making this .dat change, rather than add a custom lint exemption that's only necessary for this one entry.

This is a minor lint cleanup, so most of the PR template does not apply. Applicable items:
 - **Description of change**: above.
 - **make test**: runs successfully, no errors.
 - **Authority**: I'm not authoritative for `by` or `of.by`. However I am not adding/removing any suffixes, only adjusting whitespace.
 - **Justification**: I studied the `by` [registry policy](https://cctld.by/en/documents/instruction-on-the-procedure-of-registration-of-domain/), and verified that neither `com.by` or `of.by` are listed. However I can purchase a 3rd level domain in those suffixes right now, so empirically they should be treated as public suffixes.
   - I'm digging into the registry history and policies more to figure out if a bigger update is justified. This PR is just recognizing the fact that `com.by` and `of.by` are both in the same category of "not on an official list but seem to be public suffixes in BY".

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -675,7 +675,6 @@ mil.by
 // second-level domain, but it's being used as one (see www.google.com.by and
 // www.yahoo.com.by, for example), so we list it here for safety's sake.
 com.by
-
 // http://hoster.by/
 of.by
 


simon-friedberger:
This is only a whitespace change that doesn't affect the meaning of the list.

groundcat:
I contacted the `.by` registry via `info@cctld.by` (https://cctld.by/contacts/), and they confirmed that the registration of `.com.by` and `.of.by` is not managed by them but by private companies "Надежные программы" and "Открытый контакт".

Perhaps they should be moved to the private section? @simon-friedberger 

Below is the response I received:

> Добрый день.
> 
> Регистраторами данных доменов второго уровня являются ООО "Надежные программы" и Открытый контакт, они могут регистрировать домены третьего уровня, можете обратиться к ним.
> 
> С уважением, Управление продаж beCloud
> +375 17 287 11 11
> +375 17 287 11 49
> 
> www.becloud.by

image

image

danderson:
This was going to be my followup, yeah. .by is weird because _any_ 2nd-level domain can offer registrations for 3rd level domains, so if you buy `foo.by` you can just start selling `blah.foo.by` and (if I read the rules correctly) your registrar is obligated to help you record that 3rd level in the .by registry.

The difficulty with moving to the private section is that the private section requires owners. So we would have to contact the registrars that own/sell access to com.by and of.by, and make them take ownership of their suffixes. Otherwise we would just be moving orphaned suffixes around, which doesn't help much :/

simon-friedberger:
It does sound a bit like it should actually be in the private section. Is the situation different from other 2nd level domains reselling 3rd level domains?

danderson:
The only difference from generic "host that gives you a 3rd level domain", afaict, is that the 3rd level domain gets formally added to the `by` registry, with the owner pointing at the 3rd level owner.

For example, `whois com.by` shows it's owned by domain.by, a registrar/webhost company. `whois google.com.by` shows that domain is owned by Google. When google purchased `google.com.by`, the owner of `com.by` got that recorded in the top level registry.

So it's a bit of a weird mix. The 3rd level domains are in the registry, similar to `co.uk` and other registry-reserved domains. But the 2nd level domains aren't special, the `by` registry rules seem to be that anyone who buys a 2nd level domain can start recording 3rd level registrations underneath it.

So yes, that does sound like these should move to the private section, and probably more added. Complete coverage seems impossible to me if the `by` registry isn't publishing a full list, so it would have to be up to the individual owners of the 2nd level domains to take ownership of their PSL entries...

Issue #1989, opened 2024-06-01
Incorrect PSL evaluation rules in the wiki regarding implicit wildcard rules


ko-zu:
Hello,

I believe there is confusion regarding the wildcard evaluation rules in the wiki.
https://github.com/publicsuffix/list/wiki/Format

In the example, 

>
> 1. com
> 2. *.foo.com
>
>...
>
>Cookies may be set for foo.com.

This is incorrect. Cookies may NOT be set for `foo.com`.  According to the linter, `*.example.com` implies `example.com`.

https://github.com/publicsuffix/list/blame/de747b657fb0f479667015423c12f98fd47ebf1d/linter/pslint.py#L230

The current rules in the wiki states that the domain `example.com` does not match `*.example.com`. 
~~However, some TLDs (e.g. *.bd) do not have explicit TLD declarations, which are required and checked by the linter.~~
~~According to the wiki rules, we can set cookies on bd. This is not the expected behavior. ~~

`kobe.jp` does have explicit public suffix declarations while `*.kobe.jp` exists, which is required and checked by the linter. According to the wiki rules, we can set cookies on `kobe.jp`. (Fixed to reflect simon-friedberger's comment. Thanks!)

Implementations should assume `example.com` is also declared if `*.example.com` exists.

In the DNS world, this assumption is incorrect. However, the test data includes lines like:
```
checkPublicSuffix('city.kobe.jp', 'city.kobe.jp');
checkPublicSuffix('www.city.kobe.jp', 'city.kobe.jp');
```
While the PSL only has:
```
jp
*.kobe.jp
!city.kobe.jp
```

If we follow the wiki rules, `kobe.jp` should be on the right.

### Problems in the Wiki

The example above comes from the following evaluation rule definition:

> - A domain is said to match a rule if and only if all of the following conditions are met:
> - When the domain and rule are split into corresponding labels, that the domain contains as many or more labels than the rule.
> - Beginning with the right-most labels of both the domain and the rule, and continuing for all labels in the rule, one finds that for every pair, either they are identical, or that the label from the rule is "*".

According to this definition, the domain `example.com` cannot match `*.example.com` because the domain has fewer labels than the listed rule.

If we need to maintain this scheme, I believe we should add: 

- If the rule is a wildcard, prepend a label `*` for the domain.

Additionally, the statement *"leading and trailing dots are ignored"* is also incorrect. The test data explicitly disallows leading dot.

If you have write permission to the wiki, please take a look into the above issues.

[...]

PR #1998, opened 2024-06-14
Fix test case for exception rule

simon-friedberger:
As pointed out in #1989 there seems to be a bug in a test case where something that has an explicit exception is returned as a public suffix. This fixes that test case.

### DIFF
--- a/tests/test_psl.txt
+++ b/tests/test_psl.txt
@@ -57,8 +57,8 @@ checkPublicSuffix('a.b.ide.kyoto.jp', 'b.ide.kyoto.jp');
 checkPublicSuffix('c.kobe.jp', null);
 checkPublicSuffix('b.c.kobe.jp', 'b.c.kobe.jp');
 checkPublicSuffix('a.b.c.kobe.jp', 'b.c.kobe.jp');
-checkPublicSuffix('city.kobe.jp', 'city.kobe.jp');
-checkPublicSuffix('www.city.kobe.jp', 'city.kobe.jp');
+checkPublicSuffix('city.kobe.jp', 'kobe.jp');
+checkPublicSuffix('www.city.kobe.jp', 'kobe.jp');
 // TLD with a wildcard rule and exceptions.
 checkPublicSuffix('ck', null);
 checkPublicSuffix('test.ck', null);


simon-friedberger:
At least in Firefox it behaves as I would expect:
```
Services.eTLD.getPublicSuffixFromHost("city.kobe.jp")
"kobe.jp"
Services.eTLD.getPublicSuffixFromHost("www.city.kobe.jp")
"kobe.jp" 
```

if anybody knows how to easily test in other browsers please let me know!

simon-friedberger:
Note, this code expects the public suffix:
```
function checkPublicSuffix(host, expectedSuffix)
{
  var actualSuffix = null;
  try {
    actualSuffix = etld.getBaseDomainFromHost(host);
  } catch (e if e.result == Cr.NS_ERROR_INSUFFICIENT_DOMAIN_LEVELS ||
                e.result == Cr.NS_ERROR_ILLEGAL_VALUE) {
  }
  // The EffectiveTLDService always gives back punycoded labels.
  // The test suite wants to get back what it put in.
  if (actualSuffix !== null && expectedSuffix !== null &&
      /(^|\.)xn--/.test(actualSuffix) && !/(^|\.)xn--/.test(expectedSuffix)) {
    actualSuffix = idna.convertACEtoUTF8(actualSuffix);
  }
  do_check_eq(actualSuffix, expectedSuffix);
}
```

while the discussion here https://github.com/rockdaboot/libpsl/issues/48#issuecomment-184582216 indicates that this function may in the past have returned the shortest registrable part. So maybe it was improperly converted at some point?

simon-friedberger:
Could you elaborate a bit? I don't understand how something that has an explicit exception rule could possibly be an expectedSuffix.

ko-zu:
> checkPublicSuffix('city.kobe.jp', 'city.kobe.jp');
> checkPublicSuffix('www.city.kobe.jp', 'city.kobe.jp');

> 
> ```
> Services.eTLD.getPublicSuffixFromHost("city.kobe.jp")
> "kobe.jp"
> Services.eTLD.getPublicSuffixFromHost("www.city.kobe.jp")
> "kobe.jp" 
> ```
> 
> if anybody knows how to easily test in other browsers please let me know!

`!city.kobe.jp` declares that it is not a public suffix. The longest public suffix here is `kobe.jp` implied by `*.kobe.jp`. Firefox follows the linter's rule and the test case, not the wiki.

simon-friedberger:
Oh, the naming here is just confusing `expectedSuffix` and `actualSuffix` are not public suffixes but actually eTLD+1. Thanks for pointing this out!

PR #1999, opened 2024-06-16
tools/internal/parser: refactor to separate text processing from parser main logic

danderson:
This splits up the parser logic: a lower level helper breaks up the input bytes in useful ways and tracks source information for error reports. The main parser uses the helper, so that it can focus on the "business logic" of checking the format and linting, without being polluted with tracking offsets and partial state and stuff.

The PR is in 2 commits, each one is ~50% of the change: the first adds the new helper without using it, in `text.go` and `text_test.go`. The second commit updates `parser.go` to use the helper.

The helper has a bunch more unit tests, but the parser functionality is unchanged, and all pre-existing tests pass without modification. Later PRs (see [todo](https://gist.github.com/danderson/f2d5a7a7e0400ea2a7c9dff18058615c#file-psl-todo-md)) will add missing features, this is just a cleanup to prepare for that.

### DIFF
--- a/tools/internal/parser/text.go
+++ b/tools/internal/parser/text.go
@@ -32,11 +32,9 @@ func (s source) Source() Source {
 		// it visible.
 		panic("can't construct a zero-line Source")
 	}
-	start := s.lineOffset + 1
-	end := start + len(s.lines) - 1
 	return Source{
-		StartLine: start,
-		EndLine:   end,
+		StartLine: s.lineOffset,
+		EndLine:   s.lineOffset + len(s.lines),
 		Raw:       s.Text(),
 	}
 }
--- a/tools/internal/parser/text.go
+++ b/tools/internal/parser/text.go
@@ -1,14 +1,17 @@
 package parser
 
 import (
+	"fmt"
 	"strings"
 )
 
-// source is a block of zero or more lines of source text. It can be
-// decomposed into smaller sub-blocks using its parsing methods. It
-// also preserves source line information and can be turned into an
-// exported Source when needed.
-type source struct {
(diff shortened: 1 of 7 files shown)


tools/internal/parser/text.go
```diff
@@ -0,0 +1,141 @@
...
+// exported Source when needed.
+type source struct {
+	// The lines of source text.
+	lines []string
+	// lineOffset is how many lines are before the beginning of lines,
+	// for sources that represent a subset of the input.
+	lineOffset int
+}
+
+// newSource returns a source for bs.
```
simon-friedberger, line 19: bs?

tools/internal/parser/text.go
```diff
@@ -0,0 +1,141 @@
...
+
+import (
+	"strings"
+)
+
+// source is a block of zero or more lines of source text. It can be
+// decomposed into smaller sub-blocks using its parsing methods. It
+// also preserves source line information and can be turned into an
+// exported Source when needed.
+type source struct {
```
simon-friedberger, line 11: Does that mean there is now a Source and source type?

tools/internal/parser/text.go
```diff
@@ -0,0 +1,141 @@
...
+		StartLine: start,
+		EndLine:   end,
+		Raw:       s.Text(),
+	}
+}
+
+// slice returns the slice of s between startLine and endLine.
+//
+// startLine and endLine behave like normal slice offsets, i.e. they
+// represent the half-open range [startLine:endLine).
```
simon-friedberger, line 47: So using StartLine and EndLine from a Source won't work here because there the EndLine is included, right?

PR #2005, opened 2024-06-20
tool/internal/parser: sanitize input to clean, valid UTF-8

danderson:
The PSL's canonical is valid UTF-8 with no BOM. However, to try and report useful lint errors, the parser tries to detect and normalize all forms of UTF-16, as well as UTF-8 with BOM. Anything other than the specified canonical encoding is reported in validation errors.

---

The upcoming PR that implements validation of block ordering needs to compare block names, which requires comparing with a Unicode collation. Collation is already confusing enough by itself, without also feeding it invalid UTF-8 :). So, this is a best-effort attempt to normalize non-compliant inputs into clean UTF-8, so that we report those errors early and the rest of the parser can assume strings don't contain garbage, at least at the encoding level.

### DIFF
--- a/tools/internal/parser/text.go
+++ b/tools/internal/parser/text.go
@@ -1,8 +1,14 @@
 package parser
 
 import (
+	"bytes"
 	"fmt"
 	"strings"
+	"unicode"
+	"unicode/utf8"
+
+	"golang.org/x/text/encoding"
+	xunicode "golang.org/x/text/encoding/unicode"
 )
 
 // Source is a piece of source text with location information.
@@ -12,11 +18,28 @@ import (
 // behave the same as in Go slices, and select the half-open interval
 // [start:end).
 type Source struct {
-	// The lines of source text.
+	// The lines of source text, sanitized to valid UTF-8 and with
+	// leading and trailing whitespace removed.
 	lines []string
 	// lineOffset is how many lines are before the beginning of lines,
 	// for sources that represent a subset of the input.
 	lineOffset int
+}
+
+// newSource returns a source for bs, along with a preliminary set of
+// input validation errors.
+//
+// source always returns a usable, non-nil result, even when it
+// returns errors.
+func newSource(bs []byte) (Source, []error) {
+	lines, errs := normalizeToUTF8Lines(bs)
+
+	ret := Source{
+		lines:      lines,
+		lineOffset: 0,
+	}
+
+	return ret, errs
 }
 
 // Text returns the source text of s as a string.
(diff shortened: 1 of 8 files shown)


tools/internal/parser/text.go
```diff
@@ -148,3 +163,171 @@ func (s Source) forEachRun(classify func(line Source) bool, processBlock func(bl
...
+	}
+
+	return ret, errs
+}
+
+// guessUTFVariant guesses the encoding of bs.
+//
+// Returns the transformer to use on bs, one of utf8Transform,
+// utf16LittleEndianTransform or utf16BigEndianTransform.
+func guessUTFVariant(bs []byte) encoding.Encoding {
```
simon-friedberger, line 273: Can you elaborate a bit on the scenarios you're adding this for? Is this just for the odd case where somebody uploads UTF-16 instead?

tools/internal/parser/text_test.go
```diff
@@ -9,6 +10,242 @@ import (
...
+				0x74, 0x0, 0x65, 0x0, 0x78, 0x0, 0x74, 0x0,
+			},
+			want:     []string{"utf-16 text"},
+			wantErrs: []error{InvalidEncodingError{"UTF-16LE"}},
+		},
+		{
+			name: "utf16be_input",
+			// "utf-16 text utf-16 text utf-16 text"
+			// linewraps are after the spaces
+			in: []byte{
```
simon-friedberger, line 76: Is there a more readable way to provide these? Maybe by taking the string and encoding it?

simon-friedberger:
Since the goal here is to achieve what the current tool in the linter folder is doing, have you considered reading the test files in there?

simon-friedberger (approved):

tools/internal/parser/text_test.go
danderson: Yeah good point! I think I was trying to avoid using the same library to both encode and decode the text, in case there are bugs... But that makes no sense, it has its own unit tests and is certainly correct.

Switched to using helpers to have readable input. I also rewrote the malformed UTF-8 test to separate out specific shapes of invalid input and explain what they are.

danderson:
> Since the goal here is to achieve what the current tool in the linter folder is doing, have you considered reading the test files in there?

Hmm, I didn't think of that. Looking at the test files, It's tricky to check for 1:1 match of validation because it's checking for the specific errors the python linter outputs. But this parser definitely needs to cover all these test inputs, and more. For the "classes" of inputs that this parser already handles, the new tests are more exhaustive in every case. Obviously I still need to cover the remaining classes.

Breakdown of the pslint tests:
  - `test_allowedchars` and `test_spaces` are a subset of what this PR tests.
  - `test_section*` is partially implemented: I'm missing an error for duplicate sections, and for missing required sections. Added both to my TODO.
  - `test_{dots,duplicate,exception,punycode,wildcard}` are suffix validations, not yet implemented. Added a note to my TODO that when I implement it, tests need to cover a superset of these inputs.
  - `test_NFKC` is missing, and important to break homoglyph attacks. I need to refresh my memory on Unicode normalization forms, IDNA, homoglyph attacks, and how golang.org/x/text handles normalization, to make sure I implement it correctly. Added to my TODOs.

PR #2011, opened 2024-06-24
tools/internal/parser: detect and report section markers within suffix blocks

danderson:
Small extra validation that I remembered while I was implementing the big text/utf-8 bits that have now been merged.

Note this PR does not pass the tests in tools/internal/parser as-is, because it trips over the expired exemptions that get fixed in #2010. This doesn't break PSL CI right now since tools/internal/parser isn't run there, but I recommend merging #2010 first and then this. The two changes are touching different parts of the codebase, so they should rebase cleanly in either direction.

### DIFF
--- a/tools/internal/parser/errors.go
+++ b/tools/internal/parser/errors.go
@@ -59,6 +59,16 @@ func (e LeadingWhitespaceError) Error() string {
 	return fmt.Sprintf("%s has leading whitespace", e.Line.LocationString())
 }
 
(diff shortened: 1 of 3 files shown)


danderson:
Yup, that's the next change, I'm still fighting with Unicode but should have it tomorrow. It will involve some painful reordering or a bunch of exceptions though, there are ~50 blocks in the wrong order atm. I may be able to reduce that by the time I send the PR though.

Issue #1996, opened 2024-06-14
Recommended Deletions of Expired Domains


groundcat:
## Expired Domains

Below is a list of all private section domains that have expired and are either in the **pending delete** state, where the original owner cannot renew them, or have already been **deleted** by the registry and show available/free status. In either case, they will not serve the purpose for which they were originally requested to be added to the PSL and probably should be removed from the PSL:

- lab.ms
- ro.im
- blogspot.mr
- cn.vu
- instantcloud.cn

## WHOIS Records

Below is the WHOIS record as of today, Friday, June 14, 2024, around 03:00 AM EDT, which indicates the status of these domains. Please advise on the best way to help remove these domains from the list. Since the original registrants won't be able to add the `_psl` TXT record to them, maybe I should create a pull request to remove these domains? Thanks.

WHOIS records for the domain `ro.im`:

```
Domain Name: ro.im
The domain ro.im was not found.
```

WHOIS records for the domain `blogspot.mr`:

```
Domain Name: blogspot.mr
Domain Status: No Object Found
>>> Last update of WHOIS database: 2024-06-14T06:58:43.545Z <<<
```

WHOIS records for the domain `cn.vu`:

```
No Data Found
URL of the ICANN Whois Inaccuracy Complaint Form: https://www.icann.org/wicf/
>>> Last update of WHOIS database: 2024-06-14T06:59:24Z <<<
```

WHOIS records for the domain `lab.ms`:

```
Domain Name: lab.ms
Registry Domain ID: 604616-CoCCA
Updated Date: 2024-05-10T03:02:06Z
Creation Date: 2022-03-10T17:36:19Z
Registry Expiry Date: 2024-05-11T18:25:14Z
First Registration: 1998-09-28T00:00:00Z
Domain Status: active
Domain Status: pending delete
Domain Status: server hold
```

WHOIS records for the domain `instantcloud.cn`:

```
Domain Name: instantcloud.cn
ROID: 20230613s10001s52894822-cn
Domain Status: pendingDelete
Domain Status: clientHold
Domain Status: inactive
Registrant: ******
Registrant Contact Email: ******
Sponsoring Registrar: ******
Registration Time: 2023-06-13 04:03:39
Expiration Time: 2024-06-13 04:03:39
DNSSEC: unsigned
```

simon-friedberger:
I am in favor of removing these. @dnsguru, what's your opinion?

[...]

PR #2014, opened 2024-06-27
PSL Private Section Domains WHOIS Checker

groundcat:
This PR is related to #1996. It introduces the `tools/private_domains_checker`, a Python script created to fetch data from the PSL and check the domain status and expiry dates of the private section domains. It performs WHOIS checks on these domains and saves the results into CSV files for manual review. 

Please feel free to make any edits! 

The [README file](https://github.com/groundcat/PSL-private-domains-checker/blob/main/README.md) has been updated to reflect the usage instructions.

Example CSV outputs from real PSL data:

- [Example datasets](https://github.com/groundcat/PSL-private-domains-checker/tree/main/data)

### DIFF
--- a/tools/private_domains_checker/PSLPrivateDomainsProcessor.py
+++ b/tools/private_domains_checker/PSLPrivateDomainsProcessor.py
@@ -1,67 +1,164 @@
 import datetime
 import json
-import os
+import time
+
 import pandas as pd
 import requests
-import whois
+import whoisdomain as whois
+
+
+def make_dns_request(domain, record_type):
+    """
+    Makes DNS requests to both Google and Cloudflare DNS APIs.
+
+    Args:
+        domain (str): The domain to query.
+        record_type (str): The type of DNS record to query.
+
+    Returns:
+        list: A list containing the JSON responses from Google and Cloudflare.
+    """
+    urls = [
+        f"https://dns.google/resolve?name={domain}&type={record_type}",
+        f"https://cloudflare-dns.com/dns-query?name={domain}&type={record_type}"
+    ]
+
+    headers = {
+        "accept": "application/dns-json"
+    }
+
+    responses = []
+    for url in urls:
+        try:
+            response = requests.get(url, headers=headers)
+            if response.status_code == 200:
+                json_response = response.json()
+                # print(f"URL: {url}, Response: {json_response}")
+                responses.append(json_response)
+            else:
+                # print(f"URL: {url}, Status Code: {response.status_code}")
+                responses.append(None)
+        except Exception as e:
+            print(f"URL: {url}, DNS Exception: {e}")
+            responses.append(None)
+
+    return responses
 
 
 def check_dns_status(domain):
+    """
+    Checks the DNS status of a domain using Google and Cloudflare DNS APIs.
+
+    Args:
+        domain (str): The domain to check.
+
+    Returns:
+        str: The DNS status of the domain.
+    """
+
     def make_request():
-        try:
-            url = f"https://dns.google/resolve?name={domain}&type=NS"
-            response = requests.get(url)
-            json_response = response.json()
-            if json_response.get("Status") == 3:
+        responses = make_dns_request(domain, "NS")
+        if None in responses:
+            return "ERROR"
+
+        google_status = responses[0].get("Status")
+        cloudflare_status = responses[1].get("Status")
+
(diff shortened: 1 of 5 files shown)


simon-friedberger:
The author of the whois package has [declared it unsupported](https://pypi.org/project/whois/), is there an alternative?

Also, I know this would be a big ask but would you be interested in trying to integrating this with the Go validator in https://github.com/publicsuffix/list/tree/master/tools/internal/parser? The idea being, that this can eventually be used to automatically add DNS & whois information to PRs with a Github action. However, that requires a little bit more effort from the parser to determine which sections have changed so only problems in the relevant section can be displayed. But integration would probably just mean providing a function that takes a URL and returns a summary of the status, like "expires in >2y/has expired/...".
This is not super important because the two things do something fairly different but it might still be nice if we could share code here.

simon-friedberger (changes requested):
The author of the whois package has [declared it unsupported](https://pypi.org/project/whois/), is there an alternative?

dnsguru:
> This PR is related to #1996. It introduces the `tools/private_domains_checker`, a Python script created to fetch data from the PSL and check the domain status and expiry dates of the private section domains. It performs WHOIS checks on these domains and saves the results into CSV files for manual review.
> 
> Please feel free to make any edits!
> 
> The [README file](https://github.com/groundcat/PSL-private-domains-checker/blob/main/README.md) has been updated to reflect the usage instructions.
> 
> Example CSV outputs from real PSL data:
> 
> * [All domains](https://github.com/groundcat/PSL-private-domains-checker/blob/main/data/all.csv)
> * [NXDOMAIN](https://github.com/groundcat/PSL-private-domains-checker/blob/main/data/nxdomain.csv)
> * [Expired](https://github.com/groundcat/PSL-private-domains-checker/blob/main/data/expired.csv)

Separate from the dialog here, I noted in an issue in @groundcat 's repo that identifying the [client|server]Hold status domain names into a separate file would be beneficial.  Domains with either of those statuses almost always get that status for a reason that would make the domain something that should not be on the PSL, PLUS the domain would be NXD as those statuses cause the domain name to not be listed with NS delegation in their TLD zone files.

tools/private_domains_checker/PSLPrivateDomainsProcessor.py
```diff
@@ -0,0 +1,206 @@
...
+import os
+import pandas as pd
+import requests
+import whois
+
+
+def check_dns_status(domain):
+    def make_request():
+        try:
+            url = f"https://dns.google/resolve?name={domain}&type=NS"
```
dnsguru, line 12: Google is a good source for this, but I do want to mention that this places authority on to Google - they could theoretically intervene in the resolution process on a given record.

Would we perhaps want to use an array of such resolvers that are randomly selected from?

groundcat: **Updated** - Good point about relying solely on Google for DNS, as it centralizes authority and potential intervention risks. Sometimes, it gives errors due to random reasons such as network fluctuations, so perhaps randomly selecting from multiple resolvers is still not error-proof. So, I decided to query two resolvers (Google and Cloudflare) and only take the value if they are consistent across the two: the updated implementation uses both Google and Cloudflare DNS resolvers. It queries both resolvers and only considers the result usable and consistent if both return the same status. If the results are inconsistent or if there are errors, it re-retries up to 5 times.

groundcat:
> The author of the whois package has [declared it unsupported](https://pypi.org/project/whois/), is there an alternative?

Thank you for spotting this issue. I have replaced it with the `[whoisdomain](https://pypi.org/project/whoisdomain/)` package, which is recommended by the author of the retired whois package.

groundcat:
> I noted in an issue in @groundcat 's repo that identifying the [client|server]Hold status domain names into a separate file would be beneficial

Thanks for the input. I added a new filter to get a CSV list of domains with any form of hold status, and another filter for CSV files with domains expiring within 2 years. I guess the latter might not be very useful at the moment since a handful of them are expiring less than 2 years and were probably submitted before the requirement was established, so they might not be aware of it, similar to the requirement for keeping the _psl TXT records at all times.

simon-friedberger:
My stance is that we want to be strict on the _psl DNS entry but lax with the expiration times because it's often impossible to check for us and often impossible to get >2y for the requester.

simon-friedberger (approved):

PR #696, opened 2018-07-22
Added instantcloud.cn by Redstar Consultants

alexanderschulze:
Redstar Consultants provides various services to companies starting in or expanding their business to China. Domains within this suffix are used to provide clients a method to internally validate their business project on Chinese cloud hosting services usually before being hosted on their own domain and/or being made publicly accessible.

Being added to the public suffix list is required to avoid lets encrypt limitations and to aim for a clear separation between multiple client projects.

PR #2015, opened 2024-06-27
Remove instantcloud.cn

alexanderschulze:
The domain got repurposed and should not be listed in the PSL anymore. The `_psl` DNS record was deleted.

Initially, it has been added in https://github.com/publicsuffix/list/pull/696.


dnsguru (approved):
Domain is NXD for extended period.  Removal of domain and section appropriate

PR #2019, opened 2024-06-28
Update comments on aland.fi

simon-friedberger:
Changes only a comment


dnsguru (commented):
I like listing the iana page link for both namespaces like this, updating it from the legacy wikipedia entry.

PR #994, opened 2020-03-16
Add g.vbrplsbx.io domain

toruta39:
* [x] Description of Organization
* [x] Reason for PSL Inclusion
* [x] DNS verification via dig
* [x] Run Syntax Checker (make test)

Description of Organization
====

Organization Website: <https://dev.viberplay.io>

Viber Play is a game platform running in the messenger platform Viber <https://chats.viber.com/rakutengames>. It helps game developers to develop and publish HTML5 game content inside Viber.

Viber Play is developed and operated by Rakuten Games, Inc. <http://corp.rakuten-games.com/>

Reason for PSL Inclusion
====

Subdomains under g.vbrplsbx.io are assigned to different game developers for hosting their HTML5 game contents.

We want to protect users from being tracked across games via cross-subdomains cookie/state sharing.

DNS Verification via dig
=======

```
dig +short TXT _psl.vbrplsbx.io
"https://github.com/publicsuffix/list/pull/994"
```

make test
=========

```
============================================================================
Testsuite summary for libpsl 0.21.0
============================================================================
# TOTAL: 5
# PASS:  5
# SKIP:  0
# XFAIL: 0
# FAIL:  0
# XPASS: 0
# ERROR: 0
============================================================================
```

PR #2022, opened 2024-07-01
Delete Rakuten Games, Inc related entries

toruta39:
Rakuten Games, Inc. was dissolved in early 2024. And related services are already closed. Due to the organization's domains can be now obtained by any other party (which can be potentially malicious). I want to request the deletion from PSL.

FYI The record was added in https://github.com/publicsuffix/list/pull/994


simon-friedberger:
Same submitter as original PR and organization website is unreachable.

PR #2023, opened 2024-07-01
WIP PR template

simon-friedberger:
The idea here would be that we add a note like this now and then also require a line saying "Acknowledged Removal Warning (v1)" to the organization header in the .dat file.

For exampled

*Note*
```
<!--
To keep the PSL free of outdated entries it is necessary to be able to remove entries. We may at some point request renewal by e-mail or start to automatically remove entries which do not have DNS validation anymore. The grace period for failures is 4 weeks.
-->
  * [ ] The submitter acknowledges the following **Removal warning (v1)**: Failure to retain the _psl DNS validation entry or failure to respond to e-mails to the supplied address may result in removal from the PSL.
```

*PSL entry*

```
// tawk.to, Inc : https://www.tawk.to
// Submitted by tawk.to developer team <dev-accounts@tawk.to>
// Acknowledged Removal Warning (v1)
p.tawk.email
p.tawkto.email
```

That would give us a known way to remove entries for new additions.

@dnsguru What do you think?
@danderson Do you have opinions on the format?

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -67,6 +67,13 @@ the PR.
 
   * [ ] This request was _not_ submitted with the objective of working around other third-party limits
 
+
+<!--
+To keep the PSL free of outdated entries it is necessary to be able to remove entries. We may at some point request renewal by e-mail or start to automatically remove entries which do not have DNS validation anymore. The grace period for failures is 4 weeks.
+-->
+  * [ ] The submitter acknowledges that failure to retain the _psl DNS validation entry may result in removal from the PSL.
+  * [ ] The submitter acknowledges that failure to respond to e-mails to the supplied address may result in removal from the PSL.
+
 <!--
 The guidelines describe which section to place the entry, what the 
 order of commented org placement, order of sorting of entries. 
@@ -78,6 +85,7 @@ neither.
 
   * [ ] The [Guidelines](https://github.com/publicsuffix/list/wiki/Guidelines) were carefully _read_ and _understood_, and this request conforms
   * [ ] The submission follows the [guidelines](https://github.com/publicsuffix/list/wiki/Format) on formatting and sorting
+
 
 <!-- 
 Sorting and formatting of the entries is outlined in the guidelines 
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -86,7 +86,6 @@ neither.
   * [ ] The [Guidelines](https://github.com/publicsuffix/list/wiki/Guidelines) were carefully _read_ and _understood_, and this request conforms
   * [ ] The submission follows the [guidelines](https://github.com/publicsuffix/list/wiki/Format) on formatting and sorting
 
-
 <!-- 
 Sorting and formatting of the entries is outlined in the guidelines 
 and non-conforming requests are one of the largest sources of delay,
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -67,12 +67,10 @@ the PR.
 
   * [ ] This request was _not_ submitted with the objective of working around other third-party limits
 
-
 <!--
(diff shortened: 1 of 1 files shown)


dnsguru:
TBH this seems like a lot of extra work on the header element, and I am not certain it would have the attention of the submitters, nor would I rely on email - we might end up with deliverability challenges.

Can I suggest we would have the template 'tickbox' be an acknowledgement that they put the entire section of the submitter in jeopardy of removal should they fail to maintain all of the domain names within it?

This way, they are less prone to "I got mine, move on" or "set and forget" attitudes that have been in play, and instead it is on them to maintain current / healthy status on the names within their section(s).



```
<!--
Submitter will maintain domains in good standing or lose section.   In making this pull request, there is a commitment by the submitter that they are going to review and maintain their relevant section, including removal of expiring or unused domains as they occur, and that their entire section may be removed should they fail to do so.  The ongoing trust of the PSL requires it to be free of outdated or problematic entries, such as those which expire or do not resolve in the DNS or contain the relevant _PSL TXT record that accompanied an entry's Pull Request, or other criteria as needed.  By submitting an entry, the requestor acknowledges that their entry and section will be removed if the domain does not maintain the respective _PSL entr(y/ies) in DNS, any the domain(s) within their section fail to resolve in DNS, the domain does not get renewed, expires or is otherwise unreachable.  Submitter further identifies that it is their responsibility to review their submitted section within the PSL, submitting updates or removals as their domain(s) may change over time.  It is also the responsibility of the submitter to provide (and keep up to date) a reachable email address within the section, and to maintain that address as it may change over time, so that they receive notices.
-->
  * [ ] The submitter acknowledges their responsibility to, while PSL-listed, maintain the domains within their section, and remove names that they are no longer using, and that their entire section may be removed should an entry within their section become expired, unreachable, or unverifiable using DNS.
```

dnsguru:
>I don't think the PSL has enough public visibility to pressure companies into maintaining their domains better. If we punish them by removing entries mostly the user's security will suffer. Plus, they will come back to us and want to get added again and I don't think we want to deal with that.

Probably lost in translation or I didn't say it well... This was more intended for the tickbox to inform a new PR (or add) that it is on them to keep their stuff up to date and they'd have consequences if not.  Making their whole section potentially at risk would diminish the entirely-too-casual attitude that clogs us up with their debris.

simon-friedberger:
Oh, did you mean we just say the entire section is at risk but we don't actually remove the entire section for now?

dnsguru:
Currently, without the additional tickbox you are suggesting, it seems like we (volunteers + community) are treating things like PR submitters are entitled to an entry, and that sense of entitlement brings with it a casual attitude after achieving their listing.

> Oh, did you mean we just say the entire section is at risk but we don't actually remove the entire section for now?

It would be at our discretion.  The submitter really should have more respect for the privilege of being listed and be accountable to remove entries after themselves and maintain thier entry in good working order.  It introduces responsibilities upon the submitter that comes along with obtaining the benefits that come with having their listing included.

dnsguru (approved):
LGTM

dnsguru:
I am cautiously hopeful that this change will work to diminish abandonment debris in the PSL.

Folks may apply as much review as they do to clicking through a click-wrap agreement to get their itunes music, but hey, we tried :)

danderson:
I defer to y'all about the wording and all that, this sounds fine to me. These are all things I would like to start enforcing, and we can take a snapshot of the PSL as of this merge to build a legacy exemption list (in addition to scanning DNS records and stuff to narrow it down, of course)

danderson:
Oh and from an automation POV: I don't need a marker in the .dat file to do that, once this is official policy and people have to ack this rule in new PRs, I can grab the list of preexisting submissions and make automation be more lenient with them, and we can either leave it like that, or slowly track people down and shrink the exemption list over time.

But as long as the new policy lets me say "the suffixes that are present at commit XYZ are all the exempted ones, everything after that is subject to the automated enforcement", that's very straightforward for me to implement :+1:

PR #2027, opened 2024-07-05
tools/internal/parser: rework metadata extraction for more accurate reformatting

danderson:
Several commits, each one is self-contained.

I promise the formatting is coming soon, this rework is making the metadata extraction more precise, so that we can look at options for reformatting the suffix block headers into some canonical form.

I added some tests specifically for metadata parsing, to cover the layouts that exist in the PSL. I also removed a couple of parser tests that were just exercising metadata parsing, and were testing behavior for formats that aren't actually present in the real PSL :facepalm: 

Next PR after this one is `pslfmt` with sorting of suffixes and suffix blocks, and possibly metadata reformatting if we find a format that we like.

### DIFF
--- a/tools/govalidate/govalidate.go
+++ b/tools/govalidate/govalidate.go
@@ -95,15 +95,18 @@ func debugPrintRec(b parser.Block, indent string) {
 		}
 		f("}")
 	case *parser.Suffixes:
-		items := []string{loc}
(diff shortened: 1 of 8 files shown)


timscharfenort8 (commented):
good

danderson:
Added a few more no-op cleanup commits. Each commit can still be reviewed mostly independently to reduce the overall size of the review.

PR #2029, opened 2024-07-06
aero: remove extra word between TLD name and URL

danderson:
This is the only TLD that isn't in the form "<tld> : <url>". It's preferable to make it match the others rather than force metadata parsers to work around it.

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -32,22 +32,30 @@ ac.ae
 gov.ae
 mil.ae
 
-// aero : https://www.information.aero/index.php?id=66
+// aero : https://information.aero/registration/policies/dmp
 aero
+// 2LDs
+airline.aero
+airport.aero
+// 2LDs (currently not accepting registration, seemingly never have)
+// As of 2024-07, these are marked as reserved for potential 3LD
+// registrations (clause 11 "allocated subdomains" in the 2006 TLD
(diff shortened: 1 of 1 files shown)


dnsguru:
That one is very ancient/legacy... Just male sure the url for registration policy is functional and current if you are updating the comment line

danderson:
Good call, thank you. I was aiming for the minimal change, but if I'm in the area anyway...

I've added a second commit to the PR with a full update to this block. The commits are both individually well-formed and can be merged separately, or squashed. I'm leaving them separate initially to facilitate incremental review.

The second commit makes the following updates:
  - Updated link to the Domain Management Policy. The policy appears unchanged, they just redid their site at some point and broke URLs.
  - Updated list of 2LD public suffixes, from the table in clause 11 of the DMP. Aside from a sorting order change, this removes `amusement.aero` (I could not find any reference to this ever being a reserved or open-for-registration 2LD), and adds `freight.aero`, `marketplace.aero` and `taxi.aero`.
  - Split the 2LD list into active and inactive subsections. The DMP indirectly provides evidence of which 2LDs are active, by requiring them to list their registration policy in the TLD's policy. From my digging, only `airport.aero` and `airline.aero` have ever been active, the others have been parked and unused for 18+ years. I left them listed defensively, since .aero doesn't provide any automated feed we could use to discover a 2LD activation.

Issue #2028, opened 2024-07-06
Automatic PSL formatting: what fixes should be applied?


danderson:
I have a WIP that can automatically fix formatting issues in a PSL file, as well as apply machine edits while preserving correct formatting. It will take a few more PRs to merge it all, but you can peek at https://github.com/danderson/list/tree/danderson/psl-parser-format .

The idea is to give contributors a tool that they run before sending a PR, and it (a) tells them everything that they need to fix, but also (b) just automatically fixes things that don't require human judgement, like "did you sort these strings correctly". This also makes automated editing easier, because scripts can operate on an abstract syntax tree, and not have to worry about exactly how to lay out text bytes. Just adjust AST nodes, then validate+format+write the file back out and get a clean diff.

**I'm opening this issue to get feedback from PSL maintainers: what formatting fixes do you want me to implement?** @dnsguru @simon-friedberger 

I've already implemented/thought of several, listed below. Each one requires a one-time PSL reformat to fix existing issues, with diffs ranging from <10 lines to >1000 lines. After that, we can require new contributions to run the formatter before sending the PR (and enforce it in CI), and there will be no random churn.

Each of the listed things is independent from the others, you can pick any combination. These changes are all purely cosmetic, and do not change the set of public suffixes, or the meaning of any comments (modulo bugs of course, but I'm confident I can reduce the risk of that with tests, pre/post format sanity checks, etc.).

1. **Basic consistency edits**: remove leading/trailing whitespace, list wildcard exceptions immediately after corresponding wildcard, consistent inter-block spacing: [patch](https://gist.github.com/danderson/289d1fdeb456ac06bd0094ac6e7c4f4c#file-1-no-cleanups-patch)
    - Technically this one does 3 independent things right now. I can separate them if you only want a subset. These are the changes that are more costly to _not_ do, because avoiding this diff means I have to make the parser more complex to preserve the inconsistencies.

2. **Sort blocks in the private section**, by company/product/owner name. This automates #1858, or at least one half of it. Somewhat large change, but not unjustified: all those blocks are indeed in the wrong order today. [patch](https://gist.github.com/danderson/289d1fdeb456ac06bd0094ac6e7c4f4c#file-2-sort-private-section-patch)

[...]

PR #2036, opened 2024-07-12
tools/internal/parser: implement automatic reformatting

danderson:
Each commit is logically separate, and passes all tests, if the smaller chunks to review helps.

The huge diffstat is entirely caused by importing the Unicode Consortium IDNA test data, to verify that domain names are being parsed correctly. It's a large ascii file that doesn't change much (<20 lines per 2-3 years), and I imported it verbatim (via a `go generate` helper) to make it easy to check and update.

With this PR, if you run `govalidate public_suffix_list.dat`, and the file has no serious parsing issues, it will reformat the file in-place in addition to reporting errors that require a human to fix. You can also `--reformat=false`, and it'll just report an error if the formatting is incorrect (for CI).

Reformat diff for current .dat is: https://gist.github.com/danderson/7f8d25a864b3fd9b9d1f58df6cfe291f . It's option 1+2+3 from #2028. I verified that the .dat contains the same suffixes before/after this patch.

Planned/WIP followups to this PR:
 - reformat of the PSL .dat to match the output of govalidate. I'm going to send the reformatting diffs as separate PRs.
 - diff support. I have this mostly working already, I can take two PSL files and only show issues/run expensive validations like whois for changed regions.
 - clean up the CLI, I made govalidate as a little demo but it needs to be more human friendly and have things like "please run checks on PR 1234" as subcommands.
 - hook it up to CI and start checking new submissions, maybe? :crossed_fingers:


tools/internal/parser/file.go
```diff
@@ -192,15 +194,15 @@ func (s *Suffix) Children() []Block { return nil }
...
 	// leading "*" label.
-	Labels []string
-	// Exceptions are the DNS label values that, when they appear in
-	// the wildcard position, cause a FQDN to _not_ match this
-	// wildcard. For example, if Labels=[foo, com] and
-	// Exceptions=[bar, qux], zot.foo.com is a public suffix, but
-	// bar.foo.com and qux.foo.com are not.
-	Exceptions []string
+	Domain domain.Name
+	// Exceptions are the domain.Labels that, when they appear in the
```
simon-friedberger, line 205: autocorrect typo?

simon-friedberger: Oh, this is referring to the other property.

tools/internal/parser/clean.go
```diff
@@ -0,0 +1,400 @@
...
+	return errs
+}
+
+func sortSuffixes(s *Suffixes) []error {
+	// Suffix sorting has the same problem as section sorting: inline
+	// comments act as barriers that prevent movement of a suffix
+	// across them (e.g. "all 3rd-level public suffixes come after
+	// this").
+	//
+	// We do the same thing as for section sorting: sort as much as we
```
simon-friedberger, line 219: Do the comments stay attached to the other lines?

Your example is
```
foo.example.com
bar.example.com
// non com domains
foo.example.org
*.example.pl
internal.example.rs
```

but what about
```
foo.example.com
// next is our bar!
bar.example.com
abc.example.com
```

In this case it wouldn't be right to sort `abc` before `bar` and keeping the comment on top.

simon-friedberger:
LGTM. I wouldn't bother with the e-mail obfuscation and just de-obfuscate it. This kind of obfuscation does probably nothing to stop e-mail scrapers these days.

danderson (commented):
I removed the obfuscation commit from this PR. In the current implementation the existing entry will be stable until its metadata needs to be changed, we can always reexamine when we get to the point of making format changes.

simon-friedberger:
Hm, if the ICANN section has both kinds of comments I have a slight preference for the last option: Make the comments which start subgroups have a different format. For now, can you just disable sorting for the ICANN section so that we don't accidentally mess it up at some point? Then we can put whatever else we do in a different PR.

tools/internal/parser/clean.go
simon-friedberger: Hm, if the ICANN section has both kinds of comments I have a slight preference for the last option: Make the comments which start subgroups have a different format. For now, can you just disable sorting for the ICANN section so that we don't accidentally mess it up at some point? Then we can put whatever else we do in a different PR.

danderson: Sounds good. The code already skipped all sorting in the ICANN section, so if you're ok with the current logic for the private section (temporarily, while we implement the new comment format), the PR is good as-is. The private section has no inline comments atm so the sorting logic is a no-op that we can change freely.

dnsguru: > Hm, if the ICANN section has both kinds of comments I have a slight preference for the last option: Make the comments which start subgroups have a different format. For now, can you just disable sorting for the ICANN section so that we don't accidentally mess it up at some point? Then we can put whatever else we do in a different PR.

For historical context about the diversity (AKA Sloppy-ness) in the ICANN section, the sub-section stuff became necessary in situations where the IANA Delegee would in turn have a delegee (Example: for .uk, Nominet runs the TLD and most of the SLD subspace like .co.uk, but also there are some which are still administratively part of their delegation but run by other entities like JANET).

The we had the .jp namespace with all the prefectures and their specifically defined structures, which needed annotation as defined by JPRS - which needed a slightly different approach.  Those are just two examples, we have had a number of different approaches used;   When I first started cleaning that up, it would be upon the TLD administrator requesting so, and we'd update their section to whatever the style in use at the time was.

There's some opportunity for cleanup there, but the sub-comments within a 'section' are helpful for context for readers to understand the quite diverse namespace approaches in the ccTLD world.

When we do start to look at the ICANN section, one thing we might want to do is engage with people from their Office of the CTO (OCTO) team and or coordinate with the IANA about it.

That is a much much larger topic with much much larger nuance, and it should be done with elegance (scalpel vs cudgel)

PR #2052, opened 2024-07-23
remove `pcloud.host`

wdhdev:
This domain returns a for sale page, which is against the PSL's terms of service.

image


pzduniak:
I support this change, though I am not involved with the company anymore. I am nearly certain Nucleos Inc. does not operate any such service anymore.

cc @noahfreedman

wdhdev:
That seems good enough to merge as it isn't used anymore, cc @simon-friedberger :)

PR #404, opened 2017-02-21
Add cya.gg to the private section.

Codeusa:
cya.gg is the routing domain that provides users of the FOSS project [Ulterius ](https://ulterius.io) with sub-domains that point to their self-hosted instances. Their sub-domain is generated when an account is registered, the software handles keeping their IP synced to their domain. 

A example domain for instance is: 2aafcc59.cya.gg, which would point to the public address of their server. 

The addition of cya.gg to the public suffix list will help ensure new and current users of Ulterius are able to keep their domains isolated from one another.

PR #2053, opened 2024-07-23
remove `cya.gg`

wdhdev:
This domain is setup to point to a parking page, which I believe is against the ToS:
image

It was originally added to the PSL in #404 however it seems the domain name registration lapsed and someone else took the domain name as the registration date via a WHOIS lookup shows `Registered on 15th November 2023 at 14:29:51.068`:
image

The `_psl.cya.gg` TXT record no longer exists either, so that is also reason to believe the domain name has changed registrants.

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -14615,10 +14615,6 @@ obl.ong
 observablehq.cloud
 static.observableusercontent.com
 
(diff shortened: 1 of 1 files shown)


andrewmd5:
I can confirm as the one who added it that the domain can be removed from the list. I apologize for not getting to that sooner.

PR #2054, opened 2024-07-23
remove `nid.io`

wdhdev:
Domain is up for sale, which is against the PSL terms of service:
image

Nameservers of `nid.io` point to Afternic:
image

---

I am just trying to clean up some entries on the PSL that are invalid with some spare time I have as I really appreciate this project. Would you guys prefer me make one big PR with changes like these (with evidence for removal of each domain, of course) instead of individual PRs for each domain/section? Thanks :)

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -14653,10 +14653,6 @@ simplesite.pl
 123paginaweb.pt
 123minsida.se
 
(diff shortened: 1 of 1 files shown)


simon-friedberger:
I prefer individual PRs but if it's easier for you, and the cases are as cut-n-dry as this feel free to group them.

PR #2058, opened 2024-07-23
remove old domains

titanous:
Removes two domains that I am responsible for that are no longer used as public suffixes.


simon-friedberger:
@titanous I've sent you an email to confirm because you don't seem to be the original submitter for cupcake.is or that connection got lost during a migration.

titanous:
I no longer have control over `cupcake.is`, it was transferred to a new owner in 2022.

PR #2088, opened 2024-08-08
Run 'psltool fmt' to reformat PSL to canonical form

danderson:
These are all pure block movements, no suffix or metadata changes. Just out of paranoia, I verified the change with some extra manual steps:
 - `ripgrep -v '^//' | ripgrep -v '^$' | sort >suffixes.txt` before and after the reformat shows no diffs, IOW exactly the same suffixes are present.
 - `sort <public_suffix_list.dat >lines.txt` shows only 1 diff, post reformat the file has 1 blank line less than the original, no changes to any comment or suffix text.
 - `psltool validate public_suffix_list.dat` reports that the PSL file is valid, i.e. no formatting issues and no issues with the validations implemented so far.
 - `make test` of course. All tests pass.
 - After opening the PR, `psltool check-pr 2088` (with code from #2087) reports that the file is valid and has no changed suffix blocks. IOW, the diff logic saw that block ordering changed, but the semantic content of the file did not.

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -2010,18 +2010,18 @@ yamanashi.jp
 // jp geographic type names
 // http://jprs.jp/doc/rule/saisoku-1.html
 *.kawasaki.jp
-*.kitakyushu.jp
-*.kobe.jp
-*.nagoya.jp
-*.sapporo.jp
-*.sendai.jp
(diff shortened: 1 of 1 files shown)


simon-friedberger:
I ran my old check script as well and had a quick look. Looks good.

PR #2093, opened 2024-08-09
Fix newline handling of automatic ICANN updater

simon-friedberger:
This fixes the newline handling in the ICANN update script to match the new format.


simon-friedberger:
@danderson Does this look reasonable to you?

danderson (commented):
Yup looks reasonable to me!

PR #2099, opened 2024-08-11
remove `mc.eu.org`

wdhdev:
This domain name I believe was originally added along with all of the other eu.org suffixes (pre moving to GitHub?)

There are no nameservers under this subdomain nor does there seem to be any registered subdomains for it.

If you search Google using `site:mc.eu.org`, there is no results. The domain is not listed on the eu.org registration website either: https://nic.eu.org/opendomains.html

After running a SSL certificate lookup, no certificates were found: https://crt.sh/?q=mc.eu.org

I believe there is enough evidence to support the idea that this subdomain is not used nor is active.

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -13109,7 +13109,6 @@ kr.eu.org
 lt.eu.org
 lu.eu.org
 lv.eu.org
-mc.eu.org
 me.eu.org
 mk.eu.org
 mt.eu.org


groundcat:
According to the "[Open domains and policies](https://nic.eu.org/opendomains.html)" page at eu.org:

> If the country (or other, like NET.eu.org for example) of your dreams is not in the list below, and you can provide name service and administration for it (this is now a required condition to avoid creating many almost empty country domains that we then have to manage), you can write to [hostmaster@EU.org](mailto:hostmaster@EU.org) to request that it is created and delegated to you. Note that, in that case, it must obviously be open to everyone.

Assuming "mc" means the Principality of Monaco, it could fall under this rule and might be once open for registration some years ago. However, I agree that it appears this suffix is no longer in use currently. It would be great to get confirmation from the operator of eu.org (Pierre Beyssac?), as eu.org is still an active project and might intend to bring it back for open registration in the future.

simon-friedberger:
@pbeyssac Would you mind having a look? Can we remove mc.eu.org?

simon-friedberger:
Is eu.org generally gone? Maybe this should have been `*.eu.org` in the first place.

@pbeyssac Any comments?

wdhdev:
`*.eu.org` would not work as eu.org does issue subdomains at the 3rd level (e.g. `exampledomain.eu.org`) which would in result in those subdomains being declared as TLDs which would break some services such as Cloudflare which does not allow zone additions for "TLDs" included in the PSL (e.g. `web.example.eu.org` would work, however `example.eu.org` wouldn't if eu.org was using a wildcard.)

simon-friedberger:
I guess that point is moot if it doesn't come back.

wdhdev:
What do you mean? As in the service is effectively dead?

simon-friedberger:
Yes, it appears to be down for me.

wdhdev:
https://nic.eu.org/ loads for me.

groundcat:
The registry site https://nic.eu.org/ loads for me, but very slowly. Based on historical snapshots from archive.org, it's mostly online, so this project seems to be active, at least for existing users.

image
(Blue dots indicate the website returns a 200 status code)

However, it seems they've stopped reviewing new registrations and are not responding to emails, according to multiple user reports linked below. The eu.org operator has gone silent and hasn't been heard from in a while. The current status is unknown, but the existing eu.org nameservers are still functioning properly for now.

- https://www.reddit.com/r/selfhosted/comments/1av6awf/euorg_validation_speed/?rdt=50279  
- https://www.reddit.com/r/selfhosted/comments/1asfoc4/eu_org_domain_creation_questions/

As @wdhdev mentioned, we can't add them as `*.eu.org` because `example.eu.org` was open for registration, and most registrants are using the `eu.org`-level domains. Using a wildcard would disrupt things for existing `eu.org`-level users.

https://www.google.com/search?q=site%3Aeu.org

I suspect the slowness and downtimes of https://nic.eu.org/ may be due to a high amount of brute-force hacking activities in recent months since the project went silent, according to a security warning thread on a China-based forum: https://linux.do/t/topic/157058. In this thread, a malicious party is reportedly brute-forcing user accounts in batches to take over their domains, and multiple users have reported that their domains were transferred out without authorization. Similarly, other user reports, like this one on Cloudflare (https://community.cloudflare.com/t/my-website-is-getting-redirected-to-camp/587607/4), mention domains being taken over. This makes me question the current state of the eu.org project.

wdhdev:
The site loads normally, it's not slow at least for me. I can confirm they are not responding to emails nor domain registration requests from personal experience. 

It sort of seems the project is completely dead.

wdhdev:
@simon-friedberger It should be safe to remove this specific entry as it does not have any subdomains when looking it up on https://subdomainfinder.c99.nl, as well as results are not returned on Google when searching `site:mc.eu.org` either.

PR #300, opened 2016-09-12
Add Revitalised Limited private domains

JackPrice:
Domains under `wellbeingzone.co.uk`, `wellbeingzone.eu` and `ptplus.fit` are issued to Revitalised Ltd. customers to serve our SaaS instances.

For example; `customer.wellbeingzone.co.uk`.

DNS verification will be added once I get the pull request ID.

PR #2101, opened 2024-08-11
remove Revitalised Limited section

wdhdev:
Both domains were added in #300, however since their service has shutdown and one of the domain names expired nearly a month ago and I believe they are going to let the other one expire as well.

## wellbeingzone.co.uk
This domain expired 2024-07-13, 28 days ago as of writing this.

It seems like they are not going to renew it as it has been that long since and the WHOIS server returns `Renewal required.`

When looking up the domain on Google using `site:wellbeingzone.co.uk`, all that is returned is:
image

## wellbeingzone.eu
The registry does not seem to return the domain expiry date for this domain via a WHOIS lookup, however I would estimate it will expire fairly soon.

When going to the site it shows a service shutdown page:
image

Running a subdomain scan returns no results: https://subdomainfinder.c99.nl/scans/2024-08-11/wellbeingzone.eu

When searching `site:wellbeingzone.eu` on Google, this is al that is returned:
image

The second result returns a 403 error for me:
image

I believe on these grounds both domain names should be eligible for removal as it seems neither of them are in use and once of which has expired.

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -14928,11 +14928,6 @@ devices.resinstaging.io
 // Submitted by Chris Kastorff <info@rethinkdb.com>
 hzc.io
 
-// Revitalised Limited : http://www.revitalised.co.uk
-// Submitted by Jack Price <jack@revitalised.co.uk>
-wellbeingzone.eu
-wellbeingzone.co.uk
-
 // Rico Developments Limited : https://adimo.co
 // Submitted by Colin Brown <hello@adimo.co>
 adimo.co.uk


groundcat:
- `wellbeingzone.co.uk` has expired according to WHOIS and is for sale.
- `wellbeingzone.eu` is likely repurposed.

I agree with your assessment that it is probably not in use. It would be great if @jackprice could confirm whether these entries are safe to be removed.

jackprice:
Unfortunately I no longer work for the company involved so I can't confirm or deny, apologies!

wdhdev:
> Unfortunately I no longer work for the company involved so I can't confirm or deny, apologies!

Do you know anyone that does work there still? If not, I personally would say we can remove these.

Issue #1681, opened 2023-01-14
Tracking SAC113 on Private-Use TLDs - creation of `.INTERNAL`


dnsguru:
Back in the day, RFC1918 defined a number of IP address ranges that were not to be routed, such as 192.168.1.1 or 10.0.0.1 which became widely used for Network Address Translation.  There was not really any similar type of Top Level Domain defined in a similar manner, and this is being explored in order to avoid collisions from a diversity of ad-hoc approaches that have caused issues over the years.

ICANN's Security and Stability Advisory Committee has published a document related to TLD namespaces that might be used in private networks, with advice related to them.  No domain names were recommended, this is just being shared as a placeholder to ensure it is something that stakeholders here in the dev community are aware of as it evolves and some TLDs are perhaps defined.

https://www.icann.org/en/system/files/files/sac-113-en.pdf 

These MAY or MAY NOT end up being defined within the PSL.  Not without some discussion, but that discussion should likely occur after more information is available, with abundant consideration for being non-disruptive to status quo expectations and functionality wherever possible.

dnsguru:
As a follow-up to this, ICANN is opening a [comment period ](https://www.icann.org/en/public-comment/upcoming-proceedings)related to the introduction of a private-use TLD called INTERNAL shortly...  Before that introduction would happen, or typically any significant changes, ICANN open comment periods to allow for inputs from the community.  Often the comment periods quietly appear and quietly disappear for those not closely monitoring their existence.   So I raise this to the attention of the project stakeholders to not get caught off guard.

Their timeframe on opening a comment period states "January 2024"

Their description is:

>IANA has made a provisional determination that “.INTERNAL” should be reserved for private-use and internal network applications. This string was identified according to a previously consulted procedure on implementing SAC113, produced by the SSAC. Feedback is sought on whether the identified string meets the requirements of SAC113, or whether there are any other issues with the string that need to be considered prior to its reservation.

**Browsers and other stakeholders:**  Would this belong in the PSL or not?  Things to consider as part of the comment period and perhaps even popping your head into the "ICANN-o-SPHERE" and commenting might be to consider including a subdomain of ".INTERNAL" that might be added to PSL so that it would/could be used for cookie-related stuff.  Or Not.

The comment period is opening soon.

dnsguru:
The [Comment Period](https://www.icann.org/en/public-comment/proceeding/proposed-top-level-domain-string-for-private-use-24-01-2024) is now open

[...]

PR #2104, opened 2024-08-13
remove `localzone.xyz`

yahesh:
`localzone.xyz` was added in https://github.com/publicsuffix/list/pull/915 with the intention to introduce a domain that can be used for internal services across different organizations without the risk of collisions.

With the latest [board decision of the ICANN](https://www.icann.org/en/board-activities-and-meetings/materials/approved-resolutions-special-meeting-of-the-icann-board-29-07-2024-en#section2.a) to finally reserve the `.internal` TLD for this specific use-case, `localzone.xyz` is no longer needed and the use of `.internal` is recommended instead.

I therefore intend to have `localzone.xyz` lapse in the future. I already removed the corresponding `_psl` record to indicate that this PSL entry should be removed.

Sorry for the rollback and thank you very much for your hard work. 🙏

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -14145,10 +14145,6 @@ ggff.net
 // Submitted by Lann Martin <security@localcert.dev>
 *.user.localcert.dev
 
-// localzone.xyz
-// Submitted by Kenny Niehage <hello@yahe.sh>
-localzone.xyz
-
 // Log'in Line : https://www.loginline.com/
 // Submitted by Rémi Mach <remi.mach@loginline.com>
 loginline.app


dnsguru:
#1681

dnsguru:
@yahesh the challenge is with the addition of .internal is that it will not be added to the root, by design.  So the behaviour of localzone.xyz might still be materially different than the .internal TLD with respect to actual global reachability.

Worth considering this before pivoting.

yahesh:
> @yahesh the challenge is with the addition of .internal is that it will not be added to the root, by design. So the behaviour of localzone.xyz might still be materially different than the .internal TLD with respect to actual global reachability.
> 
> Worth considering this before pivoting.

Yes, this has already been considered. Actual global reachability never relevant for the `localzone.xyz` domain. It was always meant to be used internally only (i.e. in an internal split-horizon DNS setup).

dnsguru (approved):
* remover = submitter
* intentional, low/no consequences
* tests pass
* no conflict with base
APPROVED

PR #766, opened 2019-02-05
Add dweb.link

mburns:
* [x] Description of Organization
* [x] Reason for PSL Inclusion
* [x] DNS verification via dig
* [x] Run Syntax Checker (make test)

Description of Organization
====

Organization Website: https://protocol.ai

Protocol Labs develops [IPFS](https://ipfs.io/), which allows end-users to resolve content they have published on the IPFS network using content-based addressing. This content can be resolved by web browsers and other clients in the form `http://<hash>.ipfs.dweb.link` or `http://<peerid>.ipns.dweb.link`.

I am an engineer at Protocol Labs and represent the Infrastructure team, responsible for hosting the IPFS Gateways that provide this proxying service. Note that this PR was made as a fork on my personal GitHub account, not the Protocol Labs GitHub organization.

Reason for PSL Inclusion
====

Because end-users can publish arbitrary content, we would like to sandbox every subdomain as much as possible (cookies, etc).

DNS Verification via dig
=======

```
$ dig +short TXT _psl.dweb.link
"https://github.com/publicsuffix/list/pull/766"
```

make test
=========

I have successfully run `make test`: all PASS.

PR #2105, opened 2024-08-14
update contact for dweb.link and libp2p.direct

lidel:
This PR is updating contact information for `dweb.link` and `libp2p.direct` entries.
Interplanetary Shipyard took maintenance of public good utilities for [IPFS](https://ipfs.tech/) and [Libp2p](https://libp2p.io/) ecosystems
(https://docs.ipfs.tech/concepts/public-utilities/)

Original PRs: https://github.com/publicsuffix/list/pull/766 & https://github.com/publicsuffix/list/pull/2084

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -13912,6 +13912,10 @@ iopsys.se
 // Submitted by Matthew Hardeman <mhardeman@ipifony.com>
 ipifony.net
 
+// IPFS Project : https://ipfs.tech/
+// Submitted by Interplanetary Shipyard <domains@ipshipyard.com>
+*.dweb.link
+
 // ir.md : https://nic.ir.md
 // Submitted by Ali Soizi <info@nic.ir.md>
 ir.md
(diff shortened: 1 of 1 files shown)


simon-friedberger:
Please put the PR template back, it makes sense even for changes like this. Also, please update the _psl DNS entries to this PR.

simon-friedberger:
@lidel any updates?

lidel:
@simon-friedberger apologies, missed notification. restored the template, added TXT records (kept old ones, can remove once this PR is merged), and run tests. lmk if anything else is needed from our end.

wdhdev:
@lidel Please remove the old _psl TXT records, we only require the latest one.

simon-friedberger:
If there are multiple TXT records that is totally fine with me. Thanks for adding them!

simon-friedberger:
@lidel If these are handled by the same organization would it make sense to combine them into a single section?

wdhdev:
> If there are multiple TXT records that is totally fine with me. Thanks for adding them!

Oh, I thought we were only meant to have one.

simon-friedberger:
@wdhdev Maybe I am missing something? What would be the downside of having multiple?

wdhdev:
Not really any downside, but I was just thinking it would be more simple to just have one so we can easily find the latest PR for that domain. Just a little nitpick I guess.

lidel:
@simon-friedberger I believe it is better to keep them separate, because these are two separate open source projects. Currently, infrastructure is handled by the same entity, thus the same contact email, but this is only an operational detail.

PR #2094, opened 2024-08-09
merge `wdh.app` entries together using wildcard

wdhdev:
I completely forgot about wildcard support in the PSL 😆, so I've consolidated the 2 entries together. This is because I'm likely going to start deploying some more 3rd/4th level subdomains for my hosting service, not just `preview.wdh.app`.

I've also updated my website URL.

I've updated the `_psl.wdh.app` TXT record to this PR as well.

PR #2108, opened 2024-08-20
chore: update contact info + revert wildcard change for `wdh.app`

wdhdev:
Original PRs #2067 and #2094

Reasons for changes:
- Changed from a redirect domain (wdh.gg) to the website itself (wdharrison.com)
- Reverted wildcard change made in #2094 due to some behaviour that was not intended

I promise this will be the last PR for this domain, sorry about this!

I will update the `_psl.wdh.app` TXT record once I get the PR number.

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -15667,9 +15667,9 @@ toolforge.org
 wmcloud.org
 wmflabs.org
 
-// William Harrison : https://wdh.gg
-// Submitted by William Harrison <domains@wdh.gg>
-*.wdh.app
+// William Harrison : https://wdharrison.com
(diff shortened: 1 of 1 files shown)


simon-friedberger:
Would you mind sharing what the issue with the wildcard is?

wdhdev:
It considers 3rd level subdomains (e.g. www.wdh.app) as TLDs instead of the 2nd level (wdh.app) therefore sometimes causing issues with some services considering 3rd level subdomains as TLDs and not letting them be added to their service while only allowing 4th level tlds being considered domains themselves. (tl:dr; I misconsidered how wildcards are handled on the PSL)

simon-friedberger:
It would be great if you could update the wiki docs and explain the issues you were having!

PR #1390, opened 2021-08-05
Added Lokalized domain

ntaheij:
* [x] Description of Organization
* [x] Reason for PSL Inclusion
* [x] DNS verification via dig
* [x] Run Syntax Checker (make test)

* [x] Each domain listed in the PRIVATE section has and shall maintain at least two years remaining on registration, and we shall keep the _PSL txt record in place

__Submitter affirms the following:__ 
  * [x] We are listing any third party limits that we seek to work around in our rationale such as those between IOS 14.5+ and Facebook (see [Issue #1245](https://github.com/publicsuffix/list/issues/1245) as a well-documented example)
  * [x] This request was _not_ submitted with the objective of working around other third party limits
  * [x] The [Guidelines](https://github.com/publicsuffix/list/wiki/Guidelines) were carefully _read_ and _understood_, and this request conforms
  * [x] The submission follows the [guidelines](https://github.com/publicsuffix/list/wiki/Format) on formatting

---

Description of Organization
====

Organization Website: https://lokalized.nl

Providing hosting for websites and servers. We are working on a platform that can host websites and provide a secure environment for testing and deployment. We also strive to make an option to proxy to your own server (like ngrok.io). Use cases are: proxying, hosting your static content and deploy scalable apps with for example multiple nodes.

Reason for PSL Inclusion
====

Needed for Cookie Security due to the different websites (with "subdomains") on the hosting. Due to having multiple websites hosted by different people, we want to make it as secure as possible, therefore needing the PSL which will make the sites less vulnerable for same-site attacks.

DNS Verification via dig
=======

```
dig +short TXT _psl.servers.run
"https://github.com/publicsuffix/list/pull/1390"
```

make test
=========
Done.

PR #2116, opened 2024-08-25
remove `servers.run`

wdhdev:
Original PR #1390

Reasons for removal:
- Organisation domain (lokalized.nl) does not exist any more.
- When searching for subdomains on https://subdomainfinder.c99.nl/ there are no results.
- http://servers.run does not load, even though there is an A record defined
- When searching `site:servers.run` on Google, no results are returned.

The `_psl` TXT record does still exist however there does not seem to be any evidence of usage for what the original submitter said in the original PR:

> Needed for Cookie Security due to the different websites (with "subdomains") on the hosting. Due to having multiple websites hosted by different people, we want to make it as secure as possible, therefore needing the PSL which will make the sites less vulnerable for same-site attacks.

Specifically, they said there would be different websites using subdomains however there is no evidence of any subdomains existing. I believe the hosting service shut down, however they kept renewing this domain.

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -14152,10 +14152,6 @@ loginline.site
 // Submitted by Heiki Lõhmus <hostmaster at lohmus dot me>
 lohmus.me
 
-// Lokalized : https://lokalized.nl
-// Submitted by Noah Taheij <noah@lokalized.nl>
-servers.run
-
 // LubMAN UMCS Sp. z o.o : https://lubman.pl/
 // Submitted by Ireneusz Maliszewski <ireneusz.maliszewski@lubman.pl>
 krasnik.pl


ntaheij:
Hello, thank you for your inquiry! Yes, the domain is still active and used. Currently it is used for dynamic tunneling for some clients that need the separation. That is also the reason you do not see any active subdomains, as it is currently still only a few clients that use it occasionally.
Our company domain has been inactive indeed, as we switched to a new one.

Please let me know if you need any information.

wdhdev:
@ntaheij Thanks for the reply, I will close this PR now. However, can you please update your contact details in a new PR, thanks!

ntaheij:
Hi, yes, ofcourse. I will do so next week :)

groundcat:
Hi @ntaheij the domain `servers.run` has expired for over 1 month! 
Expiration Date: 2025-08-05 10:46:04. 
Do you still need this to be in the PSL?

ntaheij:
@groundcat sorry, this can be removed :)

PR #2124, opened 2024-09-02
remove `dapps.earth` section

wdhdev:
Original PR was #708 

Reasons for removal:
- Domain was added to the PSL on 2018-12-28, however the domain's registration date according to a WHOIS lookup is 2023-08-06, which likely means the domain lapsed and a different person registered it.
- Domain has expired as of 2024-08-06.
- No TXT record at `_psl.dapps.earth`
- No results when looking up `site:dapps.earth` on Google, other than a for sale page on the root which no longer loads.
- No active SSL certificates: https://crt.sh/?q=dapps.earth

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -12560,11 +12560,6 @@ localhost.daplie.me
 // Submitted by Abel Boldu / DAppNode Team <community@dappnode.io>
 dyndns.dappnode.io
 
-// dapps.earth : https://dapps.earth/
-// Submitted by Daniil Burdakov <icqkill@gmail.com>
-*.dapps.earth
-*.bzz.dapps.earth
-
 // Dark, Inc. : https://darklang.com
 // Submitted by Paul Biggar <ops@darklang.com>
 builtwithdark.com


simon-friedberger:
@wdhdev When creating removal PRs can you please also try to contact the email for that entry?

wdhdev:
Yes, I can. I will email the address listed on this entry and let you know when I receive a response.

groundcat:
Hello @burdakovd Is the dapps.earth entry still needed in the PSL? Our volunteer has identified and flagged the project associated with your domain as potentially inactive. Your domain may be removed from the PSL if we don't hear back from you. If the project has been decommissioned and the domain is no longer tied to an active project as described in your original request, please confirm here that it can be safely removed. This will help avoid unnecessary clutter for PSL volunteers and future registrants. Thank you.

groundcat:
I just sent an email to the submitter at `<icqkill@gmail.com>` to invite them to confirm here if possible.

wdhdev:
@simon-friedberger I have not received a response yet. However the domain is expired, so it is probably safe to remove.

groundcat:
> @simon-friedberger I have not received a response yet. However the domain is expired, so it is probably safe to remove.

Agree. This is highly likely a "set and forget" entry, and the domain was left to expire and then re-registered. I haven’t heard back either.

PR #2126, opened 2024-09-03
Fix syntax inconsistency

awsjulian:
I know this is like an OCD thing and pretty insignificant, but I just wanted to tidy up some inconsistencies in the comments. So, here's a quick PR to fix those.

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -12100,7 +12100,7 @@ eero-stage.online
 // Submitted by Apigee Security Team <security@apigee.com>
 apigee.io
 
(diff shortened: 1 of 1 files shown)


danderson:
The parser and formatter deliberately did not alter the style of existing blocks, to avoid excessive diff churn. It only reformats blocks if some other edit forces it to rewrite the comment.

That said, the format in this PR is what psltool uses as its canonical format, when it does rewrite metadata. So psltool should be able to parse this change, and will not revert it if those blocks need to be rewritten later. :+1:

PR #2146, opened 2024-09-06
remove `q-a.eu.org`

wdhdev:
Similar PR to #2099.

Reasons for removal:
- Not listed on nic.eu.org's open domains page: https://nic.eu.org/opendomains.html
- No sites found when searching `site:q-a.eu.org`
- No SSL certificates have been issued for this subdomain: https://crt.sh/?q=q-a.eu.org
- No subdomains were found when searching on https://subdomainfinder.c99.nl/

It should be safe to remove this domain as there does not seem to be evidence of usage.

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -13100,7 +13100,6 @@ nz.eu.org
 paris.eu.org
 pl.eu.org
 pt.eu.org
-q-a.eu.org
 ro.eu.org
 ru.eu.org
 se.eu.org


wdhdev:
It is worth noticing that paris.eu.org should possibly be removed as well, for the same reasons. However I did find one subdomain under paris.eu.org (which was cdn.paris.eu.org, which has NS records pointing to Cloudflare) which is why I did not include it in this PR, however I couldn't find any A, AAAA records or a live site on it. The only evidence of usage were some active SSL certificates issued for `cdn.paris.eu.org, *.cdn.paris.eu.org` (https://crt.sh/?q=paris.eu.org) however it is likely those are just automatically generated by Cloudflare. Let me know if that would be eligible for removal as well.

wdhdev:
@simon-friedberger This PR should be good for approval, as there does not seem to be any use on it.

wdhdev:
@simon-friedberger This PR and #2147 should be able to be merged. There is no evidence of usage for either of these domains and the hostmaster is not responsive as noted by @groundcat in #2099.

PR #2159, opened 2024-09-17
Enable DNS validation in psltool

simon-friedberger:
@danderson If you have a few minutes maybe you can answer some questions for me. :)
1. I disabled `wholeSuffixBlocks bool` because I need to distinguish between suffixes which have changed in this PR, so I can compare the PR ID and suffixes which belong to the same entity so I can check if the old _psl DNS record is still present.
2. I am now nesting a loop for the private section, with a loop for the entity, with a loop for the suffixes. That seems a bit weird since there is an interface `for _, suffix := range BlocksOfType[*Suffix](suffixes) {` which seems like I should be able to just iterate all the suffixes, but there doesn't seem to be any filtering. Could you clarify what the intent was here?

### DIFF
--- a/tools/internal/parser/errors.go
+++ b/tools/internal/parser/errors.go
@@ -2,6 +2,8 @@ package parser
 
 import (
 	"fmt"
+
+	"github.com/publicsuffix/list/tools/internal/domain"
 )
 
 // ErrInvalidEncoding reports that the input is encoded with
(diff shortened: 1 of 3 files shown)


simon-friedberger:
>    However, I think you can skip the middle layer: `BlocksOfType` walks the tree recursively starting at the node you select, so you can give it the private section and tell it you want all `Suffix` nodes, and it will give you all suffixes in the private section. You still need to filter on !changed, but the rule is that if a node is changed() then all its parents are changed() also. 

Right, that was my hack around the fact that I disabled `wholeSuffixBlocks`. I check if the middle layer changed and then make sure there is an _psl record on all the children, and if the child has changed itself I check that the PR ID actually matches.

PR #2160, opened 2024-09-18
tools/internal/parser: check validity of TXT records

danderson:
The validator looks up TXT records for non-exempt domains, then examines the PR they link to and verifies that the PR does in fact change the suffix that the TXT record is for. IOW, if you have a suffix `foo.com` and `_psl.foo.com` points to PR 1234, then PR 1234 must change `foo.com` in some way.

The exemption list is empirical: I just took all the domains that fail TXT validation because they have no TXT record, as well as Amazon's suffixes since they are still managed specially. There are a few at the bottom of the list that should get rechecked in a few weeks because it looks like they have various temporary failures... Or maybe they're defunct domains that should be deleted :shrug: 

Note that `psltool validate --online-checks` revalidates _all_ suffixes, which takes several minutes. It currently also doesn't quite work, because it's a little tricky to get the correct information out of github for submitted PRs and I need to do a bit more digging to find the right logic. So for now, assume that doesn't quite work yet and stick to `psltool check-pr`. I should be able to get it working in a followup PR tomorrow, I just ran out of time today and it's not essential for checking currently open PRs.

`psltool check-pr` only takes a few seconds, since it only checks the suffixes that have been changed by the PR, and in general requires only looking at the diff of that one PR.


danderson:
Of course, shortly after sending the PR, I figured out how to do the github PR stuff for merged PRs. I've added that as a 3rd commit to this PR.

Revalidating the entire PSL still doesn't work quite perfectly, so that is still a WIP:
 - It takes a _long_ time, I had to increase the timeout to 20min and that might still not be enough. We can remedy this with some caching, since state for merged PRs will never change. I'll hack on that next.
 - Some TXT records link to non-existent PRs, which the API client doesn't handle correctly atm so the error is confusing and talks about rebases when it should be reporting an invalid TXT record.
 - Some TXT records report false positive errors. For example, `repl.co` has a TXT record pointing to https://github.com/publicsuffix/list/pull/1568/files . This PR added new suffixes to the block that contains repl.co, and I guess for some reason the owners updated all their _psl records to point at that PR instead of keeping each record pointed at a PR relevant to that suffix. I haven't completed a full PSL check yet, but I would guess there's 10-30 suffixes in this situation.
   - I don't know if we want to be strict about this, but we can easily work around this using the `wholeSuffixBlock` diff mode. Then, the requirement is: your TXT record must point at any PR that modifies your suffix block, even if it doesn't modify the specific suffix we're currently checking. @simon-friedberger wdyt?
 - As I mentioned initially, there are some suffixes that need more investigation, at the end of the exception list. They reported temporary failures or timed out when I checked them, so I marked them as exempted. But if the failures _are_ temporary, and they have valid TXT records, we should un-exempt them.

None of these issues should affect `psltool check-pr` when you use it on open or recently merged PRs, that check should be quick & robust. It's just checking a decade+ of PSL data that still needs  a bit of work :)

simon-friedberger:
> Then, the requirement is: your TXT record must point at any PR that modifies your suffix block, even if it doesn't modify the specific suffix we're currently checking. @simon-friedberger wdyt?

It would be okay if we don't check that the record is pointing to a correct PR for each line for old entries. Obviously, when there is a PR, we want to validate that the person sending the PR is in control of the domain and not just any domain in that organization, because those sometimes change hands. So in that case we need the TXT record to mention the exact PR of the line that has changed.

simon-friedberger:
Slightly OT: any suggestions for reviewing these largeish PRs? The GH web interface is not great for jumping to e.g. functions in the code and it makes it kind of annoying.

simon-friedberger:
This looks like I could make a PR for two domains and only put the right PR ID in one of the records and it would work as long as the second domain has any _psl record. Am I reading this right?

danderson:
Oops, sorry, I missed your replies and just made the PR even larger :facepalm: I got the global recheck of the entire PSL working well, except for some legitimate validation failures that are for domains that should be removed (which is being worked on separately).

However, it's now a very large change, and as you say github is bad for code review in this case.

Here's what I can do instead: I already have this change partially split out into multiple small changes. I'll close this PR, and split the remaining large changes into more small ones. That way we can get some of the easy "noise" out of the way, and hopefully what's left is a smaller "main" change to review.

> This looks like I could make a PR for two domains and only put the right PR ID in one of the records and it would work as long as the second domain has any _psl record. Am I reading this right?

Hmm, I don't think so. When checking a PR, we look up the TXT record on all new/changed suffixes. Then we accumulate that into a map of "PR number X needs to reference suffixes A, B, C", and then we go and compare that to the PR diff. So, if you create one PR with 2 domains, and point one TXT record at a random different PR, the validation should correctly report that you have 1 suffix with an invalid TXT record, but the other one would be validated correctly.

danderson:
Closing this PR to reorganize the change into smaller pieces, because I got carried away with the size of the change :)

PR #1076, opened 2020-07-28
Modify cloud.metacentrum.cz entry

sustr4:
* [X] Description of Organization
* [X] Reason for PSL Inclusion
* [X] DNS verification via dig
* [X] Run Syntax Checker (make test)

Description of Organization
====

Organization Website: https://www.cesnet.cz/?lang=en
Activity Website: https://www.metacentrum.cz/en/index.html

CESNET provides numerous services for academic computing. This particular activity delivers a public IaaS cloud service for academic use, provided partly through a national portal (https://www.metacentrum.cz/en/Sluzby/Cloud/) and partly through the European Grid Initiative's Federated Cloud Platform (https://www.egi.eu/federation/egi-federated-cloud/).

Reason for PSL Inclusion
====

Our subdomains for public cloud machines are already included in the list. This pull request is a modification to exclude the actual head node of the service at `cloud.metacentrum.cz` from the public suffix list since that particular machine is not a public cloud VM but has been viewed as such by certain service providers due to its match with the list.

DNS Verification via dig
=======

```
$ dig +short TXT _psl.cloud.metacentrum.cz
"https://github.com/publicsuffix/list/pull/1076"
```

make test
=========

A hereby confirm I have run the test and got all passes.

PR #438, opened 2017-04-18
metacentrum.cz cloud subdomains

sustr4:
Subdomains used in NGI HPC Cloud services. Arranging for appropriate TXT records now.

PR #2168, opened 2024-09-18
combine duplicate sections

wdhdev:
Not sure why these were ever separate sections in the first place considering they are for the same organisation.

Original PRs: #438, #622 and #1076

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -15466,14 +15466,11 @@ webspace.rocks
 lima.zone
 
 // TransIP : https://www.transip.nl
-// Submitted by Rory Breuk <rbreuk@transip.nl>
+// Submitted by Rory Breuk <rbreuk@transip.nl> and Cedric Dubois <cedric.dubois@team.blue>
 *.transurl.be
 *.transurl.eu
-*.transurl.nl
-
-// TransIP: https://www.transip.nl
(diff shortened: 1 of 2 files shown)


wdhdev:
> If you feel like it, could you also send a PR to remove the hack from psltool to work around this? tools/internal/parser/diff.go:308
> 
> ... which mentions that TransIP is another duplicate block I'd forgotten about, if you're looking for other cleanups 😂

Thanks for letting me know, I've added that to this PR.

I've also updated `diff.go` in 610d363bbe3bee942325023e36b694b1bf58d055, please let me know if I have done it correctly as I am not exactly familiar with Go 😅

danderson:
Thanks! The diff.go change looks right to me. However, I believe it's PSL policy to keep public_suffix_list.dat changes separate from code changes, so this would need to be two PRs, one for the PSL edit and one to change psltool.

They can safely be submitted in that order (PSL first, then psltool), the diff hack just made the diff logic a bit more sensitive than necessary to changes for those suffix block, so it can get removed after the edit is merged.

wdhdev:
Discarded `diff.go` changes in this PR & opened #2180 for it instead. @simon-friedberger This is good to merge, please merge this PR *then* #2180 in order to prevent the CI from breaking.

PR #2175, opened 2024-09-18
tools/internal: wrap use of collators in mutexes

danderson:
Despite being a process of reading from read-only state, somehow these APIs turn out to be thread-unsafe, and parsing multiple PSLs concurrently in the same process results in data races and panics.

Thankfully each use of the collator is quite short, so wrapping their use in a global mutex, while unfortunate, is acceptable. And enabling parallel parsing of PSLs speeds up online validations by a factor of 10x on many cored machines.

---

First step of TXT record validation. To make it go fast it parses PSL files concurrently, and that revealed that x/text/collate is not thread safe :facepalm: This is a targeted fix for that issue.

### DIFF
--- a/tools/internal/domain/domain.go
+++ b/tools/internal/domain/domain.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"slices"
 	"strings"
+	"sync"
 
 	"golang.org/x/net/idna"
 	"golang.org/x/text/collate"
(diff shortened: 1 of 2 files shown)


simon-friedberger:
Seems good. Had I known it would cause this kind of problem I would have suggested we use whatever trivial byte sorting we can get for free instead of unicode collation.

Issue #2158, opened 2024-09-16
Make psltool accept two revisions or files


simon-friedberger:
`psltool pr-check` currently really requires a PR number but internally it only diffs two files. It would be useful if this could be run locally without requiring a PR.

PR #2177, opened 2024-09-18
tools/psltool: allow checking the PSL for an arbitrary commit on github

danderson:
Initial step for #2158: no multi-version comparison yet, but this makes `psltool validate` accept a local file or a commit hash. In the next PR I'll add optional comparison against a diff base.


simon-friedberger:
I think this would be better with a marker to distinguish if something is a path or a hash instead of guessing.

PR #2180, opened 2024-09-19
tools/internal/parser: remove exception for duplicate sections

wdhdev:
Opened in relation to #2168 as it seems we are meant to keep CI changes separate, which was mentioned by @danderson in https://github.com/publicsuffix/list/pull/2168#issuecomment-2359018880.

*Only merge this PR once #2168 is merged, otherwise checks will likely fail.*


danderson (commented):
:+1: change looks good

PR #1453, opened 2021-10-08
Add ts.net and beta.tailscale.net domains

danderson:
* [x] Description of Organization
* [x] Reason for PSL Inclusion
* [ ] DNS verification via dig **(waiting on PR number to set records)**
* [x] Run Syntax Checker (make test)

__Submitter affirms the following:__ 
  * [x] We are listing *any* third-party limits that we seek to work around in our rationale such as those between IOS 14.5+ and Facebook (see [Issue #1245](https://github.com/publicsuffix/list/issues/1245) as a well-documented example)
  * [x] This request was _not_ submitted with the objective of working around other third-party limits
  * [x] The [Guidelines](https://github.com/publicsuffix/list/wiki/Guidelines) were carefully _read_ and _understood_, and this request conforms
  * [x] The submission follows the [guidelines](https://github.com/publicsuffix/list/wiki/Format) on formatting
  * [x] Each domain listed in the PRIVATE section has and shall maintain at least two years remaining on registration, and we shall keep the _PSL txt record in place

  * (note) The affirmation regarding entries that match an organization website's primary domain is N/A here, as Tailscale's primary domain is tailscale.com and isn't being submitted for inclusion. Nonetheless, I understand and accept the documented downstream propagation and rollback properties of the PSL.

Description of Organization
====

Organization Website: https://www.tailscale.com

Tailscale is software that creates a private overlay mesh network between users' devices, secured by the WireGuard VPN protocol.

Among other features, Tailscale lets users provision TLS certificates for the devices on their networks, via Let's Encrypt. The resulting certs are provisioned on shared domains operated by Tailscale, and follow the form `<device name>.<network name>.beta.tailscale.net` and `<device name>.<network name>.ts.net`.

TLS certificates may seem superfluous given that there's already an L3 VPN securing and authenticating communications, but browsers disagree: increasingly non-TLS connections result in scary warnings to users, and some newer features of the web platform are hard-disabled in "insecure contexts" - which browsers take to mean "no TLS". To avoid a downgraded web experience, Tailscale facilitates the issuance of TLS certificates by plumbing the ACME dns-01 challenge through the ts.net and beta.tailscale.net domains on behalf of user devices.

In future, we also plan to use these TLS certificates to provide "ingress" functionality, allowing controlled exposure of private devices to the internet while preserving e2e privacy of connections (since only the user device has the TLS private key, Tailscale merely facilitates issuance and L4 packet forwarding).

[...]

PR #2216, opened 2024-10-15
Remove beta.tailscale.net

danderson:
(for once, wearing my day job hat not my contributing code hat)

Following up on commitment in #1453. I originally submitted `beta.tailscale.net` for inclusion, with a note that despite the name, it was being used for production hosting and that we were working on migrating off it and onto the other suffixes Tailscale has listed. That migration is now done, so `beta.tailscale.net` can be removed.


wdhdev:
Are you able to update the _psl TXT record?

simon-friedberger:
This would be a good opportunity to update the email address to something that stays reachable in case leave tailscale.

PR #2218, opened 2024-10-15
Improve psltool PR check

simon-friedberger:
Enable DNS checks for PR action.

Rename Client to GHRepo because the internal github.Client and the github.Client from the library confuses my IDE.

Pass the GHRepo to validateTXTRecords so it uses the correct repository.

### DIFF
--- a/tools/internal/github/pr.go
+++ b/tools/internal/github/pr.go
@@ -15,7 +15,7 @@ import (
 // Client is a GitHub API client that performs PSL-specific
 // operations. The zero value is a client that interacts with the
 // official publicsuffix/list repository.
-type Client struct {
+type Repo struct {
 	// Owner is the github account of the repository to query. If
 	// empty, defaults to "publicsuffix".
 	Owner string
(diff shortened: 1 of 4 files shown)


tools/psltool/psltool.go
```diff
@@ -154,7 +156,7 @@ func runValidate(env *command.Env, pathOrHash string) error {
 		bs, err = os.ReadFile(pathOrHash)
 	} else if isHex(pathOrHash) {
 		// input looks like a git hash
-		client := github.Client{}
+		client := github.GHRepo{}
```
danderson, line 159: I think this client needs the owner/repo fields as well, otherwise PSLForHash will look up in the wrong repo

tools/internal/github/pr.go
```diff
@@ -15,7 +15,7 @@ import (
 // Client is a GitHub API client that performs PSL-specific
 // operations. The zero value is a client that interacts with the
 // official publicsuffix/list repository.
-type Client struct {
+type GHRepo struct {
```
danderson, line 18: Tiny coding style suggestion: could we make this just `Repo`, or does that still cause confusion with the github library?

In general in Go, we try to name the types so that they are nice to read when combined with the package name, because every usage of the type requires the package. `github.GHRepo` repeats itself slightly, compared to `github.Repo`.

This is just an extremely minor suggestion, to be clear. GHRepo is also fine if you prefer that :)

danderson (commented):
Looks good! I don't have the github permissions to hit the approve button, but :+1:

PR #2220, opened 2024-10-15
Add home.arpa

### LINKED ISSUE #1205: Special use domain 'home.arpa' missing
> 'home.arpa.' is designated for non-unique use in residential home networks. 

See [RFC 8375](https://tools.ietf.org/html/rfc8375).

simon-friedberger:
This is not actually a *public* suffix but it is a domain suffix.

This closes #1205.

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -206,6 +206,7 @@ tur.ar
 // Confirmed by registry <iana-questions@icann.org> 2008-06-18
 arpa
 e164.arpa
+home.arpa
 in-addr.arpa
 ip6.arpa
 iris.arpa


dnsguru:
IANA -> https://www.iana.org/domains/arpa -> https://www.rfc-editor.org/rfc/rfc8375.html

It exists

public_suffix_list.dat
```diff
@@ -206,6 +206,7 @@ tur.ar
 // Confirmed by registry <iana-questions@icann.org> 2008-06-18
 arpa
 e164.arpa
+home.arpa
```
dnsguru, line 209: This may introduce some considerations with downstream (non-browser) use cases.

Example: the home.arpa use case described in [RFC 8375](https://www.rfc-editor.org/rfc/rfc8375.html) describes a concept of `homenet` and these are typically going to be defined within that limited realm.  This means that there will be coexistence on other `homenet` configurations of subdomains with the same convention (ex: gateway.home.arpa, video.home.arpa etc.), and it seems like considerations where a device that moves from one `homenet` to another `homenet` could leak cookies, would wildcard/basic issuance of certs happen, and who wins the race to be the first to configure gateway.home.arpa at cloudflare.

simon-friedberger: Hm...so I agree with the general concerns but I'm not sure if the entry being on the PSL makes a large difference here. There may be a bit of a difference when the TLD is not on the list like `.internal`, because then maybe it is blocked in certain places but since .arpa is already on the list I expect something.home.arpa to already work everywhere.

(To be clear, I would also support adding .internal because currently if you type something.internal into Chrome it will search for it instead of navigating to it.)

wdhdev: > who wins the race to be the first to configure gateway.home.arpa at cloudflare.

I could be wrong, however I highly doubt it would be possible for someone to actually configure that at Cloudflare, especially as you have to set the nameservers specifically before it is activated. Also, I would imagine Cloudflare has something in place to stop this happening.

simon-friedberger: ISTM if we add `home.arpa` the problem becomes smaller. Because currently `anything.home.arpa` gets session cookies for `gateway.home.arpa`.

dnsguru: I know that through a browser-only use case this seems a simple matter, but it could have cascading impacts on other PSL dependant systems or services to add this (and .internal) to the PSL.

home.arpa likely has little implementation and thus a low amount of impact to this being merged, but .internal (out of scope to this PR) may be widely used and need some consideration.

I only raise these issues up front before merging something because it is truly the up-front friction that is the main area to affect these changes. If there is an addition that gets merged that ends up being disruptive or introducing instabilities, even if we quickly remove the change, the cascade time is not immediate, so something bad introduced could sit for an indeterminate amount of time.

simon-friedberger: Do you happen to have an example of something where this might be a problem? (Feel free to paraphrase the use-case if it's something you might not be allowed to speak about.)

simon-friedberger: @dnsguru Just to be clear, I agree with you. There are definitely issues here and we shouldn't add `.internal` for now but since we already have `.arpa` we should add `.home.arpa` to at least make it clear that these are separate entities.

PR #2228, opened 2024-10-18
internal/parser: add PublicSuffix and RegisteredDomain methods to List

danderson:
These calculate the public suffix or registered/registerable domain of some input domain, according to the strict PSL algorithm. Aside from being useful as a reference implementation of the algorithm, it comes in handy in validations that need to do stuff like figure out what registry to query for domain info.

---

Please check my work carefully on the PSL algorithm, in particular the handling of wildcard domains. I read [the spec](https://github.com/publicsuffix/list/wiki/Format#formal-algorithm) very carefully and also the [platform.sh wiki page](https://wiki.mozilla.org/Public_Suffix_List/platform.sh_Problem), and I _think_ I've correctly implemented the strict PSL algorithm with no deviation (a public suffix `*.foo.com` does not imply a public suffix `foo.com`, and there is an implicit hardcoded `*` public suffix so that all unknown TLDs are automatically public suffixes).

Later on I could add modifiers to these methods to implement the variants as well, but for now I stuck to the reference algorithm. I ended up needing this logic to implement RDAP support for checking domain expiry, since I have to map a random PSL entry to "which registry RDAP service should I query", and for some suffixes that requires computing the "next" public suffix of the PSL entry :)

### DIFF
--- a/tools/internal/domain/domain.go
+++ b/tools/internal/domain/domain.go
@@ -156,16 +156,31 @@ func (d Name) CutSuffix(suffix Name) (rest []Label, found bool) {
 	return ret, true
 }
 
-// AddPrefix returns d prefixed with label.
+// AddPrefix returns d prefixed with labels.
 //
-// For example, AddPrefix of "bar" to "foo.com" is "bar.foo.com".
-func (d Name) AddPrefix(label Label) (Name, error) {
+// For example, AddPrefix("qux", "bar") to "foo.com" is "qux.bar.foo.com".
+func (d Name) AddPrefix(labels ...Label) (Name, error) {
 	// Due to total name length restrictions, we have to fully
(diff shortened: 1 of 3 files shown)


wdhdev:
To confirm, a `*.foo.com` entry would indicate all subdomains at the third level (e.g. `bar.foo.com`) are declared as TLDs.

danderson:
TLD is may be a slight misnomer, in that it might imply the involvement of IANA-level registries. But with an adjusted term, yes: `*.foo.com` means that all 3-label domains ending `foo.com` are public suffixes in _some_ registry (public or private), and all 4-label domains ending `foo.com` are registered/registrable domains.

Said in terms of cookie security, a browser will allow setting cookies on `fourth.bar.foo.com` (registered domain), but not on `bar.foo.com` (public suffix).

PR #2236, opened 2024-10-25
chore: update `.io` section

wdhdev:
Updating the list of SLDs under the `.io` TLD.

I'm adding `co.io` and `net.io` as they do seem to be a domain offered by the registry, as it is not reserved, and is registered to the registry operator themselves (`Registry Operator acts as Registrar (9998)`), along with having the same nameservers of `com.io`, which we know is a SLD already. However I will confirm this with the registry, to make sure.

Note: You can technically register domains at the 3rd level, however they are restricted to residents in the British Indian Ocean Territory, however I don't believe there are any inhabitants living there. However, they are still TLDs technically offered by the registry.

This PR will remain in a draft state until I receive a response from the .io registry.

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -1428,9 +1428,9 @@ int
 eu.int
 
 // io : http://www.nic.io/rules.htm
-// list of other 2nd level tlds ?
 io
 com.io
+net.io
 
 // iq : http://www.cmc.iq/english/iq/iqregister1.htm
 iq
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -1429,6 +1429,7 @@ eu.int
 
 // io : http://www.nic.io/rules.htm
 io
+co.io
 com.io
 net.io
 
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -1431,7 +1431,12 @@ eu.int
 io
 co.io
 com.io
+edu.io
+gov.io
+mil.io
 net.io
+nom.io
+org.io
 
 // iq : http://www.cmc.iq/english/iq/iqregister1.htm
 iq


groundcat:
> I'm adding `co.io` and `net.io` as they do seem to be a domain offered by the registry, as it is not reserved, and is registered to the registry operator themselves (`Registry Operator acts as Registrar (9998)`), along with having the same nameservers of `com.io`, which we know is a SLD already. However I will confirm this with the registry, to make sure.
> 
> Note: You can technically register domains at the 3rd level, however they are restricted to residents in the British Indian Ocean Territory, however I don't believe there are any inhabitants living there. However, they are still TLDs technically offered by the registry.

May I know if there is an example showing where people can register third-level domains under these suffixes with a registrar or directly with the registry? I could not find any living example through search engines or certificate transparency of any use of `.co.io` or `.net.io` as public suffixes. Maybe I am missing something. 

When a domain is not reserved by the registry but is registered by the registry (with the registry's name as the registrant), it could be another form of reservation depending on how the registry operates, or it could be reflected in their WHOIS records, instead of being a clear indication that the domain is used as a public suffix. An arbitrary example I can think of is `net.us` (and some other similar ones under .us such as `edu.us`), which, according to WHOIS, is registered by the registry itself instead of showing as reserved, but `.net.us` is not a public suffix.

wdhdev:
> May I know if there is an example showing where people can register third-level domains under these suffixes with a registrar or directly with the registry? I could not find any living example through search engines or certificate transparency of any use of .co.io or .net.io as public suffixes. Maybe I am missing something.

They aren't in use as no one is a resident of the islands, which is why they have no public usage, however they do still seem to be existing TLDs/public suffixes, however I could be wrong, which is why I am contacting the registry to confirm.

From https://en.wikipedia.org/wiki/.io:

> Third-level domains, such as "xyz.com.io", can only be registered by an inhabitant of the area. (Since there are no legal, permanent inhabitants of the British Indian Ocean Territory, theoretically no third-level domains will be registered.)

It is also stated on their rules page: https://nic.io/rules.htm

> When a domain is not reserved by the registry but is registered by the registry (with the registry's name as the registrant), it could be another form of reservation depending on how the registry operates, or it could be reflected in their WHOIS records, instead of being a clear indication that the domain is used as a public suffix. An arbitrary example I can think of is net.us (and some other similar ones under .us such as edu.us), which, according to WHOIS, is registered by the registry itself instead of showing as reserved, but .net.us is not a public suffix.

I'm aware, that is why I am in the middle of contacting the registry to enquire about that, hence this PR is draft as well.

> On a side note, on October 3, 2024, the UK announced that it is giving up sovereignty over the Chagos Islands, transferring them to Mauritius. The ".io" domain is tied to these islands, and there is a possibility that the .io domain might eventually be phased out after a number of years, but the future remains unclear.

Yes, that's true, however at this point I believe ICANN hasn't said anything regarding it, so it is safe for now.

wdhdev:
I have received a response from Identify Digital; the registry operator; with the following list of SLDs:
```
nom.io
net.io
org.io
gov.io
mil.io
edu.io
co.io
com.io
```

@simon-friedberger This PR is good to merge.

PR #2183, opened 2024-09-19
rename `William Harrison` to `Harrison Network`

wdhdev:
Renaming my section as `Harrison Network` is more fitting considering it is not for personal use per se.

PR #2246, opened 2024-11-08
move `wdh.app` to new section

wdhdev:
Simply moving the domain `wdh.app` to a new section.

Reason for move is the domain is not really associated with Harrison Network.

The email address is a personal one, as the section is a personal section. 

Updated `_psl` TXT record:
```
$ dig +short TXT _psl.wdh.app
"https://github.com/publicsuffix/list/pull/2246"
```

All tests passed.

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -13676,7 +13676,6 @@ häkkinen.fi
 
 // Harrison Network : https://hrsn.net
 // Submitted by William Harrison <psl@hrsn.net>
-wdh.app
 hrsn.dev
 
 // Hashbang : https://hashbang.sh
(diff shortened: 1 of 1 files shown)


simon-friedberger:
I'm not sure if we should do this. Supporting personal sections sounds like it would prompt a lot of questions. And this doesn't have the PR template filled out. And what are we really gaining by splitting entries that are both effectively you?

wdhdev:
I should clarify, it's run by myself as an individual, however not under Harrison Network which would more clarify as a 'business'. This is basically just partially reverting #2183.

There seem to be at least one existing personal section (assuming it counts as a personal section):
The Lõhmus Family: https://github.com/publicsuffix/list/blob/main/public_suffix_list.dat#L14120

simon-friedberger:
I am sure there are more I am just not sure we should be reinforcing it. I'll leave it up to you.

wdhdev:
I'd personally like to keep them separate, as `wdh.app` is not under Harrison Network.

PR #2248, opened 2024-11-08
chore: add `.bw` 2LDs

wdhdev:
2LDs for `.bw` were found on Wikipedia: https://en.wikipedia.org/wiki/.bw#Structure

These 2LDs do seem to be in use when searching `site:{ac,agric,gov,me,net,shop}.bw` on Google.

These should be added ASAP as people are currently able to set cookies on the TLDs themselves, which would affect potentially thousands of other domains on these TLDs.

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -676,11 +676,16 @@ org.bt
 bv
 
 // bw : https://www.iana.org/domains/root/db/bw.html
-// http://www.gobin.info/domainname/bw.doc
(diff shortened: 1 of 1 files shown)


dnsguru:
please keep the upper section only updated by the information from a provable source at respective TLD operator(s)

Using Google site or wikipedia is not ok for this, as Google is not aware of the structure of the TLD space the official registry with the delegation from IANA operates, and wikipedia is also not authoritative.

PR #2249, opened 2024-11-08
add `co.bz`

wdhdev:
Taken from https://belizenic.bz, the URL linked in the section.

There does seem to be usage of the 2LD `.co.bz`, when searching `site:co.bz` on Google.

These should be added ASAP as people are currently able to set cookies on the 2LD itself, which would affect potentially hundreds of other domains on this 2LD.

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -698,6 +698,7 @@ of.by
 // bz : https://www.iana.org/domains/root/db/bz.html
 // http://www.belizenic.bz/
 bz
+co.bz
 com.bz
 net.bz
 org.bz


public_suffix_list.dat
```diff
@@ -698,6 +698,7 @@ of.by
 // bz : https://www.iana.org/domains/root/db/bz.html
 // http://www.belizenic.bz/
 bz
+co.bz
```
dnsguru, line 701: The selected line `+co.bz` adds a new public suffix entry for the `co.bz` domain under the Belize (`bz`) top-level domain. This pull request started with IANA and then the registrry as defined by the IANA db entry for .bz, which then has the name in the dropdown and their policy pages at the authoritative registry.

change indicates that `co.bz` is now recognized as a valid public suffix within the public suffix list.

dnsguru:
This pull request, wihich impacts the ICANN section, did use an authoritative source.

PR #2250, opened 2024-11-08
remove wildcard for `.ck`, replace with list of 2LDs

wdhdev:
List of 2LDs taken from https://en.wikipedia.org/wiki/.ck#Requirements_of_.ck_registration.

I would recommend removing the wildcard and specifying the 2LDs specifically, just in the event they allow 2nd level registrations, it won't affect anyone attempting to set cookies at the second level.

This PR isn't super urgent, and won't affect any active domains once deployed.

The `!www.ck` entry has been removed in favour of specifying the TLD itself.

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -763,8 +763,15 @@ int.ci
 gouv.ci
 
 // ck : https://www.iana.org/domains/root/db/ck.html
-*.ck
-!www.ck
+// https://en.wikipedia.org/wiki/.ck#Requirements_of_.ck_registration
+ck
+biz.ck
+co.ck
+edu.ck
+gen.ck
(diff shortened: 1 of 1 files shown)


wdhdev:
Tests seem to be failing due to https://github.com/publicsuffix/list/blob/main/tests/test_psl.txt#L62-L68

dnsguru:
please do not add namespace in the upper section based upon unauthoritative information in wikipedia.

simon-friedberger:
@wdhdev Also note that having the wildcard means more domains (potentiially too many) are public suffixes. Which means in terms of site-isolation this is the more secure state. It also requires fewer entries so we should stick with those if we have no need to change.

PR #2260, opened 2024-11-15
remove `browsersafetymark.io`

wdhdev:
Reasons for removal:
- Website does not resolve: http://browsersafetymark.io
- A record `35.161.17.166` does not respond to pings
- No subdomains found when searching using https://subdomainfinder.c99.nl
- No results when searching `site:browsersafetymark.io` on Google.
- No results on archive.org (not a factor for removal, but worth noting): https://web.archive.org/web/*/https://browsersafetymark.io/*

Reasons against removal:
- The TXT record at `_psl.browsersafetymark.io` still exists

I think this domain is safe to remove as it does not seem to be used. It is likely they have pretty much abandoned this domain.

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -12279,10 +12279,6 @@ square7.net
 shop.brendly.hr
 shop.brendly.rs
 
-// BrowserSafetyMark
-// Submitted by Dave Tharp <browsersafetymark.io@quicinc.com>
-browsersafetymark.io
-
 // BRS Media : https://brsmedia.com/
 // Submitted by Gavin Brown <gavin.brown@centralnic.com>
 radio.am


groundcat:
@dtharp Is the browsersafetymark.io project (#251) still active? If the service has been decommissioned and the domain is no longer linked to an active project, as outlined in your original request, please confirm whether it can be safely removed from the PSL. If your project or organization is still active and requires continued inclusion in the PSL, please also reply at your earliest convenience. 

If we do not hear back from you for an extended period (over 2 weeks), your entry may be subject to removal.

groundcat:
@wdhdev I agree with you that, at first glance, there is no clear evidence as to whether this domain is still actively used by the project. However, based on the initial requestor's [description](https://github.com/publicsuffix/list/pull/251#issuecomment-235634491) of the domain's purpose, it appears to be associated with a project related to EFF.org for testing ad and tracking functionality. This would explain why it does not appear in Google search results or subdomain finders. 

The domain still resolves to an AWS IP (the fact that it is not pingable simply means it does not accept ICMP packets; this does not necessarily indicate that any service using the domain is inactive). WHOIS data shows the domain is likely still with the original registrant and has not expired. According to the VirusTotal report, it has over 1,700 subdomains that continue to actively resolve to an IP address, indicating that DNS is still functioning. The main domain points to `35.161.17.166` and has done so consistently since at least 2019. Historical WHOIS lookups also show the same registrar.

The TXT record at `_psl.browsersafetymark.io` still exists, as you also observed, which indicates they still want this entry included in the PSL. This should be a strong sign that it should not be removed.

This page on EFF’s site still displays information about browsersafetymark.io: 

https://www.eff.org/https-everywhere/atlas/domains/browsersafetymark.io.html

> HTTPS Everywhere currently rewrites requests to browsersafetymark.io (or its subdomains).

So, I do not think it is safe to remove this domain from the list. 

Additionally, before creating new PRs for domain removal, I suggest it would be best to contact the initial requestor first for confirmation to avoid mistakenly removing domains.

wdhdev:
> This would explain why it does not appear in Google search results or subdomain finders.

That would definitely explain so.

> The domain still resolves to an AWS IP (the fact that it is not pingable simply means it does not accept ICMP packets; this does not necessarily indicate that any service using the domain is inactive).

I am aware that an IP not responding to pings does not mean it is active, and it is not a factor I took in when opening this PR, it was just simply a quick test I ran.

> According to the VirusTotal report, it has over 1,700 subdomains that continue to actively resolve to an IP address, indicating that DNS is still functioning.

I did not check VirusTotal, I will do that next time.

> The latest version of the "HTTPS Everywhere" extension still uses this domain, so I assume this domain remains useful for users who continue to use HTTPS Everywhere in their browsers. While major browsers now provide native support for an HTTPS-only mode and the extension is no longer maintained, it is still possible that users rely on it in their browsers.

That is true, I didn't originally know what the purpose of this domain was actually used for, now I do.

> So, I do not think it is safe to remove this domain from the list, unless the initial requestor explicitly confirms it is no longer in use or requests its removal.

I agree with this now that I have more context, I have closed this PR.

> Additionally, before creating new PRs for domain removal, I suggest it would be best to contact the initial requestor first for confirmation to avoid mistakenly removing domains.

Will do in future, thanks. I didn't think to contact the original submitter as they haven't been active on GitHub since the PR was submitted, so I assumed their account along with this domain was inactive.

PR #2264, opened 2024-11-17
Update .JO section

groundcat:
I am creating this pull request to update the .JO block in the ICANN section.

1. Started at the IANA page: https://www.iana.org/domains/root/db/jo.html and identified the registry website for registration services: https://www.dns.jo/
3. Received a direct reply from the domain registry (`Domain .jo/الاردن. <DNS@modee.gov.jo>`), confirming the requested information.

Excerpt from the email confirmation:

> 1. Operational Status: Is [name.jo](http://name.jo/) still active and registrable?
> ‘.[name.jo](http://name.jo/)’ was stopped since many years and not available for registration, it replaced with ‘.[per.jo](http://per.jo/)’.
> 2. Available Suffixes: Could you provide an updated list of all second-level domains currently operated by the .JO registry?
> see our website:  https://www.dns.jo/JoFamily.aspx

A copy of the email confirmation along with the email headers is available to be forwarded to a maintainer for review upon request.

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -1896,7 +1896,8 @@ org.je
 // jm : http://www.com.jm/register.html
 *.jm
 
-// jo : http://www.dns.jo/Registration_policy.aspx
+// jo : https://www.dns.jo/JoFamily.aspx
+// Confirmed by registry <DNS@modee.gov.jo> 2024-11-17
 jo
 agri.jo
 ai.jo
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -1901,16 +1901,16 @@ jo
 agri.jo
 ai.jo
 com.jo
+edu.jo
 eng.jo
 fm.jo
-org.jo
-net.jo
-edu.jo
-sch.jo
 gov.jo
 mil.jo
+net.jo
+org.jo
 per.jo
 phd.jo
+sch.jo
 tv.jo
 
 // jobs : https://www.iana.org/domains/root/db/jobs.html
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -1898,14 +1898,20 @@ org.je
 
 // jo : http://www.dns.jo/Registration_policy.aspx
 jo
+agri.jo
+ai.jo
 com.jo
+eng.jo
+fm.jo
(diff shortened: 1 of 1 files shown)


groundcat:
@mozfreddyb Sure! I just sent it over to you via email in `.eml` format.

mozfreddyb:
The email did not contain any DKIM signature... We will need to figure out how else to verify this. I think we should have a maintainers email group or some other solution?

groundcat:
> The email did not contain any DKIM signature... We will need to figure out how else to verify this. I think we should have a maintainers email group or some other solution?

Indeed... Unfortunately, this email did not include a DKIM signature, so there is no reliable way to confirm that it has not been tampered with.

However, for this specific TLD, the registry contact directed me to this URL: https://www.dns.jo/JoFamily.aspx. In the meantime, if we trace the path via the IANA page: https://www.iana.org/domains/root/db/jo.html and identified the registry website for registration services: https://www.dns.jo/. This site is indeed the authentic registry website.

For this specific case, by following the path (with the IANA database as the root of the evidence chain), the page at https://www.dns.jo/JoFamily.aspx can be considered an authoritative source, even in the absence of an email confirmation.

That said, I do understand your point regarding the importance of a validation chain for email-based evidence in other cases.

simon-friedberger:
I'm not sure I completely trust the translation of that site. @groundcat could you clarify if second level domains are registrable? It would be nice if we could just add `*.jo` instead.

groundcat:
> I'm not sure I completely trust the translation of that site.

The .JO registry website does have its own English version, which can be considered an official translation. Therefore, in this case, there is no need to rely on third-party translations, such as Google Translate, which can often provide questionable results.  

You can click the menu icon in the top right corner and select "English" to view the English version of the list of all suffixes. Since this page is provided directly by the registry, it should be considered authoritative as well.  

> could you clarify if second level domains are registrable? It would be nice if we could just add *.jo instead.

I do not think we can use `*.jo` because `.jo` is already an available suffix. A living example is Apple at https://apple.jo, KPMG at http://kpmg.jo/, and Google at https://www.google.jo/. Therefore, `.jo` must remain, along with separate SLD entries in the PSL.

PR #2274, opened 2024-11-19
Update `.MG` Section

groundcat:
I am creating this pull request to update the .MG block in the ICANN section.

1. Started at the IANA page: https://www.iana.org/domains/root/db/mg.html and identified the email address as well as the registry website for registration services https://www.nic.mg/
3. Received a direct reply from the domain registry `Dany RANDRIARINIVO <dany@nic.mg>`, confirming the requested information.

Fully reply to the email confirmation:

> On Tue, Nov 19, 2024 at 3:27 AM Dany RANDRIARINIVO <dany@nic.mg> wrote:
> Greetings,
> 
> The active and registrable second-level domains for the .MG are :
> org.mg
> nom.mg
> gov.mg
> prd.mg
> com.mg
> co.mg
> 
> edu.mg and mil.mg are reserved.
> 
> tm.mg is not active.
> Sincerely,

- SPF:	PASS with IP 46.105.41.16
- No DKIM

The email has been forwarded to @simon-friedberger

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -3979,17 +3979,14 @@ gov.me
 its.me
 priv.me
 
-// mg : http://nic.mg/nicmg/?page_id=39
+// mg : https://nic.mg
 mg
-org.mg
-nom.mg
-gov.mg
-prd.mg
-tm.mg
-edu.mg
-mil.mg
-com.mg
 co.mg
+com.mg
+gov.mg
+nom.mg
+org.mg
+prd.mg
 
 // mh : https://www.iana.org/domains/root/db/mh.html
 mh
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -3983,7 +3983,9 @@ priv.me
 mg
 co.mg
 com.mg
+edu.mg
 gov.mg
+mil.mg
 nom.mg
 org.mg
 prd.mg


simon-friedberger:
I am not sure about this strategy of removing things that are reserved. Does reserved mean "reserved for use by specific parties" or does it mean "reserved for future use and unavailable"? Should we check if there are any `.edu.mg` or `.mil.mg` sites?

wdhdev:
Personally, I don't think it's worth keeping SLDs that are reserved, but are not available for registration by anyone nor are in use.

An example of a SLD that shouldn't be included is `co.au`, it is reserved but it is not in use and has never been in use.

simon-friedberger:
I agree I just wanted to say, maybe they *are* in use. "Reserved" could just mean that it is not available to the public, like `.mil.` might be. It's reserved for certain entities but it is still in use.

groundcat:
An email inquiry regarding edu.mg and mil.mg has been sent to ask about their operational status.

groundcat:
Adding back edu.mg and mil.mg because these sites still appear to be working on edu.mg. Coincidentally, edu.mg shows an NXDOMAIN error when scanning the DNS, so it might just be a temporary downtime of some of the .mg domains at the moment.

groundcat:
Confirmation by the registry regarding `edu.mg` and `mil.mg`:

> edu.mg and mil.mg can be include in the public suffix List.
> edu.mg is actively use by university and research institute.
> mil.mg does not have active domain yet but the military can use it on the future.
> Thanks,

dnsguru:
"Reserved" means a variety of things.

Typically these are potential new stubs they may open later and they don't
want to deal with more pull requests later.

Also, alternatively...

Some countries want them as they may have local isp or organizations that
resolve these locally within their networks and they want the cookie
isolation or other benefits (dmarc)

Truly, the cctld to focus on should be those which cite wikipedia as source

Issue #270, opened 2016-07-23
.post has structure


pzb:
Table 2 in https://www.upu.int/fileadmin/documentsFiles/activities/dotPost/policyManagementDotPostEn.pdf lays out the heirarchy for .post, including com.post, edu.post, and org.post plus ISO 3166-1 alpha 2

Based on the .post zone file, there are currently names registered with three labels under ao, ci, es, id, kh, ma, mo, my, nl, ug, uy, za and org.

dnsguru:
Spoke with @pzb on this as a follow-up.  There is some structure within the .POST TLD according to the zone file, but we'd want the registry themselves to provide this directly, and it looks complex to parse deep into the namespace defined within the document.  

The UPU file link was no longer working, here is the [Archive.org version](https://web.archive.org/web/20181005165818/https://www.upu.int/fileadmin/documentsFiles/activities/dotPost/policyManagementDotPostEn.pdf)

Will close this for now but keep for archival purposes and flag this as a namespace with heavy structure and potentially many levels, akin to .US #274  and #276 , .NAME #277, .BR or .JP (certainly others exist) where the ICANN section might exponentially grow (#277 proposed adding nearly 20k lines to PSL).

We may look at some means to address the more heavily structured TLDs in the future, but for now, we'll close this ticket and reference it in the roadmap.

PR #2278, opened 2024-11-19
Update `.POST` Section

groundcat:
I am creating this pull request to update the .POST block in the ICANN section.

Starting today, any legal entity engaged in the postal, logistics, or supply chain sectors can register its name in .post, .com.post, .org.post and .edu.post.

**Steps taken to verify information from the authoritative source:**

1. Started at the IANA page: https://www.iana.org/domains/root/db/post.html and identified the registry website for registration services http://www.upu.int
2. From https://www.upu.int, located registry website https://trust.post
3. From https://trust.post located the policy page https://trust.post/legal/
4. Located "[Table 2 – Domain naming conventions and restrictions](https://cdn.sanity.io/files/ppne44jb/production/fd7947f2077c82acee13ed60d407360c8ee3dcd8.pdf)":

image

Therefore, adding:

```
com.post
edu.post
org.post
```

**Other supplemental information from the registrar:**

https://www.101domain.com/post.htm

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -5334,7 +5334,11 @@ edu.pn
 net.pn
 
 // post : https://www.iana.org/domains/root/db/post.html
+// https://trust.post/legal/
 post
+com.post
+edu.post
+org.post
 
 // pr : http://www.nic.pr/index.asp?f=1
 pr


simon-friedberger:
@groundcat More general discussion point: Should we actually add these? Is it worth checking if these domains are actually in use first?

After all, we are increasing the size of the list, but maybe these are just ideas the registrar had that nobody was ever interested in buying.

groundcat:
> After all, we are increasing the size of the list, but maybe these are just ideas the registrar had that nobody was ever interested in buying.

Good point. I personally doubt that many registrants would choose to purchase third-level `.post` domains, except for brand protection purposes, also given the [price](https://www.101domain.com/com_post.htm) of 129.99 USD per year. 

It might end up like the super long list of `.aero` SLD suffixes in the PSL, which almost no one uses after the registry introduced them, expecting there would be interest. Honestly, `.aero` has so many entries that many or most of them should probably be removed. 

A similar example in the ccTLD space is `.jp`, which had an overly long list of city- or regional-level second-, third-, and even fourth-level domains. In 2012, they eventually decided to decommission that system, but the suffixes remain in the PSL and currently many are returning `NXDOMAIN`.

> More general discussion point: Should we actually add these? Is it worth checking if these domains are actually in use first?

This brings us to the question: when registries in the ICANN/IANA section roll out new suffixes, should we evaluate them in the similar way as addition requests in the private section by assessing their relevance (e.g., user count)? If so, what should be the standard? Do suffixes in the ICANN/IANA section have a higher privilege and a more relaxed relevance threshold compared to the new suffixes proposed for private section domains?

dnsguru:
I agree.  Let the .post administrators direct the PSL on the configuration so it is intentional.

ALSO, in reading the table 2 from 4.9 of the registry documentation where the specify com/edu/org, we failed to include rule 2 which would have added a lot of 2 char subspace entries as well (ie us.post for United States, it.post for Italy, etc) it is not identical 1-1 with ISO 3166

groundcat:
Upon reviewing #270 (thank you for pointing me to this PR) and the registry policy, I have realized that `.post` includes not only `com.post`, `edu.post`, and `org.post` but also all `{country_code}.post` suffixes. This would make the PSL terribly long if all of these were added!

I agree that for such heavily structured TLDs, it is necessary to reconsider. Personally, I would be against adding the country-coded ones (such as `ch.post` or `italy.post`, as mentioned in the registry policy) to the PSL at the moment. Perhaps for heavily structured TLDs, we do need to evaluate relevance before selectively deciding which ones to add to the PSL.

Hence, I am going to close this PR for now, not only due to the risk of inflating the PSL to an impracticable length but also because the `com.post`, `edu.post`, and `org.post` suffixes seem to have no visible usage. We can refer back to this if the `.post` registry decides to submit directly to us, at which point we will reconsider it.

dnsguru:
As a bit of institutional knowledge, .name, if one followed the registry rules, would contain subdomains for around 200k surnames, as they accept john.smith.name and jane.smith.name as registrations - we opted at the time to allow the registry to request it at such time they intended.  Its been 20 years

Issue #2255, opened 2024-11-13
Alphabetising the ICANN section


wdhdev:
I've noticed a lot of the entries in the ICANN section are unalphabetised. Is there a reason for as to why the ICANN section isn't alphabetised however the Private section must be?

Should we write a script to alphabetise it, or is there a specific reason as to why it shouldn't be?

PR #2287, opened 2024-11-23
chore: Fix Alphabetizing for the ICANN section

groundcat:
This PR is to address #2255  

No entry has been modified, which can be verified with 

```
diff <(sort public_suffix_list.dat) <(sort public_suffix_list_sorted.dat)
```


wdhdev:
Are you able to share the script you used to sort it?

groundcat:
Script used:
https://gist.github.com/groundcat/6d81a42e009468c24a80d1c57a7211a8

simon-friedberger:
We should probably make sure that `psltool` also checks this now so it doesn't regress again.

simon-friedberger:
> We should probably make sure that `psltool` also checks this now so it doesn't regress again.

@groundcat Would you be interested in trying?

groundcat:
> > We should probably make sure that `psltool` also checks this now so it doesn't regress again.
> 
> @groundcat Would you be interested in trying?

Thank you for the suggestion regarding psltool checks. I should note that I'm not competent in Go, unfortunately.

Additionally, there might be a structural consideration - the ICANN section is formatted quite differently from the private section that the parser was originally designed for. The ICANN entries don't include email addresses or require TXT records that are present in the private section. Simply appending the same cleaning for ICANN entries would likely trigger failures in the PR checker due to these fundamental format differences. (@danderson correct me if I'm wrong though!)

https://github.com/publicsuffix/list/blob/e1b1608547b24bb84034c1083060e3293e7907d0/tools/internal/parser/clean.go#L34-L49

If someone more experienced with Go would like to take on this task, they might need to account for these structural differences in the parser implementation.

danderson:
Happy tp send a patch for this if you'd like. I didn't do sorting in the ICANN section at the request of maintainers, when we were just getting started.

I _believe_ the formatter should be able to reformat the ICANN section without causing validation errors, the validation and the reformatting are separate passes. It might, however, cause some additional unexpected reflowing of the suffixes because the ICANN section is full of fairly old conventions for suffix layout, and I'm not sure the formatter takes care of all of them.

danderson:
Just tried it. The change is mostly trivial, but formatting the ICANN section triggers a bunch of lint errors due to inline comments that prevent full sorting. The easiest fix for that, for now, is to just not produce those warnings when sorting the ICANN section, unless we want to try and remove them... But in the ICANN section a lot of those comments are load-bearing subsection separators, see for example the comments in the .aero section.

danderson:
psltool's sorting also adds a few more changes compared to this PR, all related to accented letters: https://gist.github.com/danderson/cb0fb30acbec38651cf8c0a1f4275125.

This isn't too surprising, accented letter sorting is one of the big things that change between collations, there is no "correct" answer. But, this is the collation we picked for psltool so :shrug: 

Assuming we're okay with the additional diff, I suggest merging this change (with or without the accents diff applied) and I'll send a separate PR to make psltool sort the ICANN section to maintain the ordering?

danderson:
https://github.com/publicsuffix/list/pull/2295 updates psltool to enforce ICANN section ordering to match this PR. The accent sorting changes above need to be merged one way or another, either in this PR or I can add them to the psltool PR, your call :)

groundcat:
> #2295 updates psltool to enforce ICANN section ordering to match this PR. The accent sorting changes above need to be merged one way or another, either in this PR or I can add them to the psltool PR, your call :)

Just [applied](https://github.com/publicsuffix/list/pull/2287/commits/84a741d0035d3d2199e3902eb4461ee49ac66982) the diff patch to this PR :)

PR #2295, opened 2024-11-28
tools/internal/parser: enforce suffix ordering in the ICANN section

danderson:
The ordering rule is slightly different from the private section: suffix blocks do not get reordered, only the suffixes within each block. Additionally, the ICANN section contains many load-bearing inline comments. To avoid semantic changes, the sorting does not move suffixes over such comments. Unlike in the private section, such movement failures aren't treated as errors, because the subsection comments in the ICANN block have a reasonably consistent meaning.

---

This will fail CI until #2287 is merged, along with the additional sorting fixups in https://gist.github.com/danderson/cb0fb30acbec38651cf8c0a1f4275125 for IDNA domains. Depending on whether #2287 is merged with or without the accent fixes, I can add the remaining fixups in this PR, up to you @simon-friedberger @groundcat.

### DIFF
--- a/tools/internal/parser/clean.go
+++ b/tools/internal/parser/clean.go
@@ -19,39 +19,42 @@ import (
 // changes, Clean applies as many changes as possible then returns
 // errors describing the cleanups that could not take place.
 func (l *List) Clean() []error {
-	return cleanBlock(l)
+	return cleanBlock(l, true)
 }
 
-func cleanBlock(b Block) []error {
+func cleanBlock(b Block, reportCommentBlockages bool) []error {
 	var ret []error
 
 	switch v := b.(type) {
 	case *List:
 		for _, child := range v.Blocks {
-			ret = append(ret, cleanBlock(child)...)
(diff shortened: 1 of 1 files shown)


dnsguru:
Just a note about ICANN section  that there are some ccTLD with
authoritatuve subsections (janet in .uk for example)

simon-friedberger:
Does anybody know how I can rebase this on current main to rerun the checks? It seems Github is checking this PR as if it would be applied to the old main which is just useless.

wdhdev:
It should be fine to merge regardless, it won't merge into 'the old main'.

simon-friedberger:
> Just manually update the branch, if you have the 'Recommend Branch Updates' option enabled in settings it should show up with a prompt to update the branch. Also it won't merge into 'the old main', even with an outdated branch.

Agreed, but it makes the check a lot less useful if it doesn't actually check the result of merging.

danderson:
Oops, sorry I didn't see the main branch got renamed. I'll update my clones to match in future.

PR #1395, opened 2021-08-19
Remove mil.ru, edu.ru, int.ru, gov.ru, ac.ru, test.ru

quidcain:
* [x] Reason of this Pull Request
* [x] Run Syntax Checker (make test)

* [x] Each domain listed in the PRIVATE section has and shall maintain at least two years remaining on registration, and we shall keep the _PSL txt record in place

__Submitter affirms the following:__ 
  * [x] We are listing any third party limits that we seek to work around in our rationale such as those between IOS 14.5+ and Facebook (see [Issue #1245](https://github.com/publicsuffix/list/issues/1245) as a well-documented example)
  * [x] This request was _not_ submitted with the objective of working around other third party limits
  * [x] The [Guidelines](https://github.com/publicsuffix/list/wiki/Guidelines) were carefully _read_ and _understood_, and this request conforms
  * [x] The submission follows the [guidelines](https://github.com/publicsuffix/list/wiki/Format) on formatting
  * [x] Yes, I understand.  I could break my organization's website cookies etc. and the rollback timing, etc is acceptable.  Proceed.
---

Reason of this Pull Request
====
https://ru.wikipedia.org/wiki/%D0%94%D0%BE%D0%BC%D0%B5%D0%BD%D1%8B_%D0%B4%D0%BB%D1%8F_%D0%BF%D1%80%D0%B8%D0%BC%D0%B5%D1%80%D0%BE%D0%B2#test.ru

Here some excerpt from the russian wiki page (unfortunately i can't seem to find public information in english):

>On June 5, 2018, the Council of the Coordination Center for TLD RU decided to cancel the special status of mil.ru, edu.ru, int.ru, gov.ru, ac.ru, test.ru domain names from September 1, 2018

We can even open working websites mil.ru, edu.ru, gov.ru, which means that these are not public suffix domains (correct me if I am wrong)

DNS Verification via dig
=======

make test
=========

passed

PR #2301, opened 2024-12-02
remove `test.ru`

wdhdev:
Originally moved from the ICANN section to the PRIVATE section in #815.

Reasons for removal:
- We do not allow testing domains as per the guidelines:
> Projects that are smaller in scale or are temporary or seasonal in nature will likely be declined. Examples of this might be private-use, sandbox, test, lab, beta, or other exploratory nature changes or requests. It should be expected that despite whatever site or service referred a requestor to seek addition of their domain(s) to the list, projects not serving more then thousands of users are quite likely to be declined.

- No SSL certificates found: https://crt.sh/?q=.test.ru (the first `.` is required, otherwise it returns results from unrelated domains for some reason, you can verify this yourself here: https://crt.sh/?q=test.ru)
- Domain is not delegated as per a WHOIS lookup: `status: REGISTERED, NOT DELEGATED, VERIFIED`
- No sites found when searching `site:test.ru` on Google (not a valid reason for removal, but worth noting)
- No subdomains found when using https://subdomainfinder.c99.nl/

It should be safe to remove it as it does not seem to have any usage, especially as the domain is not even delegated according to WHOIS.

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -12501,7 +12501,6 @@ edu.ru
 gov.ru
 int.ru
 mil.ru
-test.ru
 
 // COSIMO GmbH : http://www.cosimo.de
 // Submitted by Rene Marticke <rmarticke@cosimo.de>


groundcat:
Possibly related comment: https://github.com/publicsuffix/list/pull/1395#issuecomment-902531513

> Domain names ac.ru, edu.ru, gov.ru, int.ru, mil.ru, test.ru is no special-reserved by policy after 2018-09-01.
> But part of these domain names still use by registrants for third level. We (Registry for TLD .RU and .РФ) still have opinion that the ac.ru, edu.ru, gov.ru, int.ru, mil.ru, test.ru must be at PRIVATE section of Public Suffix List.

@cctldru could you please confirm whether `test.ru` can be safely removed from the Public Suffix List?

groundcat:
> I simply found it manually, ran a WHOIS lookup, and noticed it wasn't delegated. In fact quite a lot of my removal PRs, believe it or not, have just been from me manually searching through the list. I get quite bored 🤣

By the way, I have created a script that might be helpful - it runs via GitHub Actions every other day. If you are interested, you are welcome to check [these data files](https://github.com/groundcat/PSL-private-domains-checker/tree/main/data) (particularly nxdomain.csv and expired.csv) regularly to identify any entries that may have recently become obsolete due to Set-and-Forget :)

wdhdev:
I am aware of the repo and do use it occasionally, it's quite helpful.

However sometimes I'm just extremely bored so I decide to do it manually 🤣

wdhdev:
@simon-friedberger Do you think this can be merged? It seems safe to do so. Thank you!

PR #1446, opened 2021-10-02
Adding upli.io

lennybakkalian:
* [x] Description of Organization
* [x] Reason for PSL Inclusion
* [x] DNS verification via dig
* [x] Run Syntax Checker (make test)

* [x] Each domain listed in the PRIVATE section has and shall maintain at least two years remaining on registration, and we shall keep the _PSL txt record in place

__Submitter affirms the following:__ 
  * [x] We are listing any third party limits that we seek to work around in our rationale such as those between IOS 14.5+ and Facebook (see [Issue #1245](https://github.com/publicsuffix/list/issues/1245) as a well-documented example)
  * [x] This request was _not_ submitted with the objective of working around other third party limits
  * [x] The [Guidelines](https://github.com/publicsuffix/list/wiki/Guidelines) were carefully _read_ and _understood_, and this request conforms
  * [x] The submission follows the [guidelines](https://github.com/publicsuffix/list/wiki/Format) on formatting

---
__For Private section requests that are submitting entries for domains that match their organization website's primary domain:__

``` 
Seriously, carefully read the downline flow of the PSL and the guidelines.
Your request could very likely alter the cookie and certificate (as well as other) behaviours on your 
core domain name in ways that could be problematic for your business.

Rollback is really not predicatable, as those who use or incorporate the PSL do what they do, and when.
It is not within the PSL volunteers' control to do anything about that.  

The volunteers are busy with new requests, and rollbacks are lowest priority, so if something gets broken 
it will stay that way for an indefinitely long while.
```
(Link: [about propogation/expectations](https://github.com/publicsuffix/list/wiki/Guidelines#appropriate-expectations-on-derivative-propagation-use-or-inclusion))

 * [x] Yes, I understand.  I could break my organization's website cookies etc. and the rollback timing, etc is acceptable.  Proceed.
---

Description of Organization
====

Organization Website: https://upli.io
upli is a service provider for creating static websites. These websites are built from github repositories and then hosted under their own domain or a subdomain provided by upli. for example, lenny.upli.io - In addition, in the distant future it will be possible to rent virtual servers and create lambda functions (similar to aws lambda).

Reason for PSL Inclusion
====
Since each user has his own subdomain, it should be forbidden to create supercookies to compromise the main site. Even if on the main page the session token is stored in the localstorage, because not all browsers contain the public suffix list. Nevertheless we want to separate all subdomains via the PSL.

DNS Verification via dig
=======

```
dig +short TXT _psl.upli.io
[...]

PR #2302, opened 2024-12-02
remove `upli.io`

wdhdev:
Reasons for removal:
- No public website at https://upli.io (which is listed as the org website)
- No subdomains can be found either by searching `site:upli.io` on Google or using a subdomain finder like https://subdomainfinder.c99.nl
- No SSL certificates have been issued for subdomains since 2021-10-22, which was a wildcard SSL certificate, which expired 2022-01-20. The only SSLs issued since then seem to be auto-renewed for the root domain: https://crt.sh/?q=upli.io

I think it is safe to remove this domain as it does not seem to be used, nor have any SSLs been issued for subdomains in over 3 years. The purpose of the original inclusion in #1446 was for customer subdomains, however none were found.

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -15548,10 +15548,6 @@ rs.ba
 // Submitted by Marcin <dns@ath.bielsko.pl>
 bielsko.pl
 
-// Upli : https://upli.io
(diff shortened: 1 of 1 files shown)


wdhdev:
@simon-friedberger Removal was confirmed by original requester: https://github.com/publicsuffix/list/pull/1446#issuecomment-2511067033

PR #2312, opened 2024-12-09
remove `shiftedit.io`

wdhdev:
Reasons for removal:
- Domain expired 2024-10-27
- No results on Google
- No results on https://subdomainfinder.c99.nl/

Note: The domain is in a redemption period, according to a WHOIS lookup, however it seems they will not be renewing it as it expired over a month ago and still hasn't been renewed.

Pretty basic removal, domain is expired, should be safe to remove.


wdhdev:
@simon-friedberger Can this be merged?

PR #2313, opened 2024-12-09
remove `herokussl.com`

wdhdev:
Reasons for removal:
- Heroku no longer uses `herokussl.com` for SSL certificates, it now uses `herokudns.com` - you see no mention of the domain in their support article (last updated 2nd February, 2024) on SSL certificates: https://devcenter.heroku.com/articles/ssl
- Another article, on TLS, shows `herokudns.com` being used instead of `herokussl.com`: https://devcenter.heroku.com/articles/understanding-tls-on-heroku
- There have been no SSL certificates issued on the domain itself either since 2019: https://crt.sh/?q=herokussl.com (note: no SSLs have been issued on `herokudns.com` since 2020 either, however it is actively shown and used in documentation regarding SSL)

There does not seem to be any evidence of usage of this domain and it should be safe to say it is no longer in use and can be removed.

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -13718,7 +13718,7 @@ heliohost.us
 hepforge.org
 
 // Heroku : https://www.heroku.com/
-// Submitted by Tom Maher <tmaher@heroku.com>
+// Submitted by Shumon Huque <public-dns@salesforce.com>
 herokuapp.com
 
 // Heyflow : https://www.heyflow.com
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -13720,7 +13720,6 @@ hepforge.org
 // Heroku : https://www.heroku.com/
 // Submitted by Tom Maher <tmaher@heroku.com>
 herokuapp.com
-herokussl.com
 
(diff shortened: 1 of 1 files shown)


simon-friedberger:
Since heroku is still around and professionally operated, can you please confirm with them?

wdhdev:
Yep, I have emailed their security email, as well as the contact email address they have listed here. If we don't receive a response with-in 1-2 weeks I think we should proceed with this request.

wdhdev:
@simon-friedberger @dnsguru This PR was approved by Heroku, email chain has been forwarded to both of you.

tmaher:
Amusing timing given I just submitted #2315 and happened to look at your PR queue

I affirm I was the Tom Maher who submitted both Heroku domains originally, and left Heroku/Salesforce ten years ago. I apologize for not submitting a role email address originally.

Thank you for your work maintaining the list.

wdhdev:
I didn't even realise that was you, I should pay more attention to usernames, what a coincidence! No worries about the role based email address, we never actually required it up until recently, there is plenty of older entries that still use personal email addresses.

Issue #2265, opened 2024-11-17
Potentially Outdated Entries in ICANN (IANA/ICP-3) Section


groundcat:
I have noticed that the community has been working on fixing outdated domains in the ICANN (IANA/ICP-3) section. It might be helpful to take a more systematic approach to review all potentially outdated entries in this section. That said, validating these domains and reaching out to the TLD registries will likely require additional volunteer support.

Below is a list of ICANN section entries currently returning NXDOMAIN errors. This could suggest that the suffix has been decommissioned, but being on this list does not automatically mean a domain should be removed. Domains should only be removed after their status is directly confirmed by the respective TLD operator.

The goal of this list is to encourage volunteers to help verify whether these domains are still active—whether they are available for registration as second-level domains (SLDs), not available for registration but still in use, or reserved for internal networking or other specific purposes. Only after direct confirmation with the relevant registry should any domains be removed.

## ICANN (IANA/ICP-3) Section Domains Returning NXDOMAIN Error

https://gist.github.com/groundcat/cff0e30dd90e17ccc4a47b42a25e5ddc

## Possibly Affected ccTLDs

- ac   - Email inquiry sent on 2024-11-17 by @groundcat
- am  - 2024-11-18 Confirmed by `ISOC AM <isoc@isoc.am>` : "Regarding the updated list, we confirm that no changes have occurred."
- ao   - Email inquiry sent on 2024-11-17 by @groundcat
- au   - #2267
- az   - Consistent with [authoritative registry source](https://whois.az/?page_id=10)
- bo  -  #2276
- ci   - https://github.com/publicsuffix/list/pull/2198#issuecomment-2437405916
- co   - #2269
- cw   - #2281
- dm   - #2277
- eg  != [authoritative registry source](https://domain.eg/en/domain-rules/subdomain-names-types/) - Email inquiry sent on 2024-11-17 by @groundcat #2290
- ge  - #2283
- gu   - Email inquiry sent on 2024-11-17 by @groundcat
- im   - Email inquiry sent on 2024-11-17 by @groundcat
- is   - #2266
- it  - Email inquiry sent on 2024-11-17 by @groundcat
- jo - #2264
- jp - Confirmed by registry
- km - Email inquiry sent on 2024-11-17 by @groundcat
- kp   - Email inquiry sent on 2024-11-17 by @groundcat
- mg   -  #2274
[...]

PR #2316, opened 2024-12-09
update `.us`

wdhdev:
I now have root zone file access for the .us TLD, so I have decided to update the entries for it as per #2265.

- `fed.us` - Removed as there is no existing DNS records under this name (at `fed.us` or any higher levels like `example.fed.us`)
- `kids.us` - There is a single website hosted under this which is at the root (http://kids.us) which shows the 2LD was shut down back in 2012, here is the link to the suspension document: https://www.ntia.doc.gov/files/ntia/publications/ustld_27_jun_2012_mod_012-1.pdf

I have not modified geographic names at this point as I have no effective way to filter them out at this point in time. However I have gone around and updated some comments in that section.

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -5968,13 +5968,11 @@ police.uk
 *.sch.uk
 
 // us : https://www.iana.org/domains/root/db/us.html
+// Confirmed via the .us zone file by William Harrison - 2024-12-10
 us
 dni.us
-fed.us
 isa.us
-kids.us
-nsn.us
-// us geographic names
+// Geographic Names
 ak.us
 al.us
 ar.us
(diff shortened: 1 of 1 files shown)


simon-friedberger:
If you open requests like this there is very little paper trail explaining the decision for future maintainers. Could you at least document what you did to get and validate the zone file?

I would like to do something more principled but cannot think of anything right now.

wdhdev:
Ah, I forgot to mention. I emailed the IANA contact for .us, where I was redirected to GoDaddy Registry, who made me sign a contract, granting me access to the .us zone file.

I doubt I'm allowed to actually share the zone file itself, however if there's any specific information you need I can give it to you, assuming it doesn't violate the contract.

simon-friedberger:
So, anybody can just sign that contract to get access but sharing is not allowed? 🤔

wdhdev:
The contract states I'm not allowed to directly share it, according to clause 4.D:
```
(d) not distribute the Data you obtained under this Agreement or any copy thereof to any other party without the express prior written consent of RS LLC, except that you may redistribute the Data insofar as it has been incorporated by you into a value-added product or service that does not permit the extraction of a substantial portion of the Data from the value-added product or service, provided you prohibit the recipient of the Data from using the Data in a manner contrary to Section 4(a).
```

simon-friedberger:
Pretty silly if anybody can just sign the contract to get the same information.

wdhdev:
You have to have a reason for accessing the data, they don't just send out the contract for no reason.

PR #2317, opened 2024-12-09
remove `curv.dev`

wdhdev:
Originally added in #968.

Reasons for removal:
- `_psl.curv.dev` TXT record no longer resolves, which provokes the that idea that PSL status is no longer required, as this TXT record was present when the domain was added to the PSL.
- No results on Google or Bing for `site:curv.dev`, despite the reason for inclusion being that customer websites will be on this domain.
- Few results show up from a subdomain scan, most of which do not resolve properly, if at all: https://subdomainfinder.c99.nl/scans/2024-12-10/curv.dev (also, it is well below our required threshold for inclusion)
- Very few results show up when searching for SSL certificates under the domain: https://crt.sh/?q=curv.dev&exclude=expired&group=none

It is quite evident this domain is not being actively used and should be able to be safely removed.

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -12531,10 +12531,6 @@ on.crisp.email
 // Submitted by Andrew Cady <public-suffix-list@cryptonomic.net>
 *.cryptonomic.net
 
-// Curv UG : https://curv-labs.de/
-// Submitted by Marvin Wiesner <Marvin@curv-labs.de>
-curv.dev
-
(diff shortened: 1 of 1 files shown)


simon-friedberger:
I think we should not remove entries without contacting owners to confirm.

wdhdev:
I've sent an email to the contact email address.

wdhdev:
@simon-friedberger No response from the owners. I think this can be removed safely.

PR #1074, opened 2020-07-22
Add Danger Science Group domains

skylarmacdonald:
* [x] Description of Organization
* [x] Reason for PSL Inclusion
* [x] DNS verification via dig
* [x] Run Syntax Checker (make test)

Description of Organization
====

Organization Website: https://dangerscience.com

We are a software development company that hosts some of our own web applications, including the [Folio Network](https://folio.network) portfolio site creator, and a not-yet-publicly-available PaaS platform called Platform 0.

Reason for PSL Inclusion
====

The included domain names can be used as subdomains by customers to create websites/applications, and we would like to make it harder for cookies to be set across multiple of these sites in case of nefarious activity.

`folionetwork.site` and `fnwk.site` are used by [Folio Network](https://folio.network) users for their portfolios, and `platform0.app` will be the default domains on Platform 0-hosted applications.

DNS Verification via dig
=======

```
dig +short TXT _psl.folionetwork.site
"https://github.com/publicsuffix/list/pull/1074"
```

```
dig +short TXT _psl.fnwk.site
"https://github.com/publicsuffix/list/pull/1074"
```

```
dig +short TXT _psl.platform0.app
"https://github.com/publicsuffix/list/pull/1074"
```

make test
=========

Ran the test locally and all was well. [Travis is happy too.](https://travis-ci.org/github/publicsuffix/list/builds/710927980)

PR #2318, opened 2024-12-10
remove Danger Science section

wdhdev:
Reasons for removal:
- No active SSL certificates on `fwnk.site` or `folionetwork.site`: https://crt.sh/?q=fnwk.site&exclude=expired https://crt.sh/?q=folionetwork.site&exclude=expired
- Very few active SSL certificates found on `platform0.app`: https://crt.sh/?q=platform0.app&exclude=expired
  - I believe they may issue SSLs on a per domain basis, and only use wildcards on the occasion where no website exists using a wildcard (e.g. https://exampletest123.platform0.app)
- In the original PR #1074, the domains `fwnk.site` and `folionetwork.site` were intended to be used by https://folio.network, however the domain does not have any A/AAAA records, therefore it doesn't resolve.
- No results on Google or Bing for any of the domains when searching using `site:{domain}`.
- `_psl.fwnk.site` is missing, however for the other 2 domains it is still in place.

It is evident that `fwnk.site` and `folionetwork.site` are not in use any more. `platform0.app` does appear to get an extremely small amount of use, however it is nowhere near our threshold for inclusion and should be removed. I think these domains can be safely removed.

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -12544,12 +12544,6 @@ cfolks.pl
 cyon.link
 cyon.site
 
-// Danger Science Group: https://dangerscience.com/
-// Submitted by Skylar MacDonald <skylar@dangerscience.com>
-platform0.app
-fnwk.site
-folionetwork.site
-
 // Dansk.net : http://www.dansk.net/
 // Submitted by Anani Voule <digital@digital.co.dk>
 biz.dk


wdhdev:
Email sent to the contact address in the section regarding removal.

wdhdev:
@simon-friedberger @dnsguru I have forwarded the email chain to your email addresses. This PR is good to merge and has been approved by the original submitter.

Issue #2322, opened 2024-12-11
Removal of domains with a hold status


wdhdev:
@simon-friedberger @groundcat I would like to get your guys' thoughts on removing domains that have a hold status in WHOIS.

The reasons why I believe they should potentially be removed is:
- DNS cannot be resolved under these domains
- They tend to not be in use, especially if they have had a hold status for an extended period of time
- If a domain is actively in use, they would not set a clientHold status on the domain
- Said domain may not be renewed, as some domains are placed on hold due to missed payments

Some reasons against removing these domains:
- Domain may just be inactive for a certain period of time before being active again
- Could be a registry error or something of that nature

Currently there appears to be 2 domains listed on the PSL that have been in a hold status for multiple weeks now (according to https://github.com/groundcat/PSL-private-domains-checker/blob/main/data/hold.csv)

- `001www.com` - clientHold status
- `now-dns.top` - serverHold status

In this specific case I believe these domains should be removed as they have had a hold status for a while now, however this issue is primarily targetting any future domains that may receive a hold status.

groundcat:
I am generally comfortable with removing entries that have remained in a hold status for an extended period, such as 45 days. My checker occasionally detects such hold statuses, but most of them are resolved within a day or two, typically due to UGC violations on their subdomains. 

As long as the business is operating normally or the project is active, these issues should be fixed relatively quickly. If a domain remains in serverHold or clientHold status for that long, it most likely indicates that either the domain is permanently blocked (with no intention from the registrar or registry to reinstate it) or the project has been abandoned. 

So, I believe 45 calendar days is a reasonable limit, which somewhat mirrors the commonly used "45 calendar days" grace period given after a domain expires before it enters pendingDelete status in the registry industry.

groundcat:
I agree that for `001www.com` with clientHold status and `now-dns.top` with serverHold status, removal would be appropriate since they have been in hold status for an extensive period. While the requestor indicated in September in issue #2178 that the project was still active, three months have passed without any progress in resolving these holds with their registrar or registry. It appears safe to remove them from the list. However, since removing domains with hold status is not yet documented as an official policy on the wiki or website, it might be better to inform them about the planned removal first, so that they will understand why their domains are being removed from the PSL and know that they can resubmit them to the list if the hold status is resolved in the future.

PR #2324, opened 2024-12-11
remove `001www.com` and `now-dns.top`

wdhdev:
Both domains have had a hold status for an extended period of time, likely over 45 days, and the Now DNS project is still an active project however has appeared to make no attempt to remove the hold on these domains.

See #2322 for more information, specifically https://github.com/publicsuffix/list/issues/2322#issuecomment-2533488209.

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -14518,7 +14518,6 @@ notion.site
 dnsking.ch
 mypi.co
 n4t.co
-001www.com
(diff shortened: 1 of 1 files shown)


groundcat:
The domain `n4t.co` will most likely warrant removal as well because it expired on `2024-12-11`, and is no longer listed as an available suffix on https://now-dns.com/

PR #2325, opened 2024-12-11
Update comments for `.it`, `.jp`, `.se`, `.xn--fiqs8s`, `.xn--fiqz9s`

groundcat:
This PR is created to update comments for `.it`, `.jp`, `.se`, `.cn` to fix broken links and note some registry confirmation responses to improve documentation.

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -1938,9 +1938,9 @@ lg.jp
 ne.jp
 or.jp
 // jp prefecture type names
-// 2024-11-22: JPRS confirmed that since regional JP domain names are no longer 
-// accepting new registrations, any registered domain names listed under FQDNs 
-// are confirmed to have been fully discontinued will be removed from the PSL.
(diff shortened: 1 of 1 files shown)


simon-friedberger:
> // 2024-11-22: JPRS confirmed that since regional JP domain names are no longer 
> // accepting new registrations, any registered domain names listed under FQDNs 
> // are confirmed to have been fully discontinued will be removed from the PSL.

I am not sure what you are saying here. Could you please rephrase? 😅

wdhdev:
I agree, the wording is a bit confusing. Did you mean something like `Regional .jp domain names no longer accept registrations, however we are waiting until all domains under them are expired, before we can remove them from the PSL.`?

groundcat:
Sorry, it was confusing since I had translated it literally. From what I understand, they were trying to say that regional .jp suffixes no longer accept new registrations. Once all existing registrations expire (marking full discontinuation), these suffixes will be removed from the PSL.

groundcat:
> Could you please rephrase?

Done :)

PR #2100, opened 2024-08-11
remove `blogspot.mr`

wdhdev:
## blogspot.mr
I'm not sure where this was originally added in, possibly added pre-GitHub. 

Pretty simple reason for removal; the domain name is expired.

When attempting to lookup the domain via WHOIS, the server returns:
```
$ whois blogspot.mr
Server can't process your request at the moment.
```
And to clarify, I don't believe this is a server issue as looking up, for example, `nic.mr` works perfectly fine and returns the WHOIS data.

When attempting to lookup NS records the server returns `NXDOMAIN`.

PR #2327, opened 2024-12-11
Remove `blogspot.*` ccTLD domains

groundcat:
My contact at Google attempted to internally reach out to the Blogger team on November 17, 2024, but we have not received any response after nearly a month, suggesting the service is no longer actively managed. Therefore, removal should be safe to proceed.

Their ccTLD domains have begun to expire without renewal and are being released and deleted by the registry (for example, #2100). Rather than waiting for each domain to expire and create incremental debris, it is more efficient to remove them all at once.

I conducted a quick check (as documented in [samples_result.txt](https://github.com/user-attachments/files/16578078/samples_result.txt)) which revealed that all ccTLD domains are redirecting (301/302) to blogname.blogspot.com domains. This indicates they have completely decommissioned the use of ccTLD Blogspot domains, with all redirects now pointing to the main blogspot.com domain. 

Consequently, PSL inclusion is no longer necessary.


simon-friedberger:
@domfarolino Please state an opinion on this! :)

domfarolino:
This lgtm.

wdhdev:
The Blogspot ccTLD domains do seem to be lapsing, e.g. `blogspot.cv` was registered 13/06/2024. They should be removed as soon as possible as Google likely no longer controls some of these.

PR #2328, opened 2024-12-11
chore(tools/internal/parser): remove unneeded exceptions

wdhdev:
The domains/sections removed from here have been removed from the PSL, except for is-a.dev, as that section is now correctly sorted.

I would leave this PR open until we merge a few of the other open removal PRs such as #2313, so we can update exceptions.go accordingly.

Let me know if I've missed any that no longer need to be here.


wdhdev:
In 584dc94502353b324e63fb6523900021f4ddb79a, the reason why the `incorrectSort` array is now empty, is because all entries *are* sorted correctly, so there is no need for them to be exempt.

I removed the Amazon exceptions as all their domains do in fact have `_psl` TXT records, and they consistently keep them in the zone, along with any new PRs they add all the TXT records (thank you Amazon :heart:), hence there is no reason to have an exception for them.

wdhdev:
@simon-friedberger I think this can be merged. The other PRs open that may affect the exceptions by removing old domains, may take a little while before they are merged, so might as well remove these ones and we can worry about the other ones later.

PR #2330, opened 2024-12-11
add confirmation message to `.am`

wdhdev:
#2265

> am - 2024-11-18 Confirmed by ISOC AM <isoc@isoc.am> : "Regarding the updated list, we confirm that no changes have occurred."

(@groundcat sent the email, not me, so if you want the email chain contact him.)

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -165,6 +165,7 @@ net.al
 org.al
 
 // am : https://www.amnic.net/policy/en/Policy_EN.pdf
+// Confirmed by ISOC AM <isoc@isoc.am> 2024-11-18
 am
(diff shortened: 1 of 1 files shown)


groundcat:
Adding email here for documentation:

From `ISOC AM <isoc@isoc.am>` Nov 18, 2024:

>    According to point 15 of  domain name registration policy for ".am" and ".hay" domains, only the municipalities of the Republic of Armenia are eligible to register third-level domain names under ".commune.am." 
> 
> No other entity or individual is permitted to register a ".commune.am" third-level domain name.
> 
> Regarding the updated list, we confirm that no changes have occurred.

PR #2331, opened 2024-12-11
update `.az` section

wdhdev:
The [IANA page](https://www.iana.org/domains/root/db/az.html) shows `whois.az` as the authoritative source for the TLD.

List confirmed via https://whois.az/?page_id=10 section 2.2

`pro.az` is no longer available for registration, however according to the [statistics page](https://whois.az/?page_id=783) it still has some registered domains. We will keep it for legacy reasons.


wdhdev:
@simon-friedberger Can this be merged?

PR #807, opened 2019-04-25
Add lab.ms

tomkln:
* [x] Reason for PSL Inclusion
* [x] Description of Organization
* [x] DNS verification via dig
* [x] Run Syntax Checker (make test)

Description of Organization
====

Organization Website: https://www.gentlent.com

I am a director and engineer at Gentlent, which is a brand new software development and consulting company.

Reason for PSL Inclusion
====

We want to provide "lab.ms" domains for our upcoming PaaS hosting aswell as for DynDNS soon. Users should be able to only set cookies on their .lab.ms domain and browsers should recognize the domains with the suffix being lab.ms.

DNS Verification via dig
=======

The verification of the TXT record was successful:
```
dig +short TXT _psl.lab.ms
"https://github.com/publicsuffix/list/pull/807"
```

make test
=========

We run the "make test" successfully. The output may be viewed at https://gist.github.com/tomkln/8e54f5fcb44fcf3d34ca303790b5255a.

PR #2334, opened 2024-12-12
remove `lab.ms`

wdhdev:
Originally added in #807, back in 2019.

Reasons for removal:
- The domain is no longer controlled by `Gentlent, Inc.`.
	- The domain's registration date via a WHOIS lookup is `2024-07-25`, which is very recent, when the domain was added to the PSL around 5 years ago. The new registrant's name is `Flame Lab Initiative` according to WHOIS, which is not the original registrants.

Should be safe to remove, as it is no longer owned by the original registrant.

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -13388,7 +13388,6 @@ usercontent.jp
 // Submitted by Tom Klein <tom@gentlent.com>
 gentapps.com
 gentlentapis.com
-lab.ms
 cdn-edges.net
 
 // GignoSystemJapan: http://gsj.bz


groundcat:
- The registration date exceeds the PSL inclusion date, indicating domain re-registration and ownership change. This condition alone warrants removal.
- While lab.ms is operational, none of its subdomains are currently accessible through Google. The site appears to be a different project now, suggesting PSL inclusion is no longer necessary.
- The email address listed on the site is unreachable due to missing MX records.

I think we should review other entries in this `Gentlent, Inc.` block.

wdhdev:
I'll have a look into the other entries in the block later 👍

wdhdev:
@simon-friedberger This can be merged.

@groundcat I've had a look at the other entries, they don't seem to be used, so I've commented on one of their PRs asking for them for confirmation if they still need the entries.

PR #849, opened 2019-07-16
Add lelux.site

theel0ja:
* [x] Description of Organization
* [x] Reason for PSL Inclusion
* [x] DNS verification via dig
* [x] Run Syntax Checker (make test)

Description of Organization
====

Organization Website: https://lelux.fi/

Organization providing services powered by free software. (related: https://libreho.st/)

I'm the administator of Lelux.fi.

Reason for PSL Inclusion
====

Going to be used for a public GitLab Pages instance.

DNS Verification via dig
=======

```
dig +short TXT _psl.lelux.site
"https://github.com/publicsuffix/list/pull/849"
```

make test
=========

I ran the test and it passed succesfully.

PR #2336, opened 2024-12-12
remove `lelux.site`

wdhdev:
Domain was added in #849, in 2019, however the registration date of the domain is 2023-11-30, it is clear the domain lapsed and has changed registrants.

This domain should be safe to remove.

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -14068,10 +14068,6 @@ leadpages.co
 lpages.co
 lpusercontent.com
 
-// Lelux.fi : https://lelux.fi/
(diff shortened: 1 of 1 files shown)


groundcat:
- The registration date exceeds the PSL inclusion date, which indicates the domain was re-registered and has changed ownership.
- The email address listed in the domain block is no longer reachable.
- No sites on this domain are currently accessible.

Agree that it should be safe to remove.

simon-friedberger:
I agree on this particular entry but just want to remind everyone that registration dates do get updated sometimes.

PR #2173, opened 2024-09-18
Remove inactive or expired yombo domains

groundcat:
This PR is to remove `ybo.faith`, `homelink.one`, `ybo.party`, `ybo.review`, `ybo.science`, and `ybo.trade` due to evidence suggesting domain inactivity.

### General Evidence (applies to all domains):

- Attempts to contact via email (mitch@yombo.net) were unsuccessful – no MX record at `yombo.net`. No mail servers found.
- The organization website (https://yombo.net) is no longer functioning and now points to a stationary print business.
- Unable to identify any email address. `@yombo.net` has no MX.

As identified in #2172:

```
ybo.faith TIMEOUT
ybo.party TIMEOUT
ybo.review SERVFAIL
ybo.science TIMEOUT
ybo.trade TIMEOUT
```

### Domain-Specific Checks:

---

**1. `ybo.faith`**
- The [Google search](https://www.google.com/search?q=site%3Aybo.faith) and [Bing search](https://www.bing.com/search?q=site%3Aybo.faith) show no results for the domain.
- [Certificate Transparency](https://crt.sh/?q=ybo.faith) reveals no active SSL certificates in use.
- ~~Running `dig +short TXT _psl.ybo.faith` no longer returns the required record value.~~ Not required at the time.
- No subdomains were found by third-party tools [[1](https://subdomainfinder.c99.nl)] [[2](https://www.virustotal.com/gui/domain/ybo.faith/relations)].

---

**2. `homelink.one`**
- WHOIS `Creation Date: 2021-12-10T19:56:27Z` > PSL inclusion (pre-GitHub). The WHOIS creation dates for the domains are later than the PSL inclusion dates, suggesting the domains were likely allowed to expire.
- The [Google search](https://www.google.com/search?q=site%3Ahomelink.one) and [Bing search](https://www.bing.com/search?q=site%3Ahomelink.one) show 1 results for the domain.
  - Led to `This domain is available for sale!` page.
- [Certificate Transparency](https://crt.sh/?q=homelink.one) reveals no active SSL certificates in use.
- ~~Running `dig +short TXT _psl.homelink.one` no longer returns the required record value.~~  Not required at the time.
- No subdomains were found by third-party tools [[1](https://subdomainfinder.c99.nl)] [[2](https://www.virustotal.com/gui/domain/homelink.one/relations)].

---

**3. `ybo.party`**
[...]

PR #2337, opened 2024-12-12
remove `yombo.me`

wdhdev:
Domain was originally added in #331, in 2016, however the registration date for the domain is 2023-11-30, suggesting that the domain lapsed (like the other 6 domains removed in #2173) and has changed registrants.

It should be safe to remove this domain.


groundcat:
Agree with the removal. I looked into this domain in ticket #2173, and it has become evident that it is no longer needed for the PSL.

PR #2339, opened 2024-12-12
update `.bw` section

wdhdev:
IANA page: https://www.iana.org/domains/root/db/bw.html

https://nic.net.bw is the listed registry site, and under https://nic.net.bw/bw-name-structure it has a list of all the 2LDs. This is an authoritative source and can be trusted.

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -679,10 +679,12 @@ org.bt
 bv
 
 // bw : https://www.iana.org/domains/root/db/bw.html
-// http://www.gobin.info/domainname/bw.doc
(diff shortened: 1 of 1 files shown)


simon-friedberger:
That site suggests that registering a `.bw` domain is not possible so we should probably add a wildcard instead.

wdhdev:
Sites do exist directly under .bw, you can see them by searching the following on Google: `site:bw -site:co.bw -site:net.bw -site:org.bw -site:ac.bw -site:gov.bw`

 I would not recommend a wildcard in this case.

PR #1549, opened 2022-04-04
Add ibxos.it and iliadboxos.it domains

mmakassikis:
* [x] Description of Organization
* [X] Reason for PSL Inclusion
* [ ] DNS verification via dig
* [X] Run Syntax Checker (make test)

* [ ] Each domain listed in the PRIVATE section has and shall maintain at least two years remaining on registration, and we shall keep the _PSL txt record in place

__Submitter affirms the following:__ 

  * [X] We are listing *any* third-party limits that we seek to work around in our rationale such as those between IOS 14.5+ and Facebook (see [Issue #1245](https://github.com/publicsuffix/list/issues/1245) as a well-documented example)
  * [X] This request was _not_ submitted with the objective of working around other third-party limits

  * [X] The [Guidelines](https://github.com/publicsuffix/list/wiki/Guidelines) were carefully _read_ and _understood_, and this request conforms
  * [X] The submission follows the [guidelines](https://github.com/publicsuffix/list/wiki/Format) on formatting

---
__For Private section requests that are submitting entries for domains that match their organization website's primary domain:__

``` 
Seriously, carefully read the downline flow of the PSL and the 
guidelines. Your request could very likely alter the cookie and 
certificate (as well as other) behaviours on your core domain name in 
ways that could be problematic for your business.

Rollback is really not predicatable, as those who use or incorporate 
the PSL do what they do, and when. It is not within the PSL volunteers' 
control to do anything about that.  

The volunteers are busy with new requests, and rollbacks are lowest 
priority, so if something gets broken by your PR, it will potentially 
stay that way for an indefinitely long while.
```
(Link: [about propogation/expectations](https://github.com/publicsuffix/list/wiki/Guidelines#appropriate-expectations-on-derivative-propagation-use-or-inclusion))

 * [x] *Yes, I understand*.  I could break my organization's website cookies etc. and the rollback timing, etc is acceptable.  *Proceed*.
---

Description of Organization
====

  Iliad Italia is an ISP in Italy.

Organization Website: http://www.iliad.it/

I am making this submission on behalf of my employer.

Reason for PSL Inclusion
====

Iliad subscribers can reserve a custom subdomain (e.g: {subdomain}.iliadbox.it) to access their router. In that context, each subdomain effectively points to a separate device. It is also possible a TLS certificate (possibly LetsEncrypt issued) has been configured to ensure remote access is made in a secure way.

We want to include these domains to ensure proper browser behavior with regards to cookies and certificates when users access their routers.

[...]

PR #2343, opened 2024-12-15
remove `iliadboxos.it`

wdhdev:
Domain name has expired as of `2024-12-14`.

It still appears to [have some use](https://subdomainfinder.c99.nl/scans/2024-12-15/iliadboxos.it), however if it isn't renewed in ~7 days I would say it should be removed.

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -13739,7 +13739,6 @@ iki.fi
 // iliad italia : https://www.iliad.it
 // Submitted by Marios Makassikis <mmakassikis@freebox.fr>
 ibxos.it
-iliadboxos.it
 
 // Incsub, LLC : https://incsub.com/
 // Submitted by Aaron Edwards <sysadmins@incsub.com>


groundcat:
For the `.it` TLD specifically, we cannot rely solely on the expiration date to classify a project as debris, since the `.it` registry has a known practice of delayed WHOIS updates on expiry dates, even when domains are automatically renewed. It may take weeks for the information to update. This behavior is also observed in other ccTLDs such as `.jp` during the auto-renewal period (sometimes shown as auto-renew status in WHOIS, but not always) where the expiration date remains unchanged and does not reflect the actual expiration time.

Additionally, this domain still has several resolvable [subdomains](https://subdomainfinder.c99.nl/scans/2024-12-15/iliadboxos.it), regardless of the expiration date. Based on these reasons, I posted an inquiry in the [comment](https://github.com/publicsuffix/list/pull/1549#issuecomment-2537958375) last week instead of immediately creating a removal PR. 

Since they [indicated they want to maintain their presence in the PSL](https://github.com/publicsuffix/list/pull/1549#issuecomment-2543683290), I believe we should not remove it.

PR #2349, opened 2024-12-19
remove `TwoDNS` block

wdhdev:
The service is shutting down tomorrow (20th December, 2024) according to their website: https://www.twodns.de/en

I will email them asking them to confirm this PR though.

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -15404,22 +15404,6 @@ site.transip.me
 // Submitted by TuxFamily administrators <adm@staff.tuxfamily.org>
 tuxfamily.org
 
-// TwoDNS : https://www.twodns.de/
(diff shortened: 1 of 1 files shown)


wdhdev:
@dnsguru @simon-friedberger This can be merged.

Email correspondence sent to both of you.

simon-friedberger:
@wdhdev I am closing this instead.

1. There are still devices using this, it is only closed for new registrations. We shouldn't break existing setups just to be very proactive.
2. They might start a paid service. In that case the discontinuity of removing it now and adding it again later will just mean that more people have the wrong entries. Plus, it means work now and work later.
3. Being on the PSL is mostly the more secure state so keeping entries on it is generally better.

PR #2350, opened 2024-12-19
remove `4u.com`

wdhdev:
Reasons for removal:
- No subdomains found on Google: `site:4u.com`
- No active subdomains found using a [subdomain finder](https://subdomainfinder.c99.nl/scans/2024-12-19/4u.com)
- No SSL certificates found for any subdomains excluding `www.4u.com`: https://crt.sh/?q=4u.com&exclude=expired&match=all

I believe they are either no longer issuing subdomains of this domain or the domain may have changed registrants. This domain was likely added pre-GitHub as I could not find any PRs for it.

I think this domain can be removed safely, unless anyone has any comments regarding it. cc @groundcat

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -14289,10 +14289,6 @@ netfy.app
 // Submitted by Jessica Parsons <jessica@netlify.com>
 netlify.app
 
-// Neustar Inc.
-// Submitted by Trung Tran <Trung.Tran@neustar.biz>
-4u.com
-
 // NFSN, Inc. : https://www.NearlyFreeSpeech.NET/
 // Submitted by Jeff Wheelhouse <support@nearlyfreespeech.net>
 nfshost.com


dnsguru:
Please leave this while I investigate.  it was, at a point, a subdomain registry customer of Neustar, before GoDaddy Registry acquired the Neustar platform.

dnsguru:
No updates just yet in response to my outreach to Roy, the main contact I had for 4U, by email to legacy personal contacts that I had in my rolodex from operating large domain name conferences.

There was a significant amount of ephemeral usage that the PSL entry delivered cookie separation for, such as where they had in place things like [vin number].4u.com or [phone number].4u.com as I recall their use case being.

Since that time, when they were on Neustar's back end, GoDaddy Registry (GDR) acquired them and I am not certain that there is still a relationship between GDR and 4U.

I recommend not merging this to avoid breakage being caused.

PR #2358, opened 2024-12-25
add PSL Sandbox

wdhdev:
The goal for this section of the PSL is to create a sandbox for testing the affects of different rules on the PSL.

This implements #1349.

The sandbox hasn't been deployed yet, however once these entries are listed I will deploy one so users can test out various differences domains have for when they are on the PSL.

Note: I have added these under subdomains of psl.hrsn.dev in order to prevent any potential problems or overlaps with the existence of the `hrsn.dev` rule. Also, I've added this in a new section as I would like these entries to remain as a separate "project" to Harrison Network specifically.

It also might be worth noting that these will be the first exception rules `!` in the private section. 

All `_psl` TXT records have been put in place for the corresponding names.

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -13594,6 +13594,15 @@ häkkinen.fi
 // Submitted by William Harrison <psl@hrsn.net>
 hrsn.dev
 
+// Harrison Network - PSL Sandbox : https://sandbox.psl.hrsn.dev
+// Submitted by William Harrison <psl@hrsn.net>
+!ignored.sub.wc.psl.hrsn.dev  
+!ignored.wc.psl.hrsn.dev
+!ignored.sub.psl.hrsn.dev  
+*.sub.wc.psl.hrsn.dev  
+*.wc.psl.hrsn.dev  
+sub.psl.hrsn.dev
+
 // Hashbang : https://hashbang.sh
 hashbang.sh
 
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -13596,12 +13596,12 @@ hrsn.dev
 
 // Harrison Network - PSL Sandbox : https://sandbox.psl.hrsn.dev
 // Submitted by William Harrison <psl@hrsn.net>
-!ignored.sub.wc.psl.hrsn.dev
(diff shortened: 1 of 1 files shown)


wdhdev:
I'll mark this as draft until I can add the `_psl` TXT records.

simon-friedberger:
I've considered doing something like this but adding testing entries to the production PSL should at least have clear goals. What exactly would we be testing and how?

wdhdev:
My goal for these entries would be to test cookie setting across through a web interface.  With the results from these tests we could also make the wiki more specific on how cookied are affected.

Also, it would allow people to test the effects of the PSL without having to add a staging domain or Something similar to the PSL.

dnsguru:
We ended up shooting #1349 behind the woodshed.  I think the premise of allowing cookie testing for a requestor is good. 

IF we set something like this up as the project, I suggest we should probably do so under publicsuffix.org domain vs one any of us individually are operating.

wdhdev:
Having it under publicsuffix.org would be good. Obviously I don't have access to the domain, so I just configured it on one of mine.

Would we be able to configure some entries on publicsuffix.org?

dnsguru:
Let's discuss what we would hope to see accomplished in such a test, and then determine if the juice to squeeze is worth it on dealing with Mozilla's IT kafka pain to get records set up.

wdhdev:
It would certainly be useful for the public to test the real world effects of the PSL, which would help avoid requests like #2288.

simon-friedberger:
How do you imagine this will work? If we set up such a domain, somebody else cannot run anything on it. Would you want to offer some kind of generic backend?

My previous conclusion was, that for testing browser behavior it's much easier to edit `/etc/hosts` and redirect a domain that is on the PSL to something local.

wdhdev:
I was thinking something like a simple static site to easily set cookies across different subdomains to test how different entries work on the PSL.

> If we set up such a domain, somebody else cannot run anything on it.

What do you mean by this? Wouldn't editing `/etc/hosts` solve this like you said before?

simon-friedberger:
I am experimentally adding this.

PR #733, opened 2018-11-25
Add Hakaran domains [caa.li | conf.se | fin.ci | free.hr | ua.rs | aero.al | pro.bj | fin.bj]

hakaran:
Added domains caa.li, conf.se, fin.ci, free.hr, ua.rs, aero.al, pro.bj, fin.bj

* [x] Description of Organization
* [x] Reason for PSL Inclusion
* [x] DNS Verification via dig
* [x] make test

Description of Organization
====
Hakaran group is medium-size web/vps/server hosting provider in EMEA region.

Reason for PSL Inclusion
====
Hakaran group have many clients who need valid hostname for their application, who need dyn-dns service for their dynamic IP address. We want open registration on this domains for everyone who want short URL of your services what havent got domain or want use free-of-charge dyn-dns service.
Secondary, we want for our users open options for use (sub)domains on services as cloudflare, where is need detect is as separate domain, not only subdomain.

DNS Verification via dig
=======
```
# dig +short TXT _psl.caa.li
"https://github.com/publicsuffix/list/pull/733"

# dig +short TXT _psl.conf.se
"https://github.com/publicsuffix/list/pull/733"

# dig +short TXT _psl.fin.ci
"https://github.com/publicsuffix/list/pull/733"

# dig +short TXT _psl.free.hr
"https://github.com/publicsuffix/list/pull/733"

# dig +short TXT _psl.ua.rs
"https://github.com/publicsuffix/list/pull/733"

# dig +short TXT _psl.aero.al
"https://github.com/publicsuffix/list/pull/733"

# dig +short TXT _psl.pro.bj
"https://github.com/publicsuffix/list/pull/733"

# dig +short TXT _psl.fin.bj
"https://github.com/publicsuffix/list/pull/733"
```

make test
=========
`make test` completed successfully

PR #2361, opened 2024-12-28
Remove Hakaran group domains

groundcat:
Originally added in
- #733

Steps:
- I attempted to email `security@hakaran.cz` on 2024-09-05, but it was unreachable.
- The organization's website, `http://hakaran.cz`, is not working.
- I tried [tagging](https://github.com/publicsuffix/list/pull/733#issuecomment-2332990195) the requester on 2024-09-05 in the initial GitHub PR, but received no response.
- No subdomains were observed (CT, Google) for all domains, except for `free.hr`, which seems to have changed ownership (`GRANSY d.o.o.` from WHOIS no longer match `Hakaran group`) and is no longer owned by the initial requesting organization.
- `fin.bj` has already expired and is now available for registration. According to WHOIS: `Domain Status: No Object Found`.
- @wdhdev @fakeboboliu might have more information if you would like to share your other findings. Thanks to @fakeboboliu for reminding me about this.

These indications strongly suggest this no longer belongs to the original requester, and the original organization is out of operation. The domains are no longer with them. I have exhausted all means to contact the original requester; no response after 3 months should be ample time for a missing-in-action status.

If the current registrant/owner of any of these domains would like to keep them in the PSL, they should submit an inclusion request to PSL or update their contact info with reasonable rationales, `_psl` records, expiration, abuse contact info, etc. - we would treat it as equivalent to a brand new PSL inclusion request with all requirements and criteria applicable.

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -13578,14 +13578,6 @@ grayjayleagues.com
 günstigbestellen.de
 günstigliefern.de
 
-// Hakaran group : http://hakaran.cz
-// Submitted by Arseniy Sokolov <security@hakaran.cz>
-fin.ci
-free.hr
-caa.li
-ua.rs
-conf.se
-
 // Häkkinen.fi : https://www.häkkinen.fi/
 // Submitted by Eero Häkkinen <Eero+psl@Häkkinen.fi>
 häkkinen.fi


fakeboboliu:
My research suggests the "hakaran group" may not have been what they claimed.

 - Despite claiming to be an ISP and hosting provider, there's no evidence of their activity on Google. Their website remained "under construction" for years before being suspended in 2021.
 - While they have related [IP ranges](https://bgp.he.net/net/2a10:24c0:6c2::/48#_whois) (but no AS Number) in 2022, [RIPE's database](https://apps.db.ripe.net/db-web-ui/lookup?source=ripe&key=hakaran-mnt&type=mntner) shows these ranges are now controlled by a lawyer specializing in IT disputes.

BTW, the original PR claims `caa.li | conf.se | fin.ci | free.hr | ua.rs | aero.al | pro.bj | fin.bj` but only 5 of them appear in the final list, that's weird.

PR #2369, opened 2025-01-19
Update `ml`

groundcat:
Steps taken:

1. From https://www.iana.org/domains/root/db/ml.html, I identified the registry website http://www.nic.ml. 
2. I found the email address ml@agetic.gouv.ml on http://www.nic.ml. 
3. I sent emails to ml@agetic.gouv.ml and dotml@agetic.gouv.ml. 
4. I received a response from bndiaye@agetic.gouv.ml. Email file in `.eml`: [message-15.zip](https://github.com/user-attachments/files/18468018/message-15.zip)

```
Dear ,

Confirm whether presse.ml should be removed from the PSL
Please keep the presse.ml in PSL

Verify if the remaining entries are accurate and up-to-date
OK


Indicate if there are any other changes needed to the .ML section of the PSL
inst.ml
info.ml
asso.ml
pr.ml
art.ml
ac.ml
```

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -4031,13 +4031,20 @@ net.mk
 org.mk
 
 // ml : https://www.iana.org/domains/root/db/ml.html
+// Confirmed by Boubacar NDIAYE <bndiaye@agetic.gouv.ml> 2025-01-19
 ml
+ac.ml
+art.ml
+asso.ml
 com.ml
 edu.ml
 gouv.ml
 gov.ml
+info.ml
+inst.ml
 net.ml
 org.ml
+pr.ml
 presse.ml
 
 // mm : https://www.iana.org/domains/root/db/mm.html


dnsguru:
concern: 
volunteers@publicsuffix.de address is not part of this project, when was this set up?

wdhdev:
I believe @groundcat owns the publicsuffix.de domain name, so he must've set it up.

groundcat:
No, `volunteers@publicsuffix.de` is not part of the PSL project. It is a mailbox I recently set up, and I plan to share access with @wdhdev and others who are interested. This way, emails can be cross-checked without needing to CC many people and track emails being delivered and responded to. I have not yet figured out how to set it up as a shared inbox—maybe I will just send the login and password to @wdhdev which might be the easiest. I apologize for any confusion. I have been very careful not to misrepresent, and I explicitly noted in this outgoing email that it is not from the project or any maintainer, but just from a volunteer contributor, as you can see  in the .eml file attached above.

dnsguru:
hopefully you can set up an alias much easier than navigating any approval kafka on publicsuffix.org due to mozilla managing it w policies etc, so I get the agility desire.  I've been waiting 15+ years for a publicsuffix.org email address, which would speed things up for dialog with ccTLDs etc.

My ask would be the same as before on the outreach, to be in the loop.

because I am face to face with a lot of the ccTLD admins or know many of them, it would be helpful to be cc'd on requests and responses.  It also projects us better and ensures that when I am standing in front of them presenting to them about reviewing their listings, that if they mention responding to outreach that I am aware of what they're talking about.

dnsguru (approved):
* Chain of communications to the listed IANA Administrator, with response, are documented in the comments originating signed from the admin domain.
* Sorting followed guidelines
* no conflicts with base

APPROVED

simon-friedberger:
Can we maybe move this to `*.ml` like other TLDs which mostly use SLDs?

groundcat:
> Can we maybe move this to `*.ml` like other TLDs which mostly use SLDs?

It seems many websites use the `.ml` top-level domain (e.g., example.ml), as seen from [Google results](https://www.google.com/search?q=site%3Aml). Therefore, consolidating all entries to `*.ml` is not possible without disrupting these sites. 😔

dnsguru:
> Can we maybe move this to `*.ml` like other TLDs which mostly use SLDs?

While this could function, I think that doesn't stretch to fit, or puts in place a reduction in elegance.

In the case where there are explicit, registry-defined stubspaces that would have to co-exist along side third party second level registrations under .ml, the wildcard approach fails to differentiate between the types of administratie boundaries appropriately.

ex:     registry has .ml and also has  ac.ml and art.ml that are etld+1 workalikes.  A third-party registrant goes and registers fac.ml and fart.ml.   being explicit keeps fac.ml or fart.ml from being treated identically to ac.ml and art.ml as etld+1 or conversely allows ac.ml and art.ml to be illustrated as being strings intended by the registry.

Another reason to be more explicit than less explicit exists in the non-browser use cases of the PSL, where there is a need to have 'known' registry subspace(s) being represented within the same administrative boundaries expressed in the ccTLD's ICANN section.

PR #2437, opened 2025-04-15
Remove ie.ua

HostFly:
Deleting a record


simon-friedberger:
Submitter seems to be the same who originally added the domain and I cannot find any usage so SGTM.

HostFly:
_psl.ie.ua IN TXT

PR #2455, opened 2025-04-30
Add biz.gh and net.gh SLD to Public Suffix List

bonbabs92:
# PR Description:
This PR proposes the inclusion of biz.gh and net.gh as entries in the Public Suffix List.

# Reason for Addition:
biz.gh and net.gh are official second-level domains under the .gh ccTLD for Ghana. The Ghana Domain Name Registry delegates these domains and are commonly used by businesses and network-related entities, respectively.

# References:
Ghana Domain Name Registry: https://www.nic.gh/

Example usage: clickmobile.biz.gh, sancfis.net.gh

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -1134,10 +1134,12 @@ org.gg
 // Although domains directly at second level are not possible at the moment,
 // they have been possible for some time and may come back.
 gh
+biz.gh
 com.gh
 edu.gh
 gov.gh
 mil.gh
+net.gh
 org.gh
 
 // gi : http://www.nic.gi/rules.html


simon-friedberger:
There seems to be no official documentation suggesting that these should be added.
We would need this information to be coming from the registry which according to IANA https://www.iana.org/domains/root/db is https://www.nic.gh/. That site does not provide a list of allowed SLDs.

bonbabs92:
Registry https://nic.gh/ page updated with allowed SLDs including net.gh and biz.gh

simon-friedberger:
Is the only change you made here to add the domains to the image? It would be great if there was an official list and in text form. The image for example, does not include .org.gh.

Are these actually available from a registrar? https://ghana.com/ghnic/ has a different list.

bonbabs92:
Text was updated on the registry site to show all available extensions.

simon-friedberger:
Do you want to remove .org.gh then?

bonbabs92:
No, we are maintaining .org.gh as well. All updates done to match, i.e. text and image, have all SLDs.

PR #2456, opened 2025-05-01
fix(readme): remove preceding dots

### LINKED ISSUE #2451: 2nd sentence of README & publicsuffix.org have suffixes with and without a preceding dot
<https://publicsuffix.org/> (emphasis mine):

> A "public suffix" is one under which Internet users can (or historically could) directly register names. Some examples of public suffixes are **.com**, **.co.uk** and **pvt.k12.ma.us**. The Public Suffix List is a list of all known public suffixes.

The same can be seen in the current README.md: <https://github.com/publicsuffix/list/blob/b69f49c54e2ec6b0ba318f9c17b5881d6316b274/README.md?plain=1#L4-L5>

".com" and ".co.uk" have a preceding dot, while "pvt.k12.ma.us" does not. 

I think the the first two should be "com" and "co.uk" instead, w/o the preceding dot.

This seems to be implied by virtually all the sources: by the list (of all known public suffixes) itself containing no preceding dot, various places in PSL repo documentation and tests (e.g. in [Wiki / Format](https://github.com/publicsuffix/list/wiki/Format) "[...] looking up `example.com` should yield `com`" and from "The public suffix" definition under algorithm section), [examples in WHATWG URL standard](https://url.spec.whatwg.org/#host-registrable-domain), tools, Wikipedia, MDN, ...

This could be fixed for both to prevent confusion
- [ ] README.md
- [ ] publicsuffix.org

wdhdev:
Closes #2451 

Website PR: https://github.com/publicsuffix/publicsuffix.org/pull/48

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 # The Public Suffix List
 
 A "public suffix" is one under which Internet users can (or historically could)
-directly register names. Some examples of public suffixes are .com, .co.uk and
+directly register names. Some examples of public suffixes are com, co.uk and
 pvt.k12.ma.us. The Public Suffix List is a list of all known public suffixes.
 
 See https://publicsuffix.org/ and the [Wiki](https://github.com/publicsuffix/list/wiki) link above for more information.


dnsguru:
Just want to make a friendly amendment here that we use markdown 'code' (example ```com```) or double quotes (example  "com") to visually identify the TLDs vs just strip off the dot, so that they stand out.

PR #2497, opened 2025-05-28
Add Cloudflare Subdomain Notice to README.md

groundcat:
This PR adds a notice to the README.md file addressing the recent surge of PSL requests from users attempting to bypass Cloudflare's subdomain limitations on free accounts, similar to the existing Google AdSense notice.

## Background

There has been a significant increase in PSL submissions from domain registrants who have been directed by Cloudflare community moderators (https://www.google.com/search?q=site%3Acommunity.cloudflare.com+%22public+suffix+list%22+AND+%22subdomain%22&ie=UTF-8) or other online forums to add their domains to the PSL simply to enable subdomain management on Cloudflare accounts. 

This misuse of the PSL:

1. Creates unnecessary burden on volunteers & maintainers
2. Potentially harms websites through improper PSL entries
3. Misrepresents the purpose of the PSL

## Rationale

This notice serves to redirect Cloudflare-specific issues back to Cloudflare :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,6 +16,11 @@ Iteration back and forth will delay PR review or inclusion. Be extremely thoroug
 
 ## Important Notices
 
+### 2025-05-27
+Were you directed here to add subdomain(s) to your **Cloudflare** accounts? Please work directly with Cloudflare for subdomain-related questions and account limitations. The PSL is **NOT** intended as a workaround for Cloudflare's subdomain restrictions. 
+
+Consult [Cloudflare's subdomain setup documentation](https://developers.cloudflare.com/dns/zone-setups/subdomain-setup/) or contact Cloudflare directly for subdomain setup questions. Only submit to PSL if your domain truly meets our criteria outlined in [Guidelines](https://github.com/publicsuffix/list/wiki/Guidelines).
+
 ### 2024-07-26
 We are sending emails asking for confirmation if certain entries are still required or need updating.
 


dknecht:
This is misleading advice and not accurate to how Cloudflare works.   Cloudflare only points people to PSL if they explain their use case is to host domains owned by completely separate entities and the need the PSL to inform the browsers to enforce the cookie boundaries correctly.

dnsguru:
> This is misleading advice and not accurate to how Cloudflare works. Cloudflare only points people to PSL if they explain their use case is to host domains owned by completely separate entities and the need the PSL to inform the browsers to enforce the cookie boundaries correctly.

Agree it could be worded better, and we kind of cover it a little bit [HERE](https://github.com/publicsuffix/list/wiki/Third-Party-Diffusion), but the objective is to not have commercial projects dump debt onto a group of volunteers.

There are two things at play here.  It IS appropriate to point projects of scale (say >1-2k established user base at a bare minimum) to the PSL if cloudflare already has identified that they are supporting the domain as an eTLD+ and cookie separation is the sole objective.

If the problem is ```Hey Cloudflare, I am splitting up my domain name into separate administrative accounts, please recognize that.``` then PSL is entirely inappopriate party to be sending people to.

dknecht:
That isn't debt. That is not something related to PSL on our side.

groundcat:
I do appreciate your perspective and acknowledge there's certainly room for improvement on both sides of this discussion. However, your characterization of this educational notice as "misleading" is inaccurate and the tone somewhat surprising, particularly coming from a corporation with a $56.14 billion market cap when engaging with unpaid volunteers maintaining an open source community resource that your corporation both claims to support and actively depends upon.

The information provided in this notice is factually accurate - people do seek advice on Cloudflare's community forum and do receive responses directing them to the PSL, as evidenced by searchable forum posts. Some referrals are indeed appropriate for legitimate use cases like `.gouv.qc.ca` scenarios, others clearly are not, such as the example linked here: https://community.cloudflare.com/t/subdomain-not-added/627389/3

The intent behind this notice is NOT to assign blame to community members or suggest corporate wrongdoing, but rather to provide educational clarity for users who arrive here without fully understanding the implications of PSL inclusion. The notice serves a protective function for both website operators/domain registrants who might inadvertently break their sites and for this volunteer-maintained resource that handles thousands of requests thanklessly.

simon-friedberger:
The PSL is mentioned here: https://developers.cloudflare.com/fundamentals/setup/manage-domains/add-site/ and it also says the following

> Enterprise customers can onboard lower-level subdomains using [Subdomain setup](https://developers.cloudflare.com/dns/zone-setups/subdomain-setup/).

which is the crucial point here. AFAIU (I haven't tried) PSL inclusion is NOT required for subdomain setup.

simon-friedberger:
@dknecht We actually very much appreciate when you send people here. It's probably the main way people find out about the PSL. Unfortunately.

I hope this just tells people to set up their subdomains at CF correctly. I am happy to take wording changes!

dknecht:
As you pointed out from documentation this is not a technical issue on our
end—seems like some are just looking to game the system. Allowing every
subdomain as a separate zone would create massive abuse risks. Grateful for
the work PSL does for the community, which is why we encourage customers to
use it when dealing with multi-tenancy.

simon-friedberger:
@dknecht Just to be clear "allowing every subdomain as a separate zone" is not something the PSL can help prevent. We do almost no vetting before adding entries.

Issue #2491, opened 2025-05-27
Request to Reopen and Approve PSL Inclusion for .in.audio – Business-Critical


anantsparrow:
Dear PSL Team,

I hope this message finds you well. I am writing to respectfully request the reconsideration and reopening of our previously submitted PSL request for the .in.audio domain, which was recently closed due to the absence of active user registrations.

As the registry operator of SparrowHost, I would like to clarify that .in.audio is a newly launched second-level domain extension targeted specifically for a niche but global market of over 40,000 businesses, podcasters, content creators, and independent artists worldwide. While we currently do not have active users, this is solely due to the domain being in its early launch phase. We are actively onboarding partners and preparing for public availability.

However, we are at a critical juncture where inclusion in the PSL is essential before we can begin selling second-level domains (SLDs) under .in.audio for the following reasons:

Security and Privacy
Without PSL inclusion, modern browsers and platforms will treat subdomains under .in.audio as part of the same origin, potentially leading to security issues like cookie leakage and cross-site scripting vulnerabilities. We cannot responsibly launch SLDs for customer use until these risks are mitigated.

Trust and Reputation
PSL inclusion is a critical step in establishing credibility with both users and registrars. If .in.audio is seen as insecure or improperly configured early on, it could irreparably damage the reputation of the TLD, hindering our ability to gain trust from downstream providers and end-users.

Registrar and Partner Requirements
Many registrar partners have explicitly requested PSL inclusion before they agree to list or resell .in.audio domains. Without this step, we are unable to finalize integrations or contracts with distribution partners.

Proper Ecosystem Functionality
Inclusion ensures correct browser behavior, cookie handling, and other technical compatibility that is critical for a commercial domain offering. We are committed to maintaining full compliance with web standards and best practices.

Our goal with SparrowHost is to offer a seamless and secure experience to customers from the very start. We are fully committed to managing .in.audio in a transparent and standards-compliant manner and strongly believe that inclusion in the PSL is not only appropriate but necessary for the safe launch and operation of our domain registry.

Please reconsider the decision to close the request and allow us to proceed with PSL integration for .in.audio. If additional details or documentation are required, I would be more than happy to provide them.

Thank you for your time and understanding.

groundcat:
@anantsparrow I appreciate your persistence, but I need to be direct about a few things that are becoming problematic.

You've created multiple issues in the past and some were duplicates

- #1788
- #1789
- #2075
- #2491

and tend to submit same PRs multiple times

[...]

PR #2498, opened 2025-05-28
Remove ind.mom from PSL

anantsparrow:
Hey there, we are requesting to delist ths .ind.mom suffix from the PSL due to closing of our Registry business. As we are not able to add more suffixes requested by our current registrars. We are not able to make more profit and have lost our invested moneys on premium domains due to not being added on suffix for our future plans.

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -15285,10 +15285,6 @@ srht.site
 apps.lair.io
 *.stolos.io
 
-// SparrowHost : https://sparrowhost.in/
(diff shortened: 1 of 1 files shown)


dnsguru:
#2484 #2493 related

dnsguru:
ties to issue #2491

A PSL entry is not a prerequisite to a registry operation like this.  Having an entry in the PSL should also not be construed by any party as inferring as any form of approval nor should be interpreted as secure or trusted and differently than any other domain name.

PR #2512, opened 2025-06-18
Kerolasa/golangci lint

kerolasa:
This pull request addresses all golangci-lint warnings in this repository. One could say some of the changes are overly pedantic, but that should be fine. It is better to get rid of all the warnings, so that the lint tooling can catch regressions if and when they happen. The golangci-lint i used is v2.

### DIFF
--- a/tools/internal/githistory/history.go
+++ b/tools/internal/githistory/history.go
@@ -55,6 +55,9 @@ func GetPRInfo(gitPath string) (*History, error) {
 		`--grep=^Merge pull request #\d+ from`,
 		"--pretty=%H@%P@%s",
 		"master")
+	if err != nil {
+		return nil, err
+	}
 
 	ret := &History{
 		GitPath: toplevel,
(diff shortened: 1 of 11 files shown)


kerolasa:
The `No suffix blocks changed` test failure might need to be ignored. This change does not attempt to change public suffixes.

weppos:
Hi @kerolasa, thanks for the PR. I think we need to add the golangci lint to the CI workflow, to ensure the compliance is verified and preserved. It will also codify the settings and the version, because I can't see them in this commit.

Can you please:

- Add a `psltool` workflow that includes (a single for now) `golangci-lint` task)? (I see the Go tool tests are currently mixed with the PSL tests in `test.yml`, I will propose to split them out)
- Add a configuration file `.golangci.yml` with the configs you used


```
  golangci-lint:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v4
      - uses: actions/setup-go@v5
        with:
          go-version-file: 'go.mod'
          cache: true
      - name: golangci-lint
        uses: golangci/golangci-lint-action@v8
        with:
          version: v2.1
```

Thanks!

kerolasa:
Hello @weppos. I am not sure if the github workflow change looks the way you hoped. That's my first try with the github workflow, so it might be severely lacking in quality or even working. Anyways, I tried something rebased the pull and added go vendor updates as a minor maintenance bonus.

Oh, and what comes to adding .golangci.yml config. I have had some fun fixing v1 vs v2 golangci issues that I really believe best option is to run without config and let the defaults take precedence.

PR #524, opened 2017-09-06
nic.it - update regions and provinces

daftano:
Update TLDs from nic.it in compliance with version 7.1 of Italian Regulation assignation.
In particular:

regions are listed in Appendix A;
provinces are listed in Appendix B.

Municipalities are omitted due to limits and constraints of PSL. See also #519 for further discussions.

PR #2546, opened 2025-07-24
Remove altoadige.it from the public suffix

vincent-ogury:
altoadige.it is a domain associated to a newspaper in the north of italy. It was created before the regional suffixes.

# Public Suffix List (PSL) Removal

* [X] Robust Reason for PSL Exclusion
* [X] The submitter acknowledges that it is their responsibility to maintain the domains within their section. This includes removing names which are no longer used, retaining the _psl DNS entry, and responding to e-mails to the supplied address. Failure to maintain entries may result in removal of individual entries or the entire section.
 * [X] The [Guidelines](https://github.com/publicsuffix/list/wiki/Guidelines) were carefully _read_ and _understood_, and this request conforms to them.
 * [X] The submission follows the [guidelines](https://github.com/publicsuffix/list/wiki/Format) on formatting and sorting.
 * [X] *Yes, I understand*. I could break my organization's website cookies and cause other issues, and the rollback timing is acceptable. *Proceed anyways*.

## Description of Organization

Alto Adige is an Italian local daily newspaper, based in [Bolzano](https://en.wikipedia.org/wiki/Bolzano). It is sold in [South Tyrol](https://en.wikipedia.org/wiki/South_Tyrol) and since 1999 also in the [province of Belluno](https://en.wikipedia.org/wiki/Province_of_Belluno). Prior to 2000, the newspaper was published with three local editions, for South Tyrol, [Trentino](https://en.wikipedia.org/wiki/Trentino) and Belluno, when was subdivided with two new local newspapers: Trentino and Corriere delle Alpi.

**Organization Website:**

https://www.altoadige.it/

## Reason for PSL Exclusion

The altoadige.it is not a suffix, and exists before the regional suffix were created. It cause trouble to parse this domain is many system.

**Number of users this request is being made to serve:**

## DNS Verification
<!--
For each domain you'd like to add to the list please create
a DNS verification record pointing to your pull request.

For example, if you'd like to add example.com and example.net
you would need to provide the following verifications:

```
dig +short TXT _psl.altoadige.it
# nothing return
```

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -1653,7 +1653,6 @@ agrigento.it
 al.it
 alessandria.it
 alto-adige.it
-altoadige.it
 an.it
 ancona.it
 andria-barletta-trani.it


mozfreddyb:
@vincent-ogury Please make sure to set a DNS txt record for verification.

vincent-ogury:
For which kind of verification?
The domain is declared as a suffix instead of behind a subdomain of "it" prefix. This PR is here to fix the problem.
The website is a newspaper website, and it is there since 1998.

I've already asked to send to our client the relevant links, including this PR, so he can come and comments.
Do he need to add the _pls subdomain even though for a removal ?

mozfreddyb:
Yes, we need the DNS TXT record. How else could we verify that you are authorized to speak for the website?

vincent-ogury:
So we are not able to report an error in your list despite the fact we are not the owner of the domain itself?

We will check with the client if he can come here and keep going on the process.

simon-friedberger:
Unfortunately, we have no way of verifying if there really is an error and removing the domains from the PSL removes some security barriers so we cannot just do it without clear signal from the domain owner.

On the flip side, adding the DNS record is pretty easy once you get a hold of the correct person.

vincent-ogury:
Checking the error, I guess is not that hard. The domain can be called, you reach a website. This website has a whois that go back to 1998. And I guess no suffix allow to host a website directly right ? like .com, .fr, I can't just open that and host a website on it.

I'll ping the owner to come here and keep going on this PR, so we can fix the problem as soon as possible.

simon-friedberger:
Actually, suffixes can host websites just fine. For example, heyflow.page. There are some issues, e.g. DMARC disallows sending emails from public suffixes and public suffixes cannot set domain cookies but hosting websites is no problem.

fakeboboliu:
The domain owner has previously expressed the intention to remove it in #518, and the reason for blocking was that nic.it listed it as one of the reserved SLDs. And the domain owner had contacted nic.it regarding this matter at the time.
It can be observed that this domain still exists in [the 2022 version of the regulations](https://www.nic.it/sites/default/files/documenti/2022/Regulation_assignation_v8.0.pdf) (Appendix B), thus I believe this domain should not be removed for now.

vincent-ogury:
So they kind of encourage the owner to let go the domain by keeping this as a suffix.
No body can ask for a domain ending by this so call suffix. So it's clearly not a suffix.

Why not fixing the issue, and if this change in the futur (which is unlikely, due to the fact the owner has this domain since last century), then we can add it back ?

mozfreddyb:
A very cursory search tells me that there are multiple subdomains (card, card2, cardapp, cardapp2, cardstagte, edicola, infocard, infocard2 and www) which are currently treated by all software as separate domains with full cookie isolation. No one here can tell whether this will just work.

Changing this without the website owner agreeing would mean we'd potentially break the websites with no easy way to roll back (a typical browser release takes 2-3 months from a change to a release). We won't do this just because you think we can always add it back later.

vincent-ogury:
Will you do it if the owner ask for it ? despite the fact the his domain is falsely register as a suffix on nic.it ?

mozfreddyb:
All we need is a DNS record.

dnsguru:
see #524

dnsguru:
@daftano the registry no longer lists the #524 and #519 requested entities can you comment on this PR

daftano:
@dnsguru apologies for a late answer.

@vincent-ogury is right about the newspaper and the domain assigned to them, at the same time the Register still lists AltoAdige.it as a reserved name (see at the bottom of the left column on page 21 of both current [v8.0](https://www.nic.it/sites/default/files/documenti/2022/Regulation_assignation_v8.0.pdf) and previous [v7.1](https://www.nic.it/sites/default/files/archivio/docs/Regulation_assignation_v7.1.pdf) versions of the regulation).

While I believe that may be a mistake on the Register side (and NIC.IT doesn't manage the domain altoadige.it), it is also true the Register reserves the right to revoke a domain name and I do not know if they plan to do so with altoadige.it.
Unless the regulation is reviewed I wouldn't change the public suffix list.

@vincent-ogury I would suggest to invite your client to [contact the register](https://www.nic.it/it/contatti) to update the regulation. If it would be me, I would use the PEC.

PR #2549, opened 2025-07-29
update .it to nic.it regulation v8.0

fakeboboliu:
Update TLDs in compliance with [version 8.0 of nic.it Regulation assignation](https://www.nic.it/sites/default/files/documenti/2022/Regulation_assignation_v8.0.pdf).

Appendix C was omitted per discussion in #519.

What's changed:
 - DUP `trentino.it` in region section (supposed to be province)
 - Province domains integrated / disestablished shown in assignation (but they were restored in 2025 June 1):
   - medio-campidano / campidano-medio: Medio Campidano (integrated, then re-established in 2025)
   - carbonia-iglesias / iglesias-carbonia: Carbonia-Iglesias (integrated, then re-established to a new name)
   - ogliastra / dell-ogliastra: Ogliastra (integrated, then re-established in 2025)
   - olbia-tempio / tempio-olbia: Olbia-Tempio (integrated, then re-established as a new name)
 - `vs.it` maybe typo

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -1527,12 +1527,11 @@ sch.ir
 is
 
 // it : https://www.iana.org/domains/root/db/it.html
-// https://www.nic.it/
+// https://www.nic.it/en/manage-your-it/forms-and-docs -> "Assignment and Management of domain names in the ccTLD.it"
 it
 edu.it
 gov.it
 // Regions (3.3.1)
-// https://www.nic.it/en/manage-your-it/forms-and-docs -> "Assignment and Management of domain names"
 abr.it
 abruzzo.it
 aosta-valley.it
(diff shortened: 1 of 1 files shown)


simon-friedberger:
Spot checking suggests that `trentino.it` and `vs.it` are both in use. Why would we remove this list?

fakeboboliu:
> Spot checking suggests that `trentino.it` and `vs.it` are both in use. Why would we remove this list?

Because they're removed in the NIC's SLD list file.

simon-friedberger:
I understand where you're coming from but there are a bunch of domains here some of which might be providing logins for government services like `https://servizi.comune.aglientu.ot.it/servizionline.aspx` but also things that look like appliances `https://hypersic.comune.tempiopausania.ot.it/`.

I don't want to suddenly create risk here three years after that list was updated by mixing up everybody's cookies. If you are motivated you can ping the ccTLD to confirm if this is what we should do.

fakeboboliu:
You're right, let me close the PR for now and wait for regulation updates: they should be changed in the near future since provinces changed before.

PR #2642, opened 2025-10-30
Remove Netfy Domains from public suffix list

ShayNeeo:
Removed entries for Netfy Domains from the public suffix list.

When checking around the list, I see this domain is expired and should be removed from the list.

'https://netfy.app/' cannot access.
'https://expireddomains.com/domain/netfy.domains/' the netfy.domains is expired.

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -14508,10 +14508,6 @@ o365.cloud.nospamproxy.com
 // Submitted by Philippe PITTOLI <security@netlib.re>
 netlib.re
 
-// Netfy Domains : https://netfy.domains
-// Submitted by Suranga Ranasinghe <security@mavicsoft.com>
-netfy.app
-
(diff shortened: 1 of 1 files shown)


hiifeng:
I noticed that the domain `netfy.app` was added to the PSL via #1991.  
According to subdomainfinder, there are currently no active subdomains.  
Also, crt.sh shows no valid certificates.

https://subdomainfinder.c99.nl/scans/2025-10-30/netfy.app  
https://crt.sh/?q=netfy.app&exclude=expired&group=none  

@Mavicsoft Are you still using this domain?

fakeboboliu:
The Record on `_psl.netfy.app` has been disappeared again since last fix in https://github.com/publicsuffix/list/pull/1991#issuecomment-2557998015 (Dec 2024)

I believe this indicates the original submitter has lost control of the domain and safe to be removed.

PR #2654, opened 2025-11-08
Add guidance on __Host- prefixed cookies to PR template

groundcat:
This PR adds a comment to the PSL submission template to help submitters evaluate whether the __Host- cookie prefix mechanism might address their subdomain isolation needs before submitting to PSL.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -103,6 +103,16 @@ This email address will be used for any required verification or inquiries regar
 
  * [ ] A role-based email address has been used and this inbox is actively monitored with a response time of no more than 30 days.
 
+<!--
+The `__Host-` cookie prefix (RFC 6265bis) provides immediate subdomain isolation 
+by restricting cookies to a specific domain and path with HTTPS requirement—no PSL 
+propagation delays needed. However, PSL serves additional purposes beyond cookie 
+isolation. If your sole need is cookie isolation and you control the application 
+code, `__Host-` prefixes may suffice.
+-->
+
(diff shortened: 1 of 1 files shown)


dnsguru:
I really reccomend we let this run its course in the RFC process within the IETF before tweaking the template.

simon-friedberger:
> I really reccomend we let this run its course in the RFC process within the IETF before tweaking the template.

Let what run its course?

dnsguru:
> > I really reccomend we let this run its course in the RFC process within the IETF before tweaking the template.
> 
> Let what run its course?

This should fully bake in IETF as a standard.  Best I can tell each browser is doing different approaches, so the suggestion while good may lead someone with one browser to tick the box next to done while another might behave differently.

simon-friedberger:
__Host cookies are for example documented here: https://developer.mozilla.org/en-US/docs/Web/Security/Practical_implementation_guides/Cookies#solution

And there are web platform tests to ensure that all browsers do roughly the same here:
https://wpt.fyi/results/cookies/prefix?label=master&label=experimental&aligned

So while there is an open draft, I think the general behavior is sufficiently clear.

IMHO, the point is also not to force people to use __Host cookies instead of being added to the list but to point out that they should be doing that on top because not everybody will have the right PSL version. @groundcat wdyt?

groundcat:
Good point, we're manually mentioning `__Host-` cookies to requestors all the time which isn't efficient, so documenting this somewhere is def needed. I agree adding checkbox in the template might not be the right place for it. Better move might be getting this documented somewhere in the wiki so it's discoverable and we're not repeating ourselves on individual PRs. Closing this for now.

PR #1299, opened 2021-04-27
Added onavstack.net to PRIVATE section of PSL

jbg:
* [x] Description of Organization
* [x] Reason for PSL Inclusion
* [x] DNS verification via dig
* [x] Run Syntax Checker (make test)

* [x] Each domain listed in the PRIVATE section has and shall maintain at least two years remaining on registration.

Description of Organization
====

Organization Website: https://avstack.io

We provide hosted realtime video infrastructure (videoconferencing, livestreaming, virtual events, education and so on).

Reason for PSL Inclusion
====

Customers are able to host realtime A/V infrastructure, including a web frontend, at subdomains of the domain `onavstack.net` proposed for inclusion here. These subdomains should be considered as different origins for security purposes (e.g. w.r.t browser cookies), as these customers are mutually untrusted.

The domain `onavstack.net` is not used for any other purpose.

`onavstack.net` is currently registered up to 2025-01-28 and will always maintain at least two years term.

DNS Verification via dig
=======

```
$ dig +short TXT _psl.onavstack.net
"https://github.com/publicsuffix/list/pull/1299"
```

make test
=========

All tests passed.

PR #2688, opened 2025-12-10
Remove onavstack.net

groundcat:
Removed entries related to AVStack Pte. Ltd. from the public suffix list.

- added by #1299 merged on May 26, 2021
- onavstack.net **expired in August 2025**
- Organization Website: https://avstack.io defunct
- re-registered on `Creation Date: 2025-10-28T12:16:50Z`

```
   Domain Name: ONAVSTACK.NET
   Registry Domain ID: 3033500682_DOMAIN_NET-VRSN
   Registrar WHOIS Server: whois.porkbun.com
   Registrar URL: http://porkbun.com
   Updated Date: 2025-10-28T14:11:59Z
   Creation Date: 2025-10-28T12:16:50Z
   Registry Expiry Date: 2026-10-28T12:16:50Z
```

- email sent to jasper@avstack.io on 2025-12-09 pending response

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -12404,10 +12404,6 @@ translated.page
 myfritz.link
 myfritz.net
 
-// AVStack Pte. Ltd. : https://avstack.io
-// Submitted by Jasper Hugo <jasper@avstack.io>
-onavstack.net
-
 // AW AdvisorWebsites.com Software Inc : https://advisorwebsites.com
 // Submitted by James Kennedy <domains@advisorwebsites.com>
 *.awdev.ca


hiifeng:
@groundcat Thanks for catching this and for opening the PR. I'm really curious how you found out that the domain had expired and got picked up by someone else. If you don’t mind sharing your method, I’d love to learn—so I can also help with this kind of checking when I’ve got some free time.

Thanks again!

groundcat:
@hiifeng Thanks for asking! The process is pretty manual with spot checking, but I do have some automation to help narrow things down.

I run a monitoring script that checks for domains returning NXDOMAIN errors over an extended period. When a domain consistently returns NXDOMAIN for a few days, it usually means the registry has suspended (grace periods) the domain or deleted the entry from the TLD's zone file, which typically indicates the domain has expired and gone through the deletion lifecycle.

I've found that WHOIS data alone isn't really reliable for determining if a domain has actually changed hands. For example, re-registration doesn't necessarily mean it's a different operator (like "123hjemmeside.no" which appears to have been re-registered in 2025 but seems to still be the same operator). WHOIS creation dates can also be misleading.

So my primary signal is **potential zone deletion by the registry** (may lead to persistent NXDOMAIN) rather than RDAP or WHOIS changes. When a domain is absent from the zone file for an extended period, that's a much stronger indicator that it's defunct or expired, and worth investigating for PSL removal.

Hope that helps.

hiifeng:
@groundcat I'm really glad to get your reply! I agree that using a script to check a domain's response via dig is more accurate than relying on WHOIS data. This is a really great idea—thanks so much for sharing!

PR #2712, opened 2025-12-18
Add gov.sk and org.sk

ghost:
gov.sk and org.sk are restricted second-level namespaces under .sk.
Registrations are issued only at the third level (e.g. x.gov.sk, x.org.sk),
and the second-level domains themselves are not registrable.

These namespaces are used for government bodies and organizations respectively,
and function as registry-style delegation boundaries similar to gov.uk and org.uk.

Evidence:
- SK-NIC documentation indicating gov.sk and org.sk are reserved namespaces
- Usage shows consistent third-level-only registrations

PR #2720, opened 2025-12-21
Add org.sk

groundcat:
- Documented at the registry website at https://sk-nic.sk/en/home/ "SK-NIC, a. s. (a member of CentralNic GROUP PLC), is the manager of the top-level domain .sk and the second-level domain org.sk."
- Confirmed by registry through email (from `Hostmaster SK-NIC <hostmaster@sk-nic.sk>`, subject: "Re: [Ticket#2025121900000539] Public Suffix List Submission Request for org.sk and gov.sk - Verification Needed") forwarded to @simon-friedberger : "Yes, we can confirm that form our side its fine. Also, the third level domain is indeed separate domain for org.sk.- org.sk is reserved for non-for-profit organizations and managed by us"
- To close #2712

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -5670,7 +5670,9 @@ si
 sj
 
 // sk : https://www.iana.org/domains/root/db/sk.html
+// https://sk-nic.sk/
 sk
+org.sk
 
 // sl : http://www.nic.sl
 // Submitted by registry <adam@neoip.com>


groundcat:
I’m just being cautious about not including `gov.sk`, although suggested in PR #2712, in this PR. SK-NIC manages `.sk` and `.org.sk`, but it doesn’t seem to manage `.gov.sk`, so the entity that manages `gov.sk` should submit a PR here if they want it added. For example, if some `gov.sk` sites rely on cross-site cookies, adding it to the PSL could break mission-critical sites. So it feels safer to leave it out for now.

PR #2754, opened 2026-01-24
Update .CU

groundcat:
Updated .cu section based on the suffixes listed at https://www.nic.cu/estadisticas.php (official registry website listed by IANA)

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -887,14 +887,19 @@ or.cr
 sa.cr
 
 // cu : https://www.iana.org/domains/root/db/cu.html
+// https://www.nic.cu/estadisticas.php
 cu
+co.cu
 com.cu
+cult.cu
 edu.cu
 gob.cu
 inf.cu
 nat.cu
 net.cu
 org.cu
+sld.cu
+tur.cu
 
 // cv : https://www.iana.org/domains/root/db/cv.html
 // https://ola.cv/domain-extensions-under-cv/


simon-friedberger:
These are listed as "Generic delegates to other organizations" ("Génericos delegados a otras organizaciones") so they should go in the PRIVATE section and be requested by those organizations, right?

dnsguru:
Two comments here, one about obtaining _PSL TXT in the added zones to ensure inclusion is desired, the other is in response to the section location question from @simon-friedberger 

## INTENTIONALITY
Do we know that the registry wants the inclusion of these added strings into the PSL?  Might we be creating an undesired change to their namespace status quo by making the change?  **Should we get this submission directly from the NIC.CU people and the respective parties or at least proof in DNS that this change is one desired by them?**

## SECTION LOCATION 
> These are listed as "Generic delegates to other organizations" ("Génericos delegados a otras organizaciones") so they should go in the PRIVATE section and be requested by those organizations, right?

Good callout - what we have done typically is treat something like this information at the IANA root db registration authority / "NIC" as indicaation of nexus to the registry admin.

For example, .UK has some where the registry delegates some stings to another entity janet in the ICANN section for the TLD, so **locating these in the ICANN section would be appropriate for these .CU namespaces, and the location in the PR is correct.**

Where the namespace being submitted is coming from a third party where there is not any nexus or connection to the NIC, the PRIVATE section would be appropriate.  Like if some third party submitted a string like WEB3.CU to the list, and they were simply a registrant, the string should be located in the PRIVATE section

groundcat:
I'll contact the relevant parties at NIC.CU and the organizations managing these delegated domains to confirm they want these entries added to the PSL. If I hear back from them and they confirm the additions are desired, I'll reopen this PR if necessary.

PR #2781, opened 2026-02-17
add a .git-blame-ignore-revs to ignore #1968

KTibow:
ignore #1968 in git blame


KTibow:
seems git doesn't handle moved lines correctly so this would be a noop.

PR #2881, opened 2026-04-27
Add `codelib.re` Record to Public Suffix List

cskanyryyn9025:
The sibling domain `netlib.re` is already listed in the PSL. To maintain consistent policy, `codelib.re` should follow the same procedure.

The current PSL registry only contains `netlib.re` and is missing the record entry for `codelib.re`. (Ref: #2308)

Explicitly add the `codelib.re` entry to the PSL data file.


simon-friedberger:
Please use the PR template.

PR #3005, opened 2026-07-05
Add namc.ir (Namacdoon)

SMAKH:
namc.ir is the customer-facing domain of Namacdoon (https://namacdoon.ir), a restaurant direct-ordering platform. Each subdomain ({shop}.namc.ir) is an independent restaurant tenant's website with its own content, operated by a different business.

We request private-section inclusion so that browsers and reputation systems treat each tenant as a separate registrable domain (cookie isolation and per-tenant reputation) — same category as other multi-tenant platform entries.

The `_psl` TXT record on namc.ir referencing this PR will be in place within minutes of opening it (record is added via our DNS API immediately after the PR number is known).

Contact: info@namacdoon.ir (verified working).

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -14778,6 +14778,10 @@ cust.retrosnub.co.uk
 // Submitted by Paulus Schoutsen <infra@nabucasa.com>
 ui.nabu.casa
 
+// Namacdoon : https://namacdoon.ir
+// Submitted by MohammadAmin Khodaee <info@namacdoon.ir>
+namc.ir
+
 // Needle Tools GmbH : https://needle.tools
 // Submitted by Felix Herbst <security@needle.tools>
 needle.run


SMAKH:
Domain-control validation is in place: `_psl.namc.ir` TXT → https://github.com/publicsuffix/list/pull/3005 (resolving globally).

mhktanbir:
Pull Requests MUST use the automated template when adding. Do not cut and paste it into a GPT to generate a response or delete / alter the form. Pull requests that do not use the template, that alter it or summarize it will be closed by volunteers without action. Why? There are specific checkboxes that are part of the volunteers' review or are important attestations for public record and transparency and consistency of processing.

SMAKH:
Thanks for the review and the guidance, @mhktanbir. On re-reading the template I realize namc.ir doesn't currently meet the stated inclusion bar for the private section (the ~2000–3000 distinct-tenant threshold — we operate on the order of a dozen restaurant tenants today), so I'm withdrawing this to avoid consuming volunteer time. We'll revisit with a proper template submission if/when we grow into that range. Closing; the _psl TXT record will be removed. Apologies for the non-conforming initial PR.

PR #3054, opened 2026-07-18
Add South Africa (za)

groundcat:
Interestingly, `za` was not included in the PSL.

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -6789,6 +6789,7 @@ net.ye
 org.ye
 
 // za : https://www.iana.org/domains/root/db/za.html
+za
 ac.za
 agric.za
 alt.za


dnsguru:
ZADNA doesn't offer at second level directly, so we should not merge this.

Closing

groundcat:
Thanks for clarifying. It does surface an inconsistency worth flagging, though. I cross-checked the current PSL against registrar & registry pricing dataset, filtering for ccTLDs where SLD forms are registrable (com.xx, net.xx, org.xx, etc.) but the bare apex isn't independently registrable:

For `za, mm, np, pg, jm, ck`: PSL already handles correctly with no bare-TLD line

But there are 13 other ccTLDs currently carrying a bare-TLD line despite the same structural profile (only SLD forms are registrable): 
`fj, gh, lb, lr, mz, ne, ni, pn, py, tt, ye, zm, zw`

Fundamentally I'm curious about whether the PSL is intended to answer the question “Can this domain be registered?”  vs answering something closer to “At what boundary should software treat independently controlled domains as separate sites?” If registry eligibility and registration policy are not what the PSL is intended to validate, should consistency and predictable representation take precedence here?

simon-friedberger:
>  If registry eligibility and registration policy are not what the PSL is intended to validate, should consistency and predictable representation take precedence here?

Q1: Fundamentally I'm curious about whether the PSL is intended to answer the question “Can this domain be registered?” vs answering something closer to “At what boundary should software treat independently controlled domains as separate sites?”
A1: it's B) “At what boundary should software treat independently controlled domains as separate sites?”

Mainly, we are trying to communicate where the organizational boundaries are. For a browser that means which sites are allowed to access domain cookies and communicate by setting the legacy `document.domain`.

Q2: If registry eligibility and registration policy are not what the PSL is intended to validate, should consistency and predictable representation take precedence here?
A2: I have no strong opinion either way. I am not aware any use-cases where it matters.

dnsguru:
What Simon said, with some 'plus'...

The approach that seems to have been in place for PSL has been that it is used as a way for an administrator of a domain to express to the internet how they would prefer that namespace is interacted with.  In the case of many of the listed TLDs, they have an opportunity to express these to us in the form of a pull-request.  

There is a demarcation point, which is the #ICANN or #PRIVATE section placement.  The former (#ICANN) is one that we want only communication and directions from the respective administrator of TLDs or those within that administrative boundary.

There are many integrators and downstream that use only the #ICANN section and skip the #PRIVATE section, where everything else and 'jump-in' points exist that are not direcfly tied to the TLD administrators.    We don't preach on this, nor if we did, would it impact anything, it would still happen.

PR #3069, opened 2026-07-20
chore: Replace broken links, replace IANA db placeholders with specific registry URLs

groundcat:
Fixed broken links in the ICANN section comments, including links with 404 or DNS resolution errors.

Most of the replacement links were sourced from IANA’s Root Zone Database: https://www.iana.org/domains/root/db

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -6507,7 +6507,7 @@ yt
 // U-Label
 
 // xn--mgbaam7a8h ("Emerat", Arabic) : AE
-// http://nic.ae/english/arabicdomain/rules.jsp
+// http://aeda.ae/
 امارات
 
 // xn--y9a3aq ("hye", Armenian) : AM
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -3941,8 +3941,8 @@ ind.kw
 net.kw
 org.kw
 
-// ky : http://www.icta.ky/da_ky_reg_dom.php
-// Confirmed by registry <kysupport@perimeterusa.com> 2008-06-17
+// ky : https://www.iana.org/domains/root/db/ky.html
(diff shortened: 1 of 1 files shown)


simon-friedberger:
@groundcat What's the logic for the `see also:`?

dnsguru:
@groundcat this is good work, and the time investment you made on this is very much appreciated.  

Can you make some tweaks to this - I suggest it gets split  to speed this through for the 'happy path' parts

## Tweaks

### 1: Just swap the default IANA DB page for the TLD with see also (where only 1 'see also')

 It seems like the majority of the `see also:` stuff is where the registration / nic from the respective IANA page is added
 
 This is the case where the first comment URL is the respective IANA db page, and the 'see also' is a different, more focused URL
 
 As a background, we used to have wikipedia as the authoritative source where we didn't have any informational URL for some of the TLDs, so the IANA db link per TLD was put there as a generic default placeholder.

Having the specific registration URL per TLD is a better and more elegant option.  I think this should just be a replacement instead of an additional line.
 
  for those instances where the entry is a more focused one, that would help get the mouse through the maze to the cheese for someone exploring the commented line for some reason, I'd suggest to simply swap out the IANA default page for the nic page that you found.
  
### 2: where > 1 'see also', split to different PR
For others, like AUDA or CIRA as examples where there were more than one 'see also', I'd like to ask you to split those to another PR so we can treat the Tweak 1 list as a 'happy path' PR that goes through fast, and then spend some time on these trickier ones.

groundcat:
Done: Updated this PR to defer the trickier multi-reference comment entries to a separate future PR, so this PR now includes only the “happy path” parts ([part 1](https://github.com/publicsuffix/list/pull/3069#issuecomment-5023802557))

PR #3077, opened 2026-07-22
PR template check

simon-friedberger:
This is not a list change it only changes a GHA.


pencilnav:
@dnsguru The "pr template incorrect" tag is for Simon's GHA PR template check script which was added in this PR. It'll automatically get removed if the author fixed their PR template and as simon said before it's more of a "notice" to let submitters know what they did wrong and fix their issue. 

Not necessarily a duplicate of the original "WRONG PULL TEMPLATE" tag because that's the close reason.

Issue #2612, opened 2025-09-30
Pull requests MUST use the standard pull template


dnsguru:
## Please do not revise the Pull Request Template.

Submit it as-is with your ticks in the right boxes and the wording intact, and make sure to include 2-3 paragraphs on the organization description and rationale for inclusion.

We have had to reject where requesters revising the template when submitting their requests. 

The reason we are rejecting these is :

- The changes side-step some specifically-worded acknowledgements, 
- Rewording or Re-ordering causes volunteers to take more time to review them as we have to hunt for the correlation to the correct ones, and 
- It can break automation testing that is hunting for specific acknowledgements.

dnsguru:
This is being ignored.  2026-05-06: [Updated the readme.md](https://github.com/publicsuffix/list/blob/main/README.md#2026-05-06 ) to add a notice that we will reject / close pull requests that do not use the template as-is.

Volunteers:  Please use the FAIL:WRONG TEMPLATE label when closing these

weppos:
@dnsguru I prepared a Triage workflow, designed to run at the first time the PR is created https://github.com/publicsuffix/list/pull/2902, but I've then realized there's a lot of caveats with it. Especially associated with auto-closing.

TL;DR: I think automating it is still worth it. But I may need some guidance on how the workflow should behave. CLosing the task on PR creation is a bit excessive, as we will need to work around all edge cases like a PR that doesn't change the list itself, or a PR that is not a submission.

Perhaps adding the flag/label alone is the easiest, and we can have a garbage collector task that runs and closes PR with certain FAIL flags after X days. Speaking of it, it may be easier if labels won't have emoji.

Let me know what you think, and I can help to create the workflow.

dnsguru:
@weppos I agree automation is always going to help us - Automatically rejecting PR via automation would save a little time, but I think we can enforce this using the label and closing manually, so that it gives the humans that are first-timers submitting a pull request an opportunity to correct the matter.

I was likely being a hard wall on this with the guidance, but we're seeing an increase in these non-conformant Pull Requests coming in where it loooks like bro just gpt'd their submission to reduce their kafka grief on their side with our intake process, not realizing all the important transparency and consistency the form contributes to the integrity of the PSL on top of our automation to reduce volunteer work.   The gpt'd stuff requires manual review by the volunteers and ends up increasing the review burden.  

We could, instead of closing those bro submissions, offer a minimum 60 month turnaround where not using the template  ;)

[...]

PR #3122, opened 2026-08-05
Added ps.ai

KingAR1000:
Added ps.ai


pencilnav:
#2612 and likely low effort PR.

PR #3019, opened 2026-07-09
Add aivencloud.com subdomains

paulwouters:
# Public Suffix List (PSL) Submission

### Checklist of required steps

* [x] Description of Organization

Organization Website: [aiven.io[ (https://aiven.io/)

Aiven is an Open source data infra-structure company offering streaming and databaser as a managed service. The company aims to simplify data storage and management in the cloud, allowing customers to deploy managed data services across multiple cloud platforms. I am a member of the security team.

* [x] Robust Reason for PSL Inclusion

We previously requested and obtained entries for [aivencloud.com PR#1508] (#1508) and [aiven.app PR#2315] (#2315)

With ~200k entries, our aivencloud.com zone became too big for DNS processing and DNS publishing, so we started migrating these entries to smaller zones (a.aivencloud.com, b.aivencloud.com, etc). These are technical limitations, and not policy restrictions. Our current split allows us to keep individual domains (x.aivencloud.com) to under 20k enties, as of writing the following counts are in these subdomains:

    aivencloud.com - 18312
    a.aivencloud.com - 10183
    b.aivencloud.com - 7808
    c.aivencloud.com - 8160
    d.aivencloud.com - 8833
    e.aivencloud.com - 8705
    f.aivencloud.com - 7124
    g.aivencloud.com - 8463
    h.aivencloud.com - 9676
    i.aivencloud.com - 8876
    j.aivencloud.com - 7962
    k.aivencloud.com - 7185
    l.aivencloud.com - 10641

This new setup now requires we use similar PSL entries to maintain multi-tenant security, such as to indicate these contain entries whose cookies and data should be considered their own registrants/users and not provide cross-side mixing between tenants.

* [x] DNS verification via dig

```
$ dig +short txt _psl.aivencloud.com
"https://github.com/publicsuffix/list/pull/3019"
```

* [x] Each domain listed in the PRIVATE section has and shall maintain at least two years remaining on registration, and we shall keep the `_psl` TXT record in place in the respective zone(s).

aivencloud.com is managed by us and will not be expired within two years. We will maintain _psl TXT records.
`
__Submitter affirms the following:__

 * [x] We are listing *any* third-party limits that we seek to work around in our rationale such as those between iOS 14.5+ and Facebook (see [Issue #1245](https://github.com/publicsuffix/list/issues/1245) as a well-documented example)
 

 - internal DNS management tooling limitations (git repo sizes)
[...]

PR #3169, opened 2026-08-18
Remove non-wildcard `aivencloud.com`

pencilnav:
Related: https://github.com/publicsuffix/list/pull/3019

This PR was submitted to fix a merge error where a `*.aivencloud.com` was merged into the PSL with the existing `aivencloud.com` entry still in place causing downstream confusion.

See also: https://github.com/publicsuffix/list/pull/2771#issuecomment-5297727061 (Comment by @dnsguru)
> We typically are fairly binary about one or the other (with or without) the wildcard - because outside of the algorythm (which different downstream interpret differently), different interpretations happen, and having both typically is problematic due to how it behaves in the wild as result.

@simon-friedberger

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -11344,7 +11344,6 @@ beep.pl
 // Aiven : https://aiven.io/
 // Submitted by Aiven Security Team <security+appdomains@aiven.io>
 aiven.app
-aivencloud.com
 *.aivencloud.com
 
 // Akamai : https://www.akamai.com/


paulwouters:
The reason for both was to allow us to continue moving out of the main zone into the subzones. It would affect both the cookies in aivencloud.com as well as the LetsEncrypt rate limit on the main domain while we are transitioning.

So merging this would break things for us.

pencilnav:
@paulwouters Let's Encrypt has a form to override the rate limit if that's what you need: https://letsencrypt.org/docs/rate-limits/#new-certificates-per-registered-domain

simon-friedberger:
I am not sure what the right way forward is here. It has been discussed in the past. In general adding both 
```
aivencloud.com
*.aivencloud.com
```
to the PSL should mean that these are all public suffixes. The interpretation that having `*.aivencloud.com` on the list also means `aivencloud.com` is a public suffix was popular at some point. Let's merge this for now to get rid of the problem. Of course that means we're punting on the general issue again.

simon-friedberger:
@pencilnav Was there a specific failure that made you create this PR or was it just the failure of the tests? Because we can change the tests...

pencilnav:
@simon-friedberger It was the test and also jothan's comment: https://github.com/publicsuffix/list/pull/2771#issuecomment-5297727061 

> We typically are fairly binary about one or the other (with or without) the wildcard - because outside of the algorythm (which different downstream interpret differently), different interpretations happen, and having both typically is problematic due to how it behaves in the wild as result.

PR #3190, opened 2026-08-25
Add nysky.app

Dr-Cyborg:
nysky.app hosts third-party customer applications on subdomains (`<app>.nysky.app`), similar to vercel.app/netlify.app which are already listed. Each subdomain belongs to a different customer of the nysky platform (https://nysky.no), a Norwegian PaaS. Listing prevents cross-customer cookie injection (`Domain=nysky.app`) and gives each customer app its own cookie/origin boundary.

The apex serves our own marketing site and uses host-only cookies only, so the listing is safe for first-party use.

Requirements:
- Sorted alphabetically in the PRIVATE DOMAINS section (between nsupdate.info and O3O.Foundation)
- `_psl.nysky.app` TXT record pointing to this PR will be published immediately after opening (DNS is operated on our own authoritative servers, ns1/ns2.nysky.no)

Contact: kontakt@nysky.no

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -15023,6 +15023,10 @@ freeddns.us
 nsupdate.info
 nerdpol.ovh
 
+// nysky : https://nysky.no
+// Submitted by Christoffer Nyborg <kontakt@nysky.no>
+nysky.app
+
 // O3O.Foundation : https://o3o.foundation/
 // Submitted by the prvcy.page Registry Team <info@o3o.foundation>
 prvcy.page


pencilnav:
#2612

Dr-Cyborg:
Thanks @pencilnav — and apologies for the non-standard template.

Rather than resubmit with the correct template, I'm closing this. Reading the
template properly made clear that we don't meet the inclusion bar: the
guidelines ask for a minimum of 2000–3000 distinct users, and nysky is an
early-stage Norwegian PaaS with a single-digit number of customers today.
Asking volunteers to review a request that plainly doesn't qualify isn't a
good use of your time.

We've addressed the underlying concern on our own side in the meantime: our
platform cookies use `__Host-`/`__Secure-` prefixes and never carry a `Domain`
attribute, and our ingress now rewrites any `Domain=nysky.app` cookie set by a
customer application down to that application's own hostname — so a cookie set
by one customer's app can no longer reach another's. That removes the
cross-tenant exposure the PSL entry would have prevented.

We'll come back with a properly templated request if and when we genuinely
meet the user threshold. The `_psl.nysky.app` TXT record will be removed now
that this PR is closed.

Thanks for maintaining the list.

PR #3260, opened 2026-09-09
Add dev.wavn.app (Wavn, Inc.)

waelalebrahim:
### Why

Wavn, Inc. hosts customer-created pages on subdomains of dev.wavn.app,
one page per subdomain, each belonging to a different customer.

Without a PSL entry, browsers and cookie libraries treat dev.wavn.app as
a registrable domain. A cookie set with Domain=dev.wavn.app from one
customer's page is therefore accepted and sent to every other customer's
page. Listing dev.wavn.app makes each subdomain its own site, which is
the same arrangement as github.io and pages.dev.

### Expected behaviour

Before:
  getPublicSuffix("a.dev.wavn.app")     -> "wavn.app"
  registrable domain                     -> "wavn.app"
  Cookie Domain=dev.wavn.app             -> accepted, shared across siblings

After:
  getPublicSuffix("a.dev.wavn.app")     -> "dev.wavn.app"
  registrable domain                     -> "a.dev.wavn.app"
  Cookie Domain=dev.wavn.app             -> rejected as a public suffix

### Domain

wavn.app is registered until 2 June 2029, more than two years beyond the
date of this submission. We will keep the registration more than one
year ahead for as long as this entry stands.

### _psl record

A TXT record at _psl.dev.wavn.app pointing at this pull request will be
added as soon as this PR has a number, and left in place permanently.

### DIFF
--- a/public_suffix_list.dat
+++ b/public_suffix_list.dat
@@ -16288,6 +16288,10 @@ wal.app
 // Submitted by Lorentz Kinde <security@wasmer.io>
 wasmer.app
 
+// Wavn, Inc. : https://wavn.app/
+// Submitted by Wael Alebrahim <admin@wavn.app>
+dev.wavn.app
+
 // Webflow, Inc. : https://www.webflow.com
 // Submitted by Webflow Security Team <security@webflow.com>
 webflow.io


waelalebrahim:
Withdrawing this submission.

On reading the template's guidance properly, Wavn does not currently meet
the PSL's stated threshold of 2,000-3,000 distinct users. We are pre-launch,
and the subdomain space this entry covers has very few users on it today.

Rather than ask maintainers to spend time on a request that doesn't meet
the published bar, we'll resubmit when the user count genuinely supports
it. The _psl TXT record stays in place meanwhile and will be repointed at
the new request when we return.

Thanks for the clear template - it was the template itself that made the
threshold obvious.

pencilnav:
#2612

PR #3275, opened 2026-09-14
Add qirsh.app to the PRIVATE section

Zai-d:
### Submission

Adding `qirsh.app` to the PRIVATE section.

**Organisation:** Qirsh — https://qirsh.app
**Submitted by:** Zaid Haddadin <zaidhadadin@gmail.com>
**`_psl` record:** `_psl.qirsh.app` TXT will be set to this PR's URL immediately after opening.

### Why

`qirsh.app` hosts two independent products that share nothing but the parent domain:

- `qirsh.app` — a personal-finance application, fully login-gated
- `settle.qirsh.app` — Settle, a bill-splitting app, published separately on the [App Store](https://apps.apple.com/app/id6803953230) and [Google Play](https://play.google.com/store/apps/details?id=app.qirsh.settle)

They have separate user bases, separate authentication, separate data stores and separate release cycles. A user of one is generally not a user of the other.

Because browsers group saved credentials by the registrable domain, a password saved for one product is offered on the other's sign-in form. The same grouping means a `Domain=.qirsh.app` cookie set by either side would be readable by the other. We would like the boundary between these two products to be a real one.

### Example outcomes

With `qirsh.app` in the PRIVATE section:

- a credential saved on `settle.qirsh.app` is no longer PSL-matched onto `qirsh.app`, and vice versa
- neither product can set a cookie scoped to the other
- the two are treated as distinct sites by anything else consuming the list

### Scope

This is a boundary between two separately published products, not an internal convenience: both are publicly distributed and one ships as a store application with its own user base. We do not delegate subdomains to third parties, and are requesting exactly one entry.

Neither host currently sets any cookie (both use token storage rather than cookies), so no existing `Domain=` cookie is affected by the apex becoming a public suffix.


Zai-d:
Withdrawing this — on reading the template and guidelines properly, it does not meet the bar for the PRIVATE section and I would rather not take up maintainer time with it.

`qirsh.app` hosts two products belonging to one owner. It does not delegate subnames to third parties, so the number of distinct individuals using subnames of the requested space is effectively zero, well short of the 2000–3000 the guidelines ask for. The isolation I was after is a reasonable thing to want, but it is not what the private section is for.

Apologies for the noise, and thanks for the work you all put into the list.

PR #3284, opened 2026-09-17
Add banglacodes.app to the PRIVATE section

shamimnasir:
**Rationale.** BanglaCodes is a Bengali-language website and app builder. Its users publish sites that are each served from their own subdomain of `banglacodes.app`, one label deep: `chayer-dokan.banglacodes.app`. The domain is used for nothing else. The application itself, and the only session cookie we set, live on `banglacodes.com`, a separate registrable domain.

Listing `banglacodes.app` makes each user's site its own cookie origin and its own reputation boundary, which is what we need given that the content is user-authored and each site belongs to a different person.

**Expected behaviour.**

| input | today | with this entry |
|---|---|---|
| `chayer-dokan.banglacodes.app` | public suffix `app`, registrable domain `banglacodes.app` | public suffix `banglacodes.app`, registrable domain `chayer-dokan.banglacodes.app` |
| `banglacodes.app` | registrable domain `banglacodes.app` | unchanged |

Only first-level labels are used; there are no deeper subdomains.

**Registration.** `banglacodes.app` is registered through 2028-09-16, and we commit to keeping more than one year of registration on it at all times.

**Validation.** A `_psl.banglacodes.app` TXT record pointing at this pull request will be added as soon as the PR number exists, and left in place permanently.


shamimnasir:
The `_psl` validation record is live:

```
$ dig +short TXT _psl.banglacodes.app
"https://github.com/publicsuffix/list/pull/3284"
```

Confirmed resolving through 1.1.1.1 and 8.8.8.8. It will stay in place permanently.

pencilnav:
#2612

